PR #1, opened 2017-06-12
Add support for ClusterRoleBinding

micahhausler:
Closes hashicorp/terraform#15194

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ If you wish to work on the provider, you'll first need [Go](http://www.golang.or
 To compile the provider, run `make build`. This will build the provider and put the provider binary in the `$GOPATH/bin` directory.
 
 ```sh
-$ make bin
+$ make build
 ...
 $ $GOPATH/bin/terraform-provider-$PROVIDER_NAME
 ...
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -101,6 +101,7 @@ func Provider() terraform.ResourceProvider {
 			"kubernetes_resource_quota":            resourceKubernetesResourceQuota(),
 			"kubernetes_secret":                    resourceKubernetesSecret(),
 			"kubernetes_service":                   resourceKubernetesService(),
+			"kubernetes_cluster_role_binding":      resourceKubernetesClusterRoleBinding(),
 		},
 		ConfigureFunc: providerConfigure,
 	}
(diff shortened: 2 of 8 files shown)


radeksimko:
Hi @micahhausler 
thank you for taking the time to submit this PR.

We certainly want to cover all reasonable K8S resources (those that make sense to manage via Terraform). ClusterRoleBinding is a reasonable resource to support, but for couple of different reasons I'd like to keep the current coverage limited to v1/stable API:

 - alpha/beta functionality may change and I'd prefer not to deal with any BCs and migration plans for users that rely on Terraform
 - some attributes returned from the API and exposed by Terraform are alpha/beta specific, e.g. `self_link` - this is the case for all K8S and GCP resources. If one day the resource reaches v1 and we upgrade the code, it may break things because of references of those attributes.

TL;DR I'd be very happy to review & merge this PR, but not so much at this point in time.

I hope it's understandable and it won't discourage you from any potential future contributions 🙈 
If you're fancy you can still [compile your own binary](https://www.terraform.io/docs/plugins/basics.html) with any desired alpha/beta resources from your fork/branch. It may complicate distribution, but this is a totally valid reason to do this.

Do you mind if we close this PR and revisit it later when this resource reaches v1/stable?

micahhausler:
Sure thing. I'd like to make a PR documenting the v1 resources preference, would the README be the best place?

Also I feel like I have a good feel of the procedure for adding a K8s resource, could I include a  walk through for doing that in this repo? Should that be in the README or another file?

radeksimko:
@micahhausler Documenting v1 preference is a good idea, README here in this repo is probably the best place.

For the procedure - I think it shouldn't differ from any other resource in any other provider and we already have a guide here: https://www.terraform.io/guides/writing-custom-terraform-providers.html If you feel there's something missing, feel free to send a PR to the main Terraform repo where this piece of documentation lives.

rolandkool:
With the release of Kubernetes 1.8, RBAC moved to v1 (stable). Does that mean this patch can be part of the next release?

radeksimko:
@rolandkool Correct - if it's v1, we "just" need to bump vendored k8s to 1.8 and then raise/reopen this PR. vendor bump needs to come first though and I spent some time on that recently and wasn't very successful. It is on my list though.

sl1pm4t:
> I spent some time on that recently and wasn't very successful.

Oh man, I'm glad I'm not the only one. I tried this recently and went around in circles.

costimuraru:
RBAC has been moved to v1 in Kubernetes. What's the plan here?

PR #11, opened 2017-06-22
vendor auth plugins

e-max:
We vendor all available in version 3.0.0-beta.0 auth plugins


vendor/vendor.json
```diff
@@ -1600,12 +1614,10 @@
...
 		{
-			"checksumSHA1": "xkQLKFsFx+jXOqOAWVAZeMiDLH4=",
+			"checksumSHA1": "GsIX2f2J67NXq1HBiSEQh0SuTB4=",
 			"path": "k8s.io/client-go/util/jsonpath",
-			"revision": "3627aeb7d4f6ade38f995d2c923e459146493c7e",
-			"revisionTime": "2017-03-31T20:34:26Z",
-			"version": "v3.0.0-beta.0",
-			"versionExact": "v3.0.0-beta.0"
+			"revision": "51bf18f5e6291a6e27dc267e99def9635dc2e1c2",
+			"revisionTime": "2017-06-22T03:54:09Z"
```
radeksimko: Can you please keep this pinned to `v3.0.0-beta.0`? From my past experience it's usually better to stick to tagged releases as they tend to be stable than untagged revisions.

e-max: Interesting.   I didn't do it manually, it was a govendor job.  I'll fix it, give me a sec.

radeksimko:
Awesome 🎉 thanks for the quick response.

PR #16, opened 2017-07-03
Expose last warnings from eventlog for Pod+PVC+SVC

radeksimko:
#### Service
```
* kubernetes_service.test: Waiting for service "default/radek-blablah" to assign IP/hostname for a load balancer
   * CreatingLoadBalancerFailed: Error creating load balancer (will retry): Failed to create load balancer for service default/radek-blablah: requested ip 10.0.0.1 is neither static nor assigned to LB a049988bc615511e781c642010a8a005(default/radek-blablah): <nil>
```

#### PVC
```
* kubernetes_persistent_volume_claim.example: timeout while waiting for state to become 'Bound' (last state: 'Pending', timeout: 5m0s)
   * ProvisioningFailed: Failed to provision volume with StorageClass "standard": claim.Spec.Selector is not supported for dynamic provisioning on GCE
```

#### Pod
```
* kubernetes_pod.test: timeout while waiting for state to become 'Running' (last state: 'Pending', timeout: 5m0s)
   * FailedSync: Error syncing pod, skipping: failed to "StartContainer" for "example" with ImagePullBackOff: "Back-off pulling image \"nginx:notexists\""

   * Failed: Failed to pull image "nginx:notexists": rpc error: code = 2 desc = Tag notexists not found in repository docker.io/library/nginx
   * FailedSync: Error syncing pod, skipping: failed to "StartContainer" for "example" with ErrImagePull: "rpc error: code = 2 desc = Tag notexists not found in repository docker.io/library/nginx"
```

### DIFF
--- a/kubernetes/resource_kubernetes_persistent_volume_claim.go
+++ b/kubernetes/resource_kubernetes_persistent_volume_claim.go
@@ -9,7 +9,6 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 	"k8s.io/apimachinery/pkg/api/errors"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	pkgApi "k8s.io/apimachinery/pkg/types"
 	api "k8s.io/kubernetes/pkg/api/v1"
 	kubernetes "k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
@@ -166,7 +165,6 @@ func resourceKubernetesPersistentVolumeClaimCreate(d *schema.ResourceData, meta 
 	name := out.ObjectMeta.Name
 
 	if d.Get("wait_until_bound").(bool) {
-		var lastEvent api.Event
 		stateConf := &resource.StateChangeConf{
 			Target:  []string{"Bound"},
 			Pending: []string{"Pending"},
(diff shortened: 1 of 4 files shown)


kubernetes/resource_kubernetes_persistent_volume_claim.go
```diff
@@ -178,32 +176,22 @@ func resourceKubernetesPersistentVolumeClaimCreate(d *schema.ResourceData, meta
...
+			warns := ""
+			lastWarnings, wErr := getLastWarningsForObject(conn, out.ObjectMeta, "PersistentVolumeClaim", 3)
+			if wErr != nil {
+				return wErr
+			}
+			if len(lastWarnings) > 0 {
+				warns = fmt.Sprintf(". Last warnings: %s", stringifyEvents(lastWarnings))
 			}
-			return fmt.Errorf("%s%s", err, reason)
+			return fmt.Errorf("%s%s", err, warns)
```
tombuildsstuff: given it's being used in multiple resources, any reason to not pull this out into a separate method?

kubernetes/resource_kubernetes_persistent_volume_claim.go
```diff
@@ -178,32 +176,22 @@ func resourceKubernetesPersistentVolumeClaimCreate(d *schema.ResourceData, meta
...
-			reason := ""
-			if lastEvent.Reason != "" {
-				reason = fmt.Sprintf(". Reason: %s: %s", lastEvent.Reason, lastEvent.Message)
+			warns := ""
+			lastWarnings, wErr := getLastWarningsForObject(conn, out.ObjectMeta, "PersistentVolumeClaim", 3)
+			if wErr != nil {
+				return wErr
+			}
+			if len(lastWarnings) > 0 {
+				warns = fmt.Sprintf(". Last warnings: %s", stringifyEvents(lastWarnings))
```
tombuildsstuff: I'd argue this would read cleaner if only returning the combined err + warns object within this if statement?

radeksimko: Sure, I'll change that. 👍

kubernetes/resource_kubernetes_persistent_volume_claim.go
radeksimko: As discussed, we can do this later if the error handling sprawls through more resources. I think there's still benefit in _not_ decoupling it as we're making an API call here and it should be more obvious from these extra LOC that something "heavy-weighted" is happening here. 🙂

PR #17, opened 2017-07-04
Add support for kubernetes_service_account

radeksimko:
### Test results

```
make testacc TEST=./kubernetes TESTARGS='-run=TestAccKubernetesServiceAccount_'
TF_ACC=1 go test ./kubernetes -v -run=TestAccKubernetesServiceAccount_ -timeout 120m
=== RUN   TestAccKubernetesServiceAccount_basic
--- PASS: TestAccKubernetesServiceAccount_basic (6.25s)
=== RUN   TestAccKubernetesServiceAccount_update
--- PASS: TestAccKubernetesServiceAccount_update (9.81s)
=== RUN   TestAccKubernetesServiceAccount_generatedName
--- PASS: TestAccKubernetesServiceAccount_generatedName (2.62s)
PASS
ok  	github.com/terraform-providers/terraform-provider-kubernetes/kubernetes	18.760s
```

### DIFF
--- a/kubernetes/resource_kubernetes_service_account.go
+++ b/kubernetes/resource_kubernetes_service_account.go
@@ -1,0 +1,220 @@
+package kubernetes
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkgApi "k8s.io/apimachinery/pkg/types"
+	api "k8s.io/kubernetes/pkg/api/v1"
+	kubernetes "k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
+)
+
+func resourceKubernetesServiceAccount() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceKubernetesServiceAccountCreate,
+		Read:   resourceKubernetesServiceAccountRead,
+		Exists: resourceKubernetesServiceAccountExists,
+		Update: resourceKubernetesServiceAccountUpdate,
+		Delete: resourceKubernetesServiceAccountDelete,
+
+		// This resource is not importable because the API doesn't offer
+		// any way to differentiate between default & user-defined secret
+		// after the account was created.
+
+		Schema: map[string]*schema.Schema{
+			"metadata": namespacedMetadataSchema("service account", true),
+			"image_pull_secret": {
+				Type:        schema.TypeSet,
+				Description: "A list of references to secrets in the same namespace to use for pulling any images in pods that reference this Service Account. More info: http://kubernetes.io/docs/user-guide/secrets#manually-specifying-an-imagepullsecret",
+				Optional:    true,
+				Elem: &schema.Resource{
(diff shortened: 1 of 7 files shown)


kubernetes/resource_kubernetes_service_account.go
```diff
@@ -0,0 +1,220 @@
...
+
+func resourceKubernetesServiceAccountCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*kubernetes.Clientset)
+
+	metadata := expandMetadata(d.Get("metadata").([]interface{}))
+	svcAcc := api.ServiceAccount{
+		AutomountServiceAccountToken: ptrToBool(false),
+		ObjectMeta:                   metadata,
+		ImagePullSecrets:             expandLocalObjectReferenceArray(d.Get("image_pull_secret").(*schema.Set).List()),
+		Secrets:                      expandServiceAccountSecrets(d.Get("secret").(*schema.Set).List(), ""),
```
tombuildsstuff: should this not be passing in `default_secret_name`?

kubernetes/resource_kubernetes_service_account.go
```diff
@@ -0,0 +1,220 @@
...
+							Type:        schema.TypeString,
+							Description: "Name of the referent. More info: http://kubernetes.io/docs/user-guide/identifiers#names",
+							Optional:    true,
+						},
+					},
+				},
+			},
+			"default_secret_name": {
+				Type:     schema.TypeString,
+				Computed: true,
```
tombuildsstuff: given [this is being used in the Read method](https://github.com/terraform-providers/terraform-provider-kubernetes/pull/17/files#diff-375ac94c5bd4e304ddeb13226700ec77R145), should this also be optional?

kubernetes/resource_kubernetes_service_account.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func resourceKubernetesServiceAccountExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	conn := meta.(*kubernetes.Clientset)
+
+	namespace, name := idParts(d.Id())
+	log.Printf("[INFO] Checking service account %s", name)
+	_, err := conn.CoreV1().ServiceAccounts(namespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		if statusErr, ok := err.(*errors.StatusError); ok && statusErr.ErrStatus.Code == 404 {
```
tombuildsstuff: we could use `http.StatusNotFound` here

radeksimko: Aye, we could do and under many circumstances I would, I just feel that for common codes like 200 and 404 it's just not worth it as three digits are shorter and nearly everyone knows what these mean.

kubernetes/resource_kubernetes_service_account.go
radeksimko: It cannot and should not be specified by the user, so no. It's a computed-only field.
You can call `d.Set` on a computed field at any time in the CRUD, the only difference is that user cannot specify it in the tf/HCL config.

kubernetes/resource_kubernetes_service_account.go
radeksimko: We have no default name at this point yet (during initial creation), so there's nothing to add.

PR #29, opened 2017-07-15
Should return empty array if LabelSelector is not set.

adieu:


radeksimko:
Hi @adieu 
thanks for this PR. The fix looks sensible, but would you mind attaching a test case and/or example of a config affected by this bug?

adieu:
I have added a test case to `resource_kubernetes_persistent_volume_claim_test.go`. This bug will cause `spec.0.selector.#` = 1 even if it's not set.

I didn't run the acceptance test but it should work.

radeksimko:
Unfortunately I was unable to reproduce the bug. The test you attached seems to be passing even without the patch.

I tried reproducing it outside of the test suite too:

```hcl
resource "kubernetes_persistent_volume_claim" "test" {
  metadata {
    annotations {
      TestAnnotationOne = "one"
    }
    labels {
      TestLabelOne = "one"
      TestLabelThree = "three"
      TestLabelFour = "four"
    }
    name = "blablah"
  }
  spec {
    access_modes = ["ReadWriteMany"]
    resources {
      requests {
        storage = "5Gi"
      }
    }
  }
  wait_until_bound = false
}
```
```
$ terraform apply
kubernetes_persistent_volume_claim.test: Creating...
  metadata.#:                               "" => "1"
  metadata.0.annotations.%:                 "" => "1"
  metadata.0.annotations.TestAnnotationOne: "" => "one"
  metadata.0.generation:                    "" => "<computed>"
  metadata.0.labels.%:                      "" => "3"
  metadata.0.labels.TestLabelFour:          "" => "four"
  metadata.0.labels.TestLabelOne:           "" => "one"
  metadata.0.labels.TestLabelThree:         "" => "three"
  metadata.0.name:                          "" => "blablah"
  metadata.0.namespace:                     "" => "default"
  metadata.0.resource_version:              "" => "<computed>"
  metadata.0.self_link:                     "" => "<computed>"
  metadata.0.uid:                           "" => "<computed>"
  spec.#:                                   "" => "1"
  spec.0.access_modes.#:                    "" => "1"
  spec.0.access_modes.1254135962:           "" => "ReadWriteMany"
  spec.0.resources.#:                       "" => "1"
  spec.0.resources.0.requests.%:            "" => "1"
  spec.0.resources.0.requests.storage:      "" => "5Gi"
  spec.0.volume_name:                       "" => "<computed>"
  wait_until_bound:                         "" => "false"
kubernetes_persistent_volume_claim.test: Creation complete (ID: default/blablah)

$ terraform show | grep spec.0.selector
  spec.0.selector.# = 0
```

Would you mind double checking it? Thanks.

adieu:
https://github.com/terraform-providers/terraform-provider-kubernetes/blob/master/kubernetes/structure_persistent_volume_claim.go#L38 This check prevents the bug from happening in PVC. In my code, I didn't do the check so I encountered the bug.

I was moving `LabelSelector` to a shared file so that other resources could use it. The plan was to use this PR to fix the bug and use another PR to restructure the code. What do you think?

radeksimko:
I'm not sure I follow. 🤔 

Do you mind expanding the explanation? Where else do you plan to use LabelSelector? AFAICT there is no bug as the check you referenced in your last comment avoids the problem.

I am certainly interested in fixing bugs, but I'm not sure there is one. I am also happy for us to restructure the code, but only if there is a meaningful plan to do so and if you can share such plan.

Thanks.

adieu:
I extended the kubernetes provider for my TPR resource which has a `LabelSelector` field. So I thought making `LabelSelector` a generic one for all kubernetes resources might be a good idea.   The plan was to make all `LabelSelector` related code into `schemas.go` and `structures.go`.

Because I didn't do the null check in my TPR resource. I found the `flattenLabelSelector` function would return an array with one element when the input is null. IMHO it is a bug. In the acceptance, the null check would not trigger the bug. But if we wrote unittest for this function, properly return an empty array when the input is null should be the correct behavior.

In order not to do two thing at the same time, I split the changes into two commits and planned to send them one by one.

Since there's no other resources in the `terraform-provider-kubernetes` repo using `LabelSelector` now, put those commits in hold until another resource intended to reuse `LabelSelector` sounds like a reasonable plan to me too. It's just a quick fix and code restructure. We could do it whenever needed.

radeksimko:
> I extended the kubernetes provider for my TPR resource

I see, that makes more sense to me now. Thanks for sharing those details, that's helpful.

In either case the current test case modification is unnecessary - in fact I'd prefer if it stayed as is, because it's testing another behaviour - just change of metadata without changing any other arguments.

If we want to cover this with a test then such test should fail before the patch and pass after patch. Perhaps more lightweight approach of testing `flattenLabelSelector` would suffice?

radeksimko:
I made the suggested modifications for you to get this in. I hope you don't mind.
Let me know if there's anything unclear in the diff.

PR #45, opened 2017-08-11
[BETA] Add kubernetes_stateful_set resource

sl1pm4t:
This PR adds the`kubernetes_stateful_set` resource.

I know the policy on beta resources, and don't expect this to be merged until the API is stable.

### DIFF
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -121,6 +121,7 @@ func Provider() terraform.ResourceProvider {
 			"kubernetes_secret":                    resourceKubernetesSecret(),
 			"kubernetes_service":                   resourceKubernetesService(),
 			"kubernetes_service_account":           resourceKubernetesServiceAccount(),
+			"kubernetes_stateful_set":              resourceKubernetesStatefulSet(),
 			"kubernetes_storage_class":             resourceKubernetesStorageClass(),
 		},
 		ConfigureFunc: providerConfigure,
--- a/kubernetes/resource_kubernetes_stateful_set.go
+++ b/kubernetes/resource_kubernetes_stateful_set.go
@@ -1,0 +1,260 @@
+package kubernetes
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkgApi "k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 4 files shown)


radeksimko:
> I know the policy on beta resources, and don't expect this to be merged until the API is stable.

I appreciate you noticed the note 🙂 

As we don't know (or certainly I don't) when the resource hits v1 API the PR may become stale and get into (git) conflict with some other PRs that we'll merge in the meantime. Do you therefore mind me closing this PR for now?

sl1pm4t:
> Do you therefore mind me closing this PR for now?

@radeksimko - yes no problem. I just posted the PR so people can find a fork with the supported resource if they need it.

In regards to handling beta resources - what if we did something like:

* **Now** - Create a resource with `_beta` suffix - e.g. (`kubernetes_stateful_set_beta`) and add to the official Provider.
* **Later** - When the API becomes stable, we port that to a GA resource - e.g. (`kubernetes_stateful_set`) and make sure the GA resource is importable. We would also leave the _beta resource in for a deprecation cycle.

In this way, it's clear the resource is beta - so "user beware". There's little or no risk of BC problems because there's no direct path from the Beta API to GA. Also we'd be providing a fairly "low friction", albeit manual path from the beta to GA resource once it becomes available.

radeksimko:
> I just posted the PR so people can find a fork with the supported resource if they need it.

I don't mind adding link(s) to the Readme for the time being if discoverability is what you're looking for. As long as it comes with clear warning along the lines of "no guarantees" (that's not to downsell your efforts, but to set expectations) I'm ok with that.

> Create a resource with _beta suffix - e.g. (kubernetes_stateful_set_beta) and add to the official Provider.

I believe the `_beta` suffix is something we did discuss in the past, in fact we kind of already tried that approach in Fastly: https://www.terraform.io/docs/providers/fastly/r/service_v1.html

The main downsides are that this potentially results in duplicated code (both code & docs) if you want a real isolation and (more importantly) that we don't have a mechanism to deprecate a whole resource yet.

-----

We're continuously (re)evaluating the idea of supporting alpha/beta APIs in Terraform. It affects other providers too, K8S is not a special case there.

Google Cloud is affected significantly as many resources there remain in alpha/beta for a long time and there's also high demand for those resources within the community, so folks from Google (maintaining that provider) are evaluating one approach. I can't predict how this will go from either code maintenance perspective or how users will or won't accept it in terms of UX. If it works out well I'd be happy to replicate a success in K8S.

Thanks for understanding and patience.

stigok:
> StatefulSets are stable (GA) in 1.9.
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

Is this pull request still up to date?

PR #54, opened 2017-08-21
tests: Skip GCloud tests if no ENV vars provided

radeksimko:
This is a first step in making the provider work outside of GKE/GCloud, e.g. minikube, minishift and generally in most popular k8s environments.

There's loads of failing tests on the mentioned platforms atm, but here we can at least isolate those which won't ever pass.

### DIFF
--- a/kubernetes/provider_test.go
+++ b/kubernetes/provider_test.go
@@ -167,9 +167,12 @@ func testAccPreCheck(t *testing.T) {
 				"KUBE_CLUSTER_CA_CERT_DATA",
 			}, ", "))
 	}
+}
 
+func skipIfNoGoogleCloudSettingsFound(t *testing.T) {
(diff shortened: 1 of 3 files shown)


kubernetes/provider_test.go
```diff
@@ -167,9 +167,11 @@ func testAccPreCheck(t *testing.T) {
 				"KUBE_CLUSTER_CA_CERT_DATA",
 			}, ", "))
 	}
+}
 
+func skipIfNoGoogleCloudSettingsFound(t *testing.T) {
 	if os.Getenv("GOOGLE_PROJECT") == "" || os.Getenv("GOOGLE_REGION") == "" || os.Getenv("GOOGLE_ZONE") == "" {
-		t.Fatal("GOOGLE_PROJECT, GOOGLE_REGION and GOOGLE_ZONE must be set for acceptance tests")
+		t.Skip("GOOGLE_PROJECT, GOOGLE_REGION nor GOOGLE_ZONE are set, skipping")
```
radeksimko: I might not agree with the exact wording, but I agree it should be changed to explain the situation better. I'll change it.

PR #58, opened 2017-08-23
[WIP] r/any: A passthrough resource

nstogner:
*I made this pull request just to get initial feedback on the idea of having this type of resource...*

It would be nice to have a way of letting users have access to pre-stable kubernetes resources without having to create non-stable terraform resources. Some people are using terraform to execute `kubectl apply` for this. However, it would be nice to have terraform also manage deletions.

This pull-request is my initial POC for creating a resource that passes through a config file to kubernetes using the [dynamic client](https://godoc.org/k8s.io/client-go/dynamic). Users could provision alpha/beta resources using this `kubernetes_any` terraform resource and migrate to a typed version once they go fully stable. Kubernetes has a similar declarative model for resource management as terraform, so it seems feasible.

If it makes sense to keep working on this, I will put more effort in to make this a valid pull-request, 
Thoughts?

Example:

```
resource "kubernetes_any" "ngx" {
  object_json = <<EOF
{
   "apiVersion": "apps/v1beta1",
   "kind": "Deployment",
   "metadata": {
      "name": "nginx-deployment"
   },
   "spec": {
      "replicas": 4,
      "template": {
         "metadata": {
            "labels": {
               "app": "nginx"
            }
         },
         "spec": {
            "containers": [
               {
                  "name": "nginx",
                  "image": "nginx:1.7.9",
                  "ports": [
                     {
                        "containerPort": 80
                     }
                  ]
               }
            ]
         }
      }
   }
}
EOF
}
```


radeksimko:
Hi @nstogner 
I understand the motivation, but there's a good reason why I translated the K8S JSON/Swagger schema into Terraform schema (HCL).

There's a number of fields which are computed by the kubelet. Once you submit the resource, the response comes back with some defaults and it would be non-trivial to figure out what is a computed field and whether or not to display diff and what has changed if we're comparing JSON string.

This approach may work for 1 of 5/10 resources, if we're lucky - for all other ones users will either see spurious diffs, or you end up abusing the whole concept of full lifecycle management (which is core feature of Terraform, wanted and expected by users) and _not_ implementing `Read()`. Neither of those situations are nice, IMO.

I appreciate the interest in alpha/beta resources and I accept this is how K8S works (that many resources stay in alpha/beta while users are encouraged to use those APIs long time prior to v1), so the idea of supporting alpha/beta APIs is not shelved definitely. It's just non-trivial to support for us.

Our friends from Google are currently in the stage of evaluating some approaches to this problem over in the GCloud provider and I plan to revisit this idea after some undefined time and potentially use the same approach for K8S provider.

Until then the best thing to do is just wait and upvote issues you'd like to see being supported in https://github.com/terraform-providers/terraform-provider-kubernetes/issues?q=is%3Aopen+is%3Aissue+label%3Aupstream so that once we get to it we know what's the right order for the community.

With all that said I appreciate the time you invested into this PR, but I'd like to close it for the reasons mentioned.

Let me know what you think.

nstogner:
Thanks for taking the time to write a detailed response. I understand the push-back. Good to close.

PR #72, opened 2017-09-19
kubernetes_persistent_volume resource should support storage_class_name

coderfi:
This parameter is optional.
Allows for the fix described in https://stackoverflow.com/a/44891419

### DIFF
--- a/website/docs/r/persistent_volume.html.markdown
+++ b/website/docs/r/persistent_volume.html.markdown
@@ -29,6 +29,7 @@ resource "kubernetes_persistent_volume" "example" {
 				volume_path = "/absolute/path"
 			}
 		}
+		storage_class_name = "standard"
 	}
 }
 ```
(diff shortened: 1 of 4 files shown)


sam-at-luther:
I believe adding this feature will fix the issues I mentioned in https://github.com/terraform-providers/terraform-provider-kubernetes/issues/74.

ccope:
I'm generally :+1: on this, it's less broken and more in-line with the recommendations in https://kubernetes.io/docs/concepts/storage/persistent-volumes/#writing-portable-configuration

I have one concern about backwards-compatibility: What would happen if a user has dynamically created a PV with a PVC claim using the current default storage_class_name value? Does Terraform maintain enough state to realize that the PV and PVC don't need to be created/updated?

The behavior of the default value of storageclassname in a PVC can be complicated, and may need some follow-up work: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1

website/docs/r/persistent_volume.html.markdown
```diff
@@ -29,6 +29,7 @@ resource "kubernetes_persistent_volume" "example" {
 				volume_path = "/absolute/path"
 			}
 		}
+		storage_class_name = "standard"
```
ccope: This should probably be left out of the example resource, because it should be left empty by default.

ccope:
I'd also mark this as a bugfix, not just an enhancement. You can't currently create a PVC in Terraform for PV.

moritzheiber:
Could we please get this PR merged? Deploying apps with state currently is pretty much impossible with this provider

moritzheiber:
@coderfi Would you mind rebasing?

fromz:
Is anyone motivated to merge this? or has there been an alternative solution merged into master? This is a huge issue for people attempting to adopt terraform for kubernetes

coderfi:
Glad this was addressed in #111. Sorry for not keeping up with this PR.

Issue #74, opened 2017-10-06
PersistentVolumeClaim remains in Pending state


puco:
When I try to create PV & PVC in Kubernetes with Terraform targeting a specific existing EBS volume the PVC never gets bound. It looks the PV created using Terraform does not create it in correct manner that it can be successfully bound.

### Terraform Version
Terraform v0.10.7

### Affected Resource(s)
- persistent_volume

### Terraform Configuration Files
```hcl
variable "vol_id" {
  # default = "aws://us-east-1a/vol-1111111111111"
}

provider "kubernetes" {
}

resource "kubernetes_persistent_volume" "pv" {
  metadata {
    name            = "test-volume"
  }
  spec {
    capacity {
      storage       = "1Gi"
    }
    
    access_modes    = ["ReadWriteOnce"]
    
    persistent_volume_source {
      aws_elastic_block_store {
        volume_id   = "${var.vol_id}"
        fs_type     = "ext4"
      }
    }
  }
}

resource "kubernetes_persistent_volume_claim" "pvc" {
  metadata {
    name            = "test-volume"
  }
  spec {
    resources {
      requests {
        storage     = "1Gi"
      }
    }
    access_modes    = ["ReadWriteOnce"]
    volume_name     = "test-volume"
    storage_class_name = "gp2"
  }
  wait_until_bound = false
  depends_on = ["kubernetes_persistent_volume.pv"]
}
```

### Debug Output
https://gist.github.com/puco/23487f4d80001f3d44a41a48b53504cf

### Expected Behavior
I should have a PV & bound PVC

### Actual Behavior
I have a PV in Available state and PVC in Pending state

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Run of the mill Kubernetes 1.8 cluster w/o RBAC on AWS provisioned with kops. I created the EBS volume manually (as that part works w/o any issues).

#### What works
When I create the PV/PVC with kubectl directly they work. When I create the PV with terraform and PVC with kubectl it works as well, so there appears to be something wrong with the PV. I compared the definition (yaml) of both PVs (tf and kubectl) and there appears not to be any issues. Originally I assumed that the issue would be creating EBS volume, PV and PVC in one go but the issue still remains with an existing EBS volume.

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: test-volume-yaml
spec:
  accessModes:
  - ReadWriteOnce
  awsElasticBlockStore:
    fsType: ext4
[...]

PR #111, opened 2018-02-05
resource/kubernetes_persistent_volume: Add support for storage_class_name

sergei-ivanov:
It is basically a partial rebase of #72 plus a few additional fixes to make it actually work.

I have tested it against minikube by creating PVs with the source `host_path`, which is the simplest setup I could produce locally without hitting the cloud.

The testing revealed that creating a PV _without_ a storage class and then amending it to the one _with_ a storage class did not work because Kubernetes sent back a JSON without `/spec/storageClassName` path in it. That resulted in:
```
Error: Error applying plan:
1 error(s) occurred:
* kubernetes_persistent_volume.example: 1 error(s) occurred:
* kubernetes_persistent_volume.example: jsonpatch replace operation does not apply: doc is missing key: /spec/storageClassName
```

I had to tweak `patchPersistentVolumeSpec` a bit to produce either `add` or `replace` operation, which worked a treat in the end: I could add, change or remove storage class in the terraform manifest, and kubernetes applied the changes to the persistent volume.
I am not very experienced in Go, and there may be a more elegant way of doing what I've done,
so please feel free to correct me here.

Closes #72

### DIFF
--- a/kubernetes/resource_kubernetes_persistent_volume.go
+++ b/kubernetes/resource_kubernetes_persistent_volume.go
@@ -60,6 +60,11 @@ func resourceKubernetesPersistentVolume() *schema.Resource {
 							MaxItems:    1,
 							Elem:        persistentVolumeSourceSchema(),
 						},
+						"storage_class_name": {
+							Type:        schema.TypeString,
+							Description: "A description of the persistent volume's class. More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class",
+							Optional:    true,
+						},
 					},
 				},
 			},
--- a/kubernetes/resource_kubernetes_persistent_volume_claim_test.go
+++ b/kubernetes/resource_kubernetes_persistent_volume_claim_test.go
@@ -594,6 +594,7 @@ resource "kubernetes_persistent_volume" "test" {
 			storage = "10Gi"
 		}
 		access_modes = ["ReadWriteOnce"]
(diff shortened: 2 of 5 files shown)


sergei-ivanov:
This should partially address #74.
To fully address it, I think I'll need to roll another PR that will allow specifying empty storage class on PVC (there is a difference in binding behaviour between specifying an empty storage class and specifying no storage class at all).

sergei-ivanov:
I had a go at trying to allow empty or null storage class name for persistent volume _claims_, but that was thwarted by the fact that `storage_class_name` is marked as a _computed_ attribute in the Terraform schema for PVCs. Which means, there is no way to change from a specific storage class to default or empty storage class. I guess it has to stay that way, because the only other option is to query for a storage class that has `storageclass.kubernetes.io/is-default-class: true` and use that to suppress diffs.

fromz:
This patch worked perfectly for me, thanks!

sergei-ivanov:
@radeksimko , thank you for looking after the tests. I did not have a spare GCP project to run them against, so I did not dare touch them. I ran a lot of manual checks against my local minikube though. Everything looks great, thank you for accepting the PR, we are looking forward to it being rolled into the next release of the provider!

fromz:
Thanks guys!

coderfi:
Thanks for picking up the work, @sergei!

PR #114, opened 2018-02-11
WIP: Support Deployments

joshperry:
I've done a little work towards supporting deployments and thought I'd get it out here for people to look at and offer suggestions. I have little go experience and I need this for my first big terraform project.

No pressure ;)

I started by copying the ReplicationController files (because of the similarity to deployments), DRY'd a couple of pieces and updated different slices as I got familiar with the stack. Now focusing on the schema and validation correctness, then onto the flatteners, expanders and API callouts.

Working from [these docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#deployment-v1-apps) and the source.


kubernetes/resource_kubernetes_deployment.go
```diff
@@ -0,0 +1,325 @@
...
+							MaxItems:    1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"rolling_update": {
+										Type:        schema.TypeList,
+										Description: "Rolling update config params. Present only if strategy type = 'RollingUpdate'.",
+										Optional:    true,
+										MaxItems:    1,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
```
joshperry: One thing I could use some help on is the best way to handle validation of one schema value in reference to another, like these two parameters here. Valid values for `max_surge` depend on the current value of `max_available`, and vice versa.

Also, these are a weird `intOrString` variant types, I'm not sure the best way to parse, store, or validate these...

joshperry:
Closing to work on existing deployments support PR

PR #120, opened 2018-02-22
test-infra: Add more targets (kops + minikube)

radeksimko:
screen shot 2018-02-22 at 14 37 42

Test failures on `>1.6` need addressing, but in a separate PR.


kubernetes/test-infra/kops-on-gce/kops-create.sh
```diff
@@ -0,0 +1,50 @@
...
+export KOPS_STATE_STORE="gs://${BUCKET_NAME}"
+export KOPS_FEATURE_FLAGS=AlphaAllowGCE
+
+# Auth for kops
+echo $GOOGLE_CREDENTIALS > $TMP_CREDS_PATH
+export GOOGLE_APPLICATION_CREDENTIALS=$TMP_CREDS_PATH
+
+# Auth for gsutil/gcloud
+gcloud auth activate-service-account $(echo $GOOGLE_CREDENTIALS | jq -r .client_email) --key-file=$TMP_CREDS_PATH
+gcloud config set pass_credentials_to_gsutil true
```
paddycarver: This scares me only a little, because it changes the host's gcloud configuration without changing it back, as far as I can tell, but I have no better suggestion. `gsutil` doesn't leave us many options.

radeksimko: I hear what you're saying. It's sad that we can't just simply pass an ENV variable with JSON and isolate the environment. I think that disabling `pass_credentials_to_gsutil` (at the end of this script) would probably cause more troubles at this point as multiple kops jobs may happen to run at the same time.

I'm open to ideas though.

kubernetes/test-infra/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## How
+
+Spinning up most environments should be as simple as
+
+```
+terraform apply -var=kubernetes_version=1.6.4
+```
+
+(assuming you have any necessary tokens and/or secrets set in your ENV variables)
```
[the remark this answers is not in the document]
radeksimko: Good point, I added explanation to each folder.

kubernetes/test-infra/kops-on-gce/kops-create.sh
paddycarver: I don't know that I have ideas, and honestly think this is as good as we can do given the tooling at hand. But wanted to call it out for future code archaeology to show a decision was made. :)

PR #147, opened 2018-03-16
Added mount_options to spec. closes #146

### LINKED ISSUE #146: FR: Support mount_options for PersistentVolumes
### Terraform Version
v0.11.4

### Affected Resource(s)
- kubernetes_persistent_volume

### Expected Behavior
Kubernetes 1.9 supports the option mount_options in their [PersistentVolumeSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#persistentvolumespec-v1-core), where the operator can add specific options, that should be used when mounting the volume. This option is currently missing from the terraform kubernetes provider.

### References

- [kubernetes-168](https://github.com/kubernetes/features/issues/168)

dploeger:
This fixes #146 (I guess) 😄 

I'm a complete Go-newbie, so please bear with me. There also are test cases missing as I couldn't follow your test setup. Maybe one can help me there. I'm willing to implement that as well. Thanks!


dploeger:
Sorry, it doesn't fix it. I was too naïve with Go. 😄 

I'll try to fix the code.

dploeger:
Ah, crap. mountOptions isn't supported in storage classes either. 😞

dploeger:
So, umm... mountOptions wasn't available in the used version of the GO client, so I had to move to 6.0.0 for this, which implied fixing **a lot** of files to comply with the completely changed code structure. I also added missing vendors.

The code compiles now and the (non-Acceptance-)tests succeed. However, I still get

    spec.0.mount_options: unknown element type (internal)

when using the new code in my terraform project.

It would be great, If somebody could hint me, what I am missing here and also run the acceptance tests.

Thanks!

alexsomesan:
@dploeger I know it's been a while but could you try to rebase this PR. 
We've since updated the vendored client-go and the one in master should have the support you need to implement this.

dploeger:
@alexsomesan I'm sorry, but I have already deleted the source repository and don't need the feature anymore anyway.

PR #152, opened 2018-04-06
validators: change label validation to DNS1123 subdomain

### LINKED ISSUE #87: kubernetes_secret should allow the `.` character in the name
The `.` character is allowed by Kubernetes and we use it for many of our existing secrets.

appilon:
Fix #87


kubernetes/resource_kubernetes_secret_test.go
```diff
@@ -302,6 +321,20 @@ func testAccCheckKubernetesSecretExists(n string, obj *api.Secret) resource.Test
...
+const testAccKubernetesSecretConfig_dotInSecretName = `
+resource "kubernetes_secret" "test" {
+	metadata {
+	  name = "dot.test"
+	}
+  
+	data {
+	  ".start" = "foo"
+	  "mid.dle" = "foo"
+	  "end." = "foo"
```
radeksimko: I don't think we run the validation function for `data` here, so these three examples would be better exercised perhaps in 3 separate `kubernetes_secret`s.

Not a big deal though as we already exercise dot in the middle of the string.

Issue #161, opened 2018-05-03
Feature Request: Support exec Authentication (client-go Credential Plugins)


bflad:
Kubernetes implemented support for an `exec` authentication provider, where the client can generically reach out to another binary to retrieve a valid authentication token. This is used for supporting authentication models such as LDAP credentials or AWS IAM credentials.

This feature is alpha in Kubernetes 1.10, however it is the required model for working with AWS EKS.

Documentation on client-go credential plugins can be found here: https://kubernetes.io/docs/admin/authentication/#client-go-credential-plugins

Presumably a few things will need to happen to land this support:
* Fix up and merge anything related to vendoring the Kubernetes 1.9 client libraries: #117 
* Vendor the Kubernetes 1.10 client libraries, existing PR: #162 

Optionally:
* Add attributes in the provider configuration to manage `exec` authentication instead of requiring an on-disk kubeconfig for this setup

tristanpemble:
I've done a sort of hacky work around for this to get EKS working:

```terraform
```

tristanpemble:
I've gotten around this in a rather hacky way to get EKS initialized with workers:

```hcl
resource "aws_eks_cluster" "default" {
  name = "my-cluster"

  # ...
}

data "external" "heptio_authenticator_aws" {
  program = ["bash", "${path.module}/authenticator.sh"]

  query {
    cluster_name = "my-cluster"
  }
}

provider "kubernetes" {
  host                   = "${aws_eks_cluster.default.endpoint}"
  cluster_ca_certificate = "${base64decode(aws_eks_cluster.default.certificate_authority.0.data)}"
  token                  = "${data.external.heptio_authenticator_aws.result.token}"
  load_config_file       = false
}
```

and the `authenticator.sh`:

```bash
#!/bin/bash
set -e

# Extract cluster name from STDIN
eval "$(jq -r '@sh "CLUSTER_NAME=\(.cluster_name)"')"

# Retrieve token with Heptio Authenticator
TOKEN=$(heptio-authenticator-aws token -i $CLUSTER_NAME | jq -r .status.token)

# Output token as JSON
jq -n --arg token "$TOKEN" '{"token": $token}'
```

aaronmell:
@tristanpemble Whenever I run your example, I get an unauthorized error. Any idea what might be the cause of that? 

[...]

PR #73, opened 2017-10-03
Add support for ClusterRoleBinding

micahhausler:
Re-create of #1, Fixes hashicorp/terraform#15194

@radeksimko I'll rebase once 1.8 is in `/vendor`

PR #162, opened 2018-05-07
Upgrade k8s client libs to 1.10

cosmopetrich:
This is based on the work of @nickschuch and @pdecat in #117. Access to a newer Kube client will help unblock several open PRs that add useful features (ClusterRoleBinging, Deployments, NetworkPolicies) and I thought it might be easier to open a new PR against Kube 1.10 and a newer `master` than to just ping the existing PR. I'm happy to close this if we'd rather proceed with the original PR.

The steps used to create this are almost identical to those used for the 1.9 PR. I also had to update github.com/Azure/go-autorest.

The tests and basic acceptance tests pass. I believe that the acceptance test failure in the 1.9 PR was fixed by #122.


evilmarty:
To help expedite this it's probably a good idea to post test results.

christopherhein:
Anything that is needed to help move this along, would be nice to see EKS support for the Kubernetes TF provider :)

v00d00:
Offering my help to get this merged also? Are we waiting on something specific?

cosmopetrich:
@christopherhein @v00d00 My impression is that this provider is unmaintained at the moment. I'm not sure if there is a way to move the open PRs along until a maintainer is found (see #178).

Until then it might be worth taking a look at the actively-maintained fork https://github.com/sl1pm4t/terraform-provider-kubernetes which includes some additional features such as RBAC and Deployments.

christopherhein:
Thank you @cosmopetrich ! Very helpful!

Jeeppler:
I am unable to use this provider without support for v1.10. This the only version supported by EKS.

alexsomesan:
I've retested this change and it works as expected on top of current master.
I think it's safe to merge.

alexsomesan:
@Jeeppler I'm thinking this change alone doesn't bring a whole lot of value to users - it mostly helps developers. Is this blocking you in any way?
I'd like to wait for https://github.com/terraform-providers/terraform-provider-kubernetes/pull/73 to get merged and then release.

Jeeppler:
@alexsomesan Yes, I work with EKS which uses Kubernetes 1.10, basically I cant use the Kubernetes provider unless I can use it with Kubernetes 1.10.

alexsomesan:
@Jeeppler As far as I could test, it seems the current release of the provider can talk to an EKS cluster just fine. Would you be so kind to provide me with more details of how exactly is it blocking you?

This particular change won't actually give you access to any new 1.10 features unless the specific resources are added or updated explicitly in the K8S provider.

coryflucas:
@alexsomesan I believe this change should allow using a kube config with the aws-iam-authenticator which is used via an ExecCredentials resource in the config. See: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins.

coryflucas:
Also see https://github.com/terraform-providers/terraform-provider-kubernetes/issues/161.

PR #177, opened 2018-07-17
kubernetes/resource_kubernetes_persistent_volume: storage_class_name is now computed

vishen:
Set 'storage_class_name' to be computed for the persistent volume
resource. The persistent volume claim resource does the same thing, so
this should mimic the behaviour in persistent volume claim.

Having one of the storage_class_name's be computed and the other not
causes issues when trying to create a persistent volume and persistent
volume claim, because if there is a StorageClass already created in
Kubernetes, that will be used for the persistent volume claim
storage_class_name, but the persistent volume will be created with an
empty storage_class_name and this fails because Kubernetes validates
that both storage_class_name is the same for both the persistent volume
and persistent volume claim.

### DIFF
--- a/kubernetes/resource_kubernetes_persistent_volume.go
+++ b/kubernetes/resource_kubernetes_persistent_volume.go
@@ -101,8 +101,9 @@ func resourceKubernetesPersistentVolume() *schema.Resource {
 						},
 						"storage_class_name": {
 							Type:        schema.TypeString,
-							Description: "A description of the persistent volume's class. More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class",
+							Description: "The name of the persistent volume's storage class. More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class",
 							Optional:    true,
+							Computed:    true,
 						},
 					},
 				},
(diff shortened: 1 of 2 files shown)


vishen:
@radeksimko I have added an acceptance test, although I only use GKE, and the default storage class name for them is "standard". I was unable to find what the Amazon EKS uses for the default storage class name so I didn't add a test for that.

radeksimko:
@vishen I tried to run your acceptance test in GKE, in cluster of last two GA versions (`1.9.7-gke.3` and `v1.10.5-gke.3`) and I saw a failure in both cases:

```
=== RUN   TestAccKubernetesPersistentVolume_googleCloud_basic
--- FAIL: TestAccKubernetesPersistentVolume_googleCloud_basic (27.60s)
	testing.go:518: Step 0 error: Check failed: 1 error(s) occurred:

		* Check 22/22 error: kubernetes_persistent_volume.test: Attribute 'spec.0.storage_class_name' expected "standard", got ""
```

Do you mind clarifying in which environment did you observe the bug and where did you run the test?

Thanks.

alexsomesan:
Hello again @vishen!

Did you ever get to the bottom of why that ACC test was failing?

vishen:
Ah, sorry, I never had time and never got around to getting to the bottom of it. I will close this as I don't have any time to debug. My apologies.

PR #187, opened 2018-09-25
Bare-metal minikube cluster on Packer

alexsomesan:
Currently the acceptance tests on Packer are non-functional because VirtualBox fails to install correctly.

This change adapts the acceptance test infrastructure on Packer to run the a bare-metal K8S cluster instead of using VirtualBox.


alexsomesan:
That makes sense, Radek. Thanks for the very good points made.

I didn't initially intend to exclude version 1.6.x from the suite, but as it turns out, regardless of the underlying virtualisation used, 1.6 isn't really supported anymore by recent minikube versions. At least not out of the box, without hacks.

I came back to these changes and resorted to use KVM instead of VirtualBox. The reason for it is similar - it installs directly from the official distro repository which is supposed to be more robust. Also, it doesn't require kernel modules to be built at install time, which save some boot time and also increases robustness of the whole provisioning.

Let me know what you think if this approach.

kubernetes/test-infra/minikube-on-packet/20-install-minikube.sh
```diff
@@ -1,10 +1,15 @@
+#!/usr/bin/bash
+
 # Install kubectl
+
 KUBECTL_VERSION=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)
 curl -LO https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl \
-  && chmod +x kubectl && mv kubectl /usr/local/bin/
+  && chmod a+x kubectl && mv kubectl /usr/local/bin/
```
radeksimko: 👍 for explicitness, thanks

kubernetes/test-infra/minikube-on-packet/20-install-minikube.sh
```diff
@@ -1,10 +1,15 @@
...
 kubectl version
 
 # Install minikube
 curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 \
-  && chmod +x minikube && mv minikube /usr/local/bin/
+  && chmod a+x minikube && mv minikube /usr/local/bin/
 minikube version
+minikube config set WantReportErrorPrompt false
```
radeksimko: Nitpick: Can you explain what this does in a brief comment above the line, please?

alexsomesan: If minikube fails for any reason and this is set to `true` (the default) then it will show an interactive prompt asking the user to confirm that minikube can phone back home with diagnostic information. It's undesirable to have this happen in an automated environment like CI. 

The actual prompt message is this
```
================================================================================
An error has occurred. Would you like to opt in to sending anonymized crash
information to minikube to help prevent future errors?
To opt out of these messages, run the command:
	minikube config set WantReportErrorPrompt false
================================================================================
```

alexsomesan:
Yes, we're limited to 1.7.x on platforms that still support it (like minikube).
GKE for one, only supports 1.9.x and 1.10.x as of now.

PR #219, opened 2018-11-20
Add NodeAffinity config to persistence volume

hardchiu:
Support node affinity setting of persistence volume, which is needed when using local volume:
https://kubernetes.io/docs/concepts/storage/volumes/#local

example:
```
resource "kubernetes_persistent_volume" "test" {
  metadata {
    name = "%s"
  }
  spec {
    capacity {
      storage = "2Gi"
    }
    access_modes = ["ReadWriteOnce"]
    persistent_volume_source {
      host_path {
        path = "/mnt/local-volume"
      }
    }
    node_affinity {
      required {
        node_selector_term {
          match_expressions = [{
            key = "%s"
            operator = "In"
            values = ["%s"]
          }]
        }
      }
    }
  }
}
```

The config is following Kubernetes one, it allows smooth transition if they add option at the "required" level in the future.

### DIFF
--- a/kubernetes/resource_kubernetes_persistent_volume_test.go
+++ b/kubernetes/resource_kubernetes_persistent_volume_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 	api "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -41,9 +42,9 @@ func TestAccKubernetesPersistentVolume_googleCloud_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("kubernetes_persistent_volume.test", "metadata.0.labels.TestLabelTwo", "two"),
 					resource.TestCheckResourceAttr("kubernetes_persistent_volume.test", "metadata.0.labels.TestLabelThree", "three"),
 					testAccCheckMetaLabels(&conf.ObjectMeta, map[string]string{
-						"TestLabelOne":                             "one",
-						"TestLabelTwo":                             "two",
-						"TestLabelThree":                           "three",
+						"TestLabelOne":   "one",
+						"TestLabelTwo":   "two",
+						"TestLabelThree": "three",
 						"failure-domain.beta.kubernetes.io/region": region,
 						"failure-domain.beta.kubernetes.io/zone":   zone,
 					}),
@@ -73,9 +74,9 @@ func TestAccKubernetesPersistentVolume_googleCloud_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("kubernetes_persistent_volume.test", "metadata.0.labels.TestLabelTwo", "two"),
 					resource.TestCheckResourceAttr("kubernetes_persistent_volume.test", "metadata.0.labels.TestLabelThree", "three"),
 					testAccCheckMetaLabels(&conf.ObjectMeta, map[string]string{
-						"TestLabelOne":                             "one",
-						"TestLabelTwo":                             "two",
-						"TestLabelThree":                           "three",
+						"TestLabelOne":   "one",
+						"TestLabelTwo":   "two",
+						"TestLabelThree": "three",
 						"failure-domain.beta.kubernetes.io/region": region,
 						"failure-domain.beta.kubernetes.io/zone":   zone,
 					}),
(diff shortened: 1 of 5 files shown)


kubernetes/structure_persistent_volume_spec.go
```diff
@@ -1028,6 +1034,14 @@ func patchPersistentVolumeSpec(pathPrefix, prefix string, d *schema.ResourceData
 			})
 		}
 	}
+	if d.HasChange(prefix + "node_affinity") {
+		v := d.Get(prefix + "node_affinity").([]interface{})
+		nodeAffinity := expandVolumeNodeAffinity(v)
+		ops = append(ops, &ReplaceOperation{
```
alexsomesan: I'm not entirely sure it's valid to pass complex types like `VolumeNodeAffinity` directly into a patch operation.

Would you mind adding a new acceptance test that exercises this patch operation?

alexsomesan:
Hello and thanks for taking the time to add this.
Overall this looks good. There's just one concern I have regarding the way patching is implemented.
Please see my comment down below.

kubernetes/structure_persistent_volume_spec.go
hardchiu: I see the RBDPersistentVolumeSource is also a similar complex type, so I think that is valid to pass.

Updated the acceptance test to cover more operations.

kubernetes/structure_persistent_volume_spec.go
```diff
@@ -1028,6 +1034,27 @@ func patchPersistentVolumeSpec(pathPrefix, prefix string, d *schema.ResourceData
 			})
 		}
 	}
+	if d.HasChange(prefix + "node_affinity") {
```
alexsomesan: There is a bit of a conflict here.
The schema sets all the sub-attributes of `node_selector_term` as `ForceNew`. This negates the need for patch operations to be implemented for them, like you have done here.

Here's the result of a quick test I did trying to modify `node_selector_term`. I cannot get the plan to show a modify (update) operation as it's always trying to replace the resource because of the `ForceNew` flags.

```
------------------------------------------------------------------------

-/+ destroy and then create replacement

-/+ kubernetes_persistent_volume.test (new resource required)
      id:                                                                                           "test-pv-affinity" => <computed> (forces new resource)
      metadata.#:                                                                                   "1" => "1"
      metadata.0.generation:                                                                        "0" => <computed>
      metadata.0.name:                                                                              "test-pv-affinity" => "test-pv-affinity"
      metadata.0.resource_version:                                                                  "2975" => <computed>
      metadata.0.self_link:                                                                         "/api/v1/persistentvolumes/test-pv-affinity" => <computed>
      metadata.0.uid:                                                                               "ef41d59b-fe8f-11e8-9f90-000c2979c097" => <computed>
      spec.#:                                                                                       "1" => "1"
      spec.0.access_modes.#:                                                                        "1" => "1"
      spec.0.access_modes.1254135962:                                                               "ReadWriteMany" => "ReadWriteMany"
      spec.0.capacity.%:                                                                            "1" => "1"
      spec.0.capacity.storage:                                                                      "10Mi" => "10Mi"
      spec.0.node_affinity.#:                                                                       "1" => "1"
      spec.0.node_affinity.0.required.#:                                                            "1" => "1"
      spec.0.node_affinity.0.required.0.node_selector_term.#:                                       "1" => "1"
      spec.0.node_affinity.0.required.0.node_selector_term.0.match_expressions.#:                   "1" => "2" (forces new resource)
      spec.0.node_affinity.0.required.0.node_selector_term.0.match_expressions.0.key:               "hostname" => "hostname"
      spec.0.node_affinity.0.required.0.node_selector_term.0.match_expressions.0.operator:          "In" => "In"
      spec.0.node_affinity.0.required.0.node_selector_term.0.match_expressions.0.values.#:          "1" => "1"
      spec.0.node_affinity.0.required.0.node_selector_term.0.match_expressions.0.values.2666930069: "foobar" => "foobar"
      spec.0.node_affinity.0.required.0.node_selector_term.0.match_expressions.1.key:               "" => "kubernetes.io/hostname" (forces new resource)
      spec.0.node_affinity.0.required.0.node_selector_term.0.match_expressions.1.operator:          "" => "In" (forces new resource)
      spec.0.node_affinity.0.required.0.node_selector_term.0.match_expressions.1.values.#:          "" => "1" (forces new resource)
      spec.0.node_affinity.0.required.0.node_selector_term.0.match_expressions.1.values.1110466152: "" => "minikube" (forces new resource)
      spec.0.persistent_volume_reclaim_policy:                                                      "Retain" => "Retain"
      spec.0.persistent_volume_source.#:                                                            "1" => "1"
      spec.0.persistent_volume_source.0.host_path.#:                                                "1" => "1"
      spec.0.persistent_volume_source.0.host_path.0.path:                                           "/mnt/local-volume" => "/mnt/local-volume"


Plan: 1 to add, 0 to change, 1 to destroy.

------------------------------------------------------------------------
```

kubernetes/resource_kubernetes_persistent_volume_test.go
```diff
@@ -743,3 +832,45 @@ resource "kubernetes_storage_class" "test2" {
 }
 `, name, refName, diskName, zone, storageClassName, storageClassName2)
 }
+
+func testAccKubernetesPersistentVolumeConfig_hostPath_nodeAffinity(name, nodeAffinity string) string {
+	return fmt.Sprintf(`
+resource "kubernetes_persistent_volume" "test" {
```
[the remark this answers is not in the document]
hardchiu: Thought the `make fmt` will take care all the formatting.

I have fixed the space for the HCL I added, didn't touch the other HCL with tabs in the same files.

kubernetes/structure_persistent_volume_spec.go
hardchiu: The node_selector_term is actually similar to the label selector in schema_label_selector.go

They are similar but not the same. It is used for selecting node for pod.
I see there is ForceNew in the label selector, so maybe better to keep it.

I have reverted the change of supporting more patch operation.
It should still work, the test pass with ReplaceOperation.

mbarrien:
Very late comment on this... I believe nodeSelectorTerms actually allows multiple values (i.e. it's a list), and this implementation seems to have hard coded a max length of 1. Is this intentional?

alexsomesan:
Good point, @mbarrien
I am working on pod affinity at the moment, which touches on this.
I will doublecheck this and fix it if necessary.

PR #223, opened 2018-11-21
Kops on AWS test env - create state S3 bucket via Terraform

alexsomesan:
We run acceptance tests on clusters constructed with Kops on AWS. 
Kops requires a pre-existing S3 bucket to store its state in.

With this change, we create the bucket using Terraform itself, as a resource, instead of the AWS CLI.
Aside from being more explicit, this enables spinning up this environment in other regions.
To do this, the AWS CLI would otherwise require some additional JSON arguments (--create-bucket-configuration).

@radeksimko

### DIFF
--- a/kubernetes/test-infra/kops-on-aws/main.tf
+++ b/kubernetes/test-infra/kops-on-aws/main.tf
@@ -39,8 +39,9 @@ resource "tls_private_key" "ssh" {
 }
 
 resource "aws_s3_bucket" "kops_state_store" {
-  bucket = "${local.bucket_name}"
-  acl    = "private"
+  bucket        = "${local.bucket_name}"
+  acl           = "private"
+  force_destroy = true
 
   tags {
     Usage     = "kops_state_store"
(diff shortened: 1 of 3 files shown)


kubernetes/test-infra/kops-on-aws/main.tf
```diff
@@ -35,15 +35,27 @@ data "http" "ipinfo" {
...
 
 resource "tls_private_key" "ssh" {
-  algorithm   = "RSA"
+  algorithm = "RSA"
+}
+
+resource "aws_s3_bucket" "kops_state_store" {
+  bucket = "${local.bucket_name}"
+  acl    = "private"
+
```
radeksimko: Are you sure we don't need [`force_destroy = true`](https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#force_destroy) here? I remember kops can sometimes leave files in the bucket even after destroying the cluster.

alexsomesan: That's a good point. It won't clean it if there are files still in there. I'll add a `force_destroy`.

PR #231, opened 2018-11-28
Convert returned types to pointers

alexsomesan:
This change aligns expander functions to return pointers to complex types rather than values.

This is a pattern most providers try to adhere to as it helps with detecting malformed data earlier, through exercising the tests and seeing potential panics.

* [x] convert returned values to pointers
* [x] exercise all tests and inspect issues

@tombuildsstuff

### DIFF
--- a/kubernetes/structures.go
+++ b/kubernetes/structures.go
@@ -184,8 +184,8 @@ func flattenResourceList(l api.ResourceList) map[string]string {
 	return m
 }
 
-func expandMapToResourceList(m map[string]interface{}) (api.ResourceList, error) {
-	out := make(map[api.ResourceName]resource.Quantity)
+func expandMapToResourceList(m map[string]interface{}) (*api.ResourceList, error) {
+	out := make(api.ResourceList)
 	for stringKey, origValue := range m {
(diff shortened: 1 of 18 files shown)


kubernetes/structures.go
```diff
@@ -310,8 +310,8 @@ func resourceListEquals(x, y api.ResourceList) bool {
 	return true
 }
 
-func expandLimitRangeSpec(s []interface{}, isNew bool) (api.LimitRangeSpec, error) {
-	out := api.LimitRangeSpec{}
+func expandLimitRangeSpec(s []interface{}, isNew bool) (*api.LimitRangeSpec, error) {
+	out := &api.LimitRangeSpec{}
```
tombuildsstuff: I don't think this needs changing - just an FYI - in Azureland we define these objects normally and then return the reference in the return e.g.

```
out := api.LimitRangeSpec{}
# ...
return &out
```

alexsomesan: Yeah, I had that inner dilemma myself.
Went for this approach to declutter a few `&` out of the code. I guess it's fine either way.

PR #246, opened 2018-12-06
Add Pod tolerations

grahamar:
Adds support for the `tolerations` field for a Pod spec.

### DIFF
--- a/kubernetes/schema_pod_spec.go
+++ b/kubernetes/schema_pod_spec.go
@@ -153,6 +153,41 @@ func podSpecFields(isUpdatable bool) map[string]*schema.Schema {
 			ValidateFunc: validateTerminationGracePeriodSeconds,
 			Description:  "Optional duration in seconds the pod needs to terminate gracefully. May be decreased in delete request. Value must be non-negative integer. The value zero indicates delete immediately. If this value is nil, the default grace period will be used instead. The grace period is the duration in seconds after the processes running in the pod are sent a termination signal and the time when the processes are forcibly halted with a kill signal. Set this value longer than the expected cleanup time for your process.",
 		},
+		"tolerations": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "If specified, the pod's tolerations. Optional: Defaults to empty",
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"effect": {
+						Type:        schema.TypeString,
+						Description: "Effect indicates the taint effect to match. Empty means match all taint effects. When specified, allowed values are NoSchedule, PreferNoSchedule and NoExecute.",
+						Optional:    true,
+					},
+					"key": {
+						Type:        schema.TypeString,
+						Description: "Key is the taint key that the toleration applies to. Empty means match all taint keys. If the key is empty, operator must be Exists; this combination means to match all values and all keys.",
+						Optional:    true,
+					},
+					"operator": {
+						Type:        schema.TypeString,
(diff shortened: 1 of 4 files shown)


kubernetes/structures_pod.go
```diff
@@ -606,6 +639,33 @@ func expandSecretVolumeSource(l []interface{}) *v1.SecretVolumeSource {
 	return obj
 }
 
+func expandTolerations(tolerations []interface{}) []v1.Toleration {
```
alexsomesan: This method should return a slice of pointers rather then values, like this`[]*v1.Toleration`.

kubernetes/schema_pod_spec.go
```diff
@@ -153,6 +153,41 @@ func podSpecFields(isUpdatable bool) map[string]*schema.Schema {
 			ValidateFunc: validateTerminationGracePeriodSeconds,
 			Description:  "Optional duration in seconds the pod needs to terminate gracefully. May be decreased in delete request. Value must be non-negative integer. The value zero indicates delete immediately. If this value is nil, the default grace period will be used instead. The grace period is the duration in seconds after the processes running in the pod are sent a termination signal and the time when the processes are forcibly halted with a kill signal. Set this value longer than the expected cleanup time for your process.",
 		},
+		"tolerations": {
```
alexsomesan: The attribute name should be singular, as in `toleration`.
This is because in HCL, to express the *tolerations* list with multiple elements, you would have to write multiple distinct blocks of:
```hcl
toleration {
  key = "key"
  operator = "Equal"
  value = "value"
  effect = "NoSchedule"
}

toleration {
  key = "key"
  operator = "Exists"
  effect = "NoSchedule"
}
```

kubernetes/structures_pod.go
grahamar: The `v1.PodSpec` type expects a slice of literals, same as the `Containers` field, so this is the same as `expandContainers`. Unless I'm missing something? 😃

kubernetes/schema_pod_spec.go
grahamar: Ah yes of course, that's a good point

kubernetes/structures_pod.go
alexsomesan: We're trying to standardize on this pattern of expanders returning pointers. The bulk of the change happened in https://github.com/terraform-providers/terraform-provider-kubernetes/pull/231
In doing this, among other things, we benefit from detecting the inadvertent use of nil return values early on during test runs (as might be the case when an error occurred).

jeffmhastings:
Has then been any progress on this @grahamar ? If not I'd definitely be willing to help out. Would you accept a PR into your fork?

karlhungus:
@grahamar @jeffmhastings I'd also like to see this happen, and have some updates that address the little comments: https://github.com/terraform-providers/terraform-provider-kubernetes/compare/master...karlhungus:tolerations?expand=1#diff-87d7af58ecf86062f1ac897d735c8c0d 

I'm hoping to start digging into acceptance tests and the update function today; but I'm not sure the protocol for contributions.

grahamar:
@karlhungus @jeffmhastings Apologies I haven't had a chance to take a look at this, as I'm getting married in the next few days. Feel free to add to this PR or create your own, or I can pick this up in a couple weeks again.

the-nicolas:
Any update? Just moved to write my k8s specs with terraform and now I figured out, that I cannot set the needed "NoSchedule" :-(

...so back to handcrafted for now

grahamar:
closing in favor of #294

PR #251, opened 2018-12-10
add missing automount_service_account_token to the pod spec

zimbatm:
This change propagates to the replication_controller, services and
deployments.

See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#pod-v1-core


jhoblitt:
@alexsomesan / @zimbatm I've opened #261, which is otherwise identical but adds the requested `go fmt` fix.

zimbatm:
thanks! I missed the original notifications for some reason. Happy for you to take over :)

PR #259, opened 2018-12-18
GKE test infra writes kubeconfig file

alexsomesan:
This change allows the template that constructs the GKE test infrastructure to write the cluster access configuration to a kubeconfig file in the local build folder. 

This is to allow the test infrastructure to be easily used for local development and test runs. Exporting the kubeconfig allows it to be easily accessed using `kubectl` as well as other client-go based apps.

Previously, seven different template outputs would be presented to the user and would have to manually be injected into clients.

To use it, set the `KUBECONFIG` environment variable like this:

`export KUBECONFIG="${KUBECONFIG}:$(terraform output kubeconfig_path)"`

### DIFF
--- a/kubernetes/test-infra/gke/kubeconfig-template.yaml
+++ b/kubernetes/test-infra/gke/kubeconfig-template.yaml
@@ -1,0 +1,23 @@
+apiVersion: v1
+kind: Config
+preferences:
+  colors: true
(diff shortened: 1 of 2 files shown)


alexsomesan:
@appilon It does. It requires changes to the CI job config to export that KUBECONFIG variable.
That's already in place.

PR #294, opened 2019-02-01
Tolerations 

karlhungus:
#### Note 
I'm not sure the best way to go about this.

#### What this does
Adds tolerations by taking @grahamar 's PR: https://github.com/terraform-providers/terraform-provider-kubernetes/pull/246 and adjusts (or attempts to) for the changes suggested by @alexsomesan 

@alexsomesan I believe this addresses all of your [comments](https://github.com/terraform-providers/terraform-provider-kubernetes/pull/246#pullrequestreview-182424878), with the exception of the update on patch, which looks to be `ForceNew` by default: 
 https://github.com/karlhungus/terraform-provider-kubernetes/blob/3aed531e376502959f5c6b3583504ced74211398/kubernetes/schema_pod_spec.go#L244


mbarrien:
Note that even with the changes, my comment at https://github.com/terraform-providers/terraform-provider-kubernetes/pull/246#issuecomment-459838959 still applies; 0 is different from missing (0 means evict immediately, missing means never evict, according to https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#toleration-v1-core). This code appears to default to 0, which may be different than what the user expects.

kubernetes/resource_kubernetes_deployment_test.go
```diff
@@ -179,6 +179,62 @@ func TestAccKubernetesDeployment_with_security_context(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccKubernetesDeploymentConfigWithTolerations(rcName, imageName, nil, operator, &value),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckKubernetesDeploymentExists(deploymentTestResourceName, &conf),
+					resource.TestCheckResourceAttr(deploymentTestResourceName, "spec.0.template.0.spec.0.toleration.0.effect", "NoExecute"),
+					resource.TestCheckResourceAttr(deploymentTestResourceName, "spec.0.template.0.spec.0.toleration.0.key", "myKey"),
+					resource.TestCheckResourceAttr(deploymentTestResourceName, "spec.0.template.0.spec.0.toleration.0.operator", operator),
+					resource.TestCheckResourceAttr(deploymentTestResourceName, "spec.0.template.0.spec.0.toleration.0.value", "value"),
+					resource.TestCheckResourceAttr(deploymentTestResourceName, "spec.0.template.0.spec.0.toleration.0.toleration_seconds", "0"),
```
karlhungus: This was unexpected: `resource.TestCheckNoResourceAttr` here

kubernetes/structures_pod.go
```diff
@@ -663,8 +663,9 @@ func expandTolerations(tolerations []interface{}) []*v1.Toleration {
 		if value, ok := m["operator"]; ok {
 			ts[i].Operator = v1.TolerationOperator(value.(string))
 		}
-		if value, ok := m["toleration_seconds"]; ok {
-			ts[i].TolerationSeconds = ptrToInt64(int64(value.(int)))
+		sec := m["toleration_seconds"].(int)
+		if sec > 0 {
```
mbarrien: The problem is that 0 and negative numbers are actually valid values; it means to evict immediately. If you treat 0 as missing which this seems to do, that means something different; missing means to never evict if it has already been scheduled but is no longer compatible with the toleration.

How do we distinguish between a user explicitly setting "toleration_seconds=0" vs omitting it? Unfortunately, thanks to [hashicorp/terraform#16796](https://github.com/hashicorp/terraform/issues/16796) we can't tell the difference.

karlhungus: Right - I didn't really fully clue into that. 

I think we can use a hacky work around and put -1 for any immediate evictions, and use 0 as the unset value.

mbarrien: 2 mild problems with this proposal.

1) The [Kubernetes reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#toleration-v1-core) for tolerationSeconds states "Zero **and negative values** will be treated as 0 (evict immediately) by the system." So even negative values like -1 actually have meaning.
2) Even if we intentionally deviated from the Kubernetes specification and said "negative numbers are disallowed/treated as null", the negative number would show up in the Terraform plan that is printed onscreen even when the user omits the tolerationSeconds value. If that user knows the above semantic meaning of negative numbers, they may get confused.

I actually had created a commit for this at https://github.com/chanzuckerberg/terraform-provider-kubernetes/commit/567a26787dc629c2b4997265dfa112892ce6f38f but haven't created a formal PR because I wasn't comfortable being the one to say "Terraform provider will intentionally deviate from the Kubernetes spec."

karlhungus: Yeah -- this sucks, but i think it's a little different than that, if negative treat it as kubernetes would (which is thankfully identical to 0), if zero don't allow it, and if positive continue along...

So far local tests for this seem to work

karlhungus: @mbarrien do you have some example tf i can test, so far i can't get this to fail?

I created a similar daemon set to your comment:

```
resource "kubernetes_daemonset" "test5" {
  metadata {
    name = "negative-seconds"

    labels {
      Test = "TfAcceptanceTest"
    }
  }

  spec {
    selector {
      match_labels {
        Test = "TfAcceptanceTest"
      }
    }

    template {
      metadata {
        labels {
          Test = "TfAcceptanceTest"
        }
      }

      spec {
        toleration {
          effect   = "NoSchedule"
          key      = "dedicated"
          operator = "Equal"
          value    = "kiam_master"
        }
        container {
          image = "redis:5.0.2"
          name  = "containername"
        }
      }
    }
  }
}
```

and got this yaml:

```
➭ kubectl get daemonsets -o yaml | grep toleration -A4
        tolerations:
        - effect: NoSchedule
          key: dedicated
          operator: Equal
          value: kiam_master
```

karlhungus:
@alexsomesan Wouldn't mind your opinion on this

kubernetes/schema_pod_spec.go
```diff
@@ -185,6 +185,42 @@ func podSpecFields(isUpdatable, isDeprecated, isComputed bool) map[string]*schem
...
+		"toleration": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "If specified, the pod's toleration. Optional: Defaults to empty",
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"effect": {
+						Type:        schema.TypeString,
+						Description: "Effect indicates the taint effect to match. Empty means match all taint effects. When specified, allowed values are NoSchedule, PreferNoSchedule and NoExecute.",
+						Optional:    true,
```
alexsomesan:
```suggestion
						Optional:    true,
						ValidateFunc: validation.StringInSlice({"NoSchedule, PreferNoSchedule and NoExecute}, false),

File: kubernetes/structures_pod.go (hunk 2 above)
Comment (alexsomesan) [reply kubernetes/structures_pod.go]: My suggestion for this issue would be this:

* make `toleration_seconds` a TypeString.
* use a distinctive string value like `unspecified` to distinguish between the 0 value and not-specified. Handle this special value in the flatten / expander.
* Convert to / from integer in the flatten / expand functions using `strconv.ParseInt` / `strconv.FormatInt`
* in the validation function, enforce the Kubernetes specified range of valid values as well as the placeholder value `unspecified`. 

This solution has precedent and has been used to work around this exact same issue in the past.

Comment (karlhungus) [reply kubernetes/structures_pod.go]: @alexsomesan I think this is addressed, not totally sure i got you right - now you must specify "unspecified" or a valid value; maybe i miss interpreted you though. 

This also unfortunatly only errors out during apply, ideally this would fail during plan:

```
Error: Error applying plan:

1 error(s) occurred:

* kubernetes_daemonset.test5: 1 error(s) occurred:

* kubernetes_daemonset.test5: invalid toleration_seconds must be int or "unspecified", got "":


```

mbarrien: There's always making blank string "" be the unspecified value. It does have the property of being distinguishable from 0.

mbarrien:
Just looks like it needs a quick unit test update to remove one final reference of the old string "unspecified" (see failing Travis test). After that, probably good for merging.

karlhungus:
@mbarrien thanks for the ping - i appear to have been easily distractible

karlhungus:
@alexsomesan I think this is ready for review

pcj:
@alexsomesan Gentle ping.  Looking forward to seeing this cross the finish line.

JordanP:
I'd love to see this merged too.

karlhungus:
I'm going to close this - if someone else wants to take it up feel free!

mbarrien:
What was left to do on this? It's been working fine as is in my side branch.

karlhungus:
I think just approval/conflict resolution

diclophis:
Is this merged into the next release or not, it seems like an mandatory feature for this module?

tdmalone:
@karlhungus Would you mind re-opening this, if it hasn't been merged yet? It's not likely to be noticed as closed.

karlhungus:
Not at all, but i think @shcallaway's got this one https://github.com/terraform-providers/terraform-provider-kubernetes/pull/448

Issue #240, opened 2018-12-04
secret_key_ref should have optional field


andyvanosdale:
Looking at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#secretkeyselector-v1-core, the `secret_key_ref` field should have an `optional` boolean field that specifies whether the Secret or it's key must be defined.

### Terraform Version
Terraform v0.11.10
+ provider.kubernetes v1.4.0

### Affected Resource(s)
- kubernetes_pod
- kubernetes_deployment
- kubernetes_replication_controller

### Expected Behavior
No error. Deployment succeeds and marks the reference as optional with Kubernetes.

### Actual Behavior
`Error: kubernetes_pod.pod:spec.0.container.0.env.0.value_from.0.secret_key_ref.0: invalid or unknown key: optional`

### Sample HCL
```hcl
resource "kuberenetes_pod" "pod" {
...
  spec {
    container {
      env {
        name = "IM_OPTIONAL"
        value_from {
          secret_key_ref {
            name = "my-secrets"
            key = "the-secret"
            optional = true
          }
        }
      }
    }
  }
}
```

### Steps to Reproduce
1. `terraform apply`

hashibot:
This issue has been open 180 days with no activity. If this issue is reproducible with the latest version of the provider and with Terraform 0.12, please comment. Otherwise this issue will be closed in 30 days.

mbarrien:
Open PR awaiting review. Still valid.

dak1n1:
Hi, thanks for submitting this report. I've started working on some similar issues and consolidated them all here: https://github.com/hashicorp/terraform-provider-kubernetes/issues/892

I'll close this issue in favor of the consolidated one.

PR #295, opened 2019-02-01
add optional secret_key_ref

karlhungus:
#### What this does

1. Add [`optional`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#secretkeyselector-v1-core) for secret_key_ref
~~1. Add [`divisor`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#resourcefieldselector-v1-core) field for resource_field_ref~~

Fix: https://github.com/terraform-providers/terraform-provider-kubernetes/issues/240

### DIFF
--- a/kubernetes/resource_kubernetes_pod_test.go
+++ b/kubernetes/resource_kubernetes_pod_test.go
@@ -43,6 +43,8 @@ func TestAccKubernetesPod_basic(t *testing.T) {
 					resource.TestCheckResourceAttrSet("kubernetes_pod.test", "metadata.0.self_link"),
 					resource.TestCheckResourceAttrSet("kubernetes_pod.test", "metadata.0.uid"),
 					resource.TestCheckResourceAttr("kubernetes_pod.test", "spec.0.container.0.env.0.value_from.0.secret_key_ref.0.name", secretName),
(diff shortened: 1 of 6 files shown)


karlhungus:
(I didn't actually need divisor, but i thought while i was in there...)

alexsomesan:
Thanks for contributing this fix!

There is a bit more to be done for this change to be functional and complete. You must also add these new attributes to the relevant expander / flattener functions. The role of these functions is to translate the attribute values between Terraform's internal representation and the Kubernetes client's own representation.

More specifically, the attribute `optional` must be handled in the functions `expandSecretKeyRef` and `flattenSecretKeyRef`.
The same way, the attribute `divisor` must be handled in both `flattenResourceFieldSelector` and `expandResourceFieldRef`.

Have a look at how existing attributes are processed to get a sense what is adequate for these new ones.

karlhungus:
@alexsomesan Hey, I intended to test this locally; and admit i should have known better having spent tie on  https://github.com/terraform-providers/terraform-provider-kubernetes/pull/294, I'll re-ping when this is ready

karlhungus:
@alexsomesan All right, this looks to pass on my acceptance tests. I removed the divisor bit because it looked bigger and I don't personally need it yet.

PR #326, opened 2019-02-20
Extended example usage on how to include config files into config maps

AlexBoulyga:
Not a major documentation update at all,
but for new users this example might prove useful, to see that it is so simple to include config files into config maps


AlexBoulyga:
This change is partially due to this issue:
https://github.com/terraform-providers/terraform-provider-kubernetes/issues/216
Where the solution was to pipe the file. In my opinion the solution presented by me is easier to use, and cleaner.
Please correct me if I am wrong

PR #333, opened 2019-02-22
EKS test infra: wait for API to be ready

alexsomesan:
This change introduces a loop-check for the K8s API to become ready before proceeding to apply the node auth ConfigMap. 
This ensures reliable builds within the same Terraform run which is nice to have in CI.

### DIFF
--- a/kubernetes/test-infra/eks/node-config/config.tf
+++ b/kubernetes/test-infra/eks/node-config/config.tf
@@ -18,9 +18,31 @@ locals {
 MAPPEDROLE
 }
 
+resource "local_file" "cluster_ca" {
+  content  = "${base64decode(var.cluster_ca)}"
+  filename = "${path.module}/cluster_ca"
+}
+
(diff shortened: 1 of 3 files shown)


kubernetes/test-infra/eks/node-config/config.tf
```diff
@@ -18,9 +18,31 @@ locals {
...
+  filename = "${path.module}/cluster_ca"
+}
+
+resource "null_resource" "wait_for_api" {
+  depends_on = ["local_file.cluster_ca"]
+
+  provisioner "local-exec" {
+    command = <<CMDEOF
+while ! curl -s --cacert ${local_file.cluster_ca.filename} ${var.cluster_endpoint}/version > /dev/null; do 
+  echo "Waiting for the cluster API to come online..."
```
tombuildsstuff: @bflad I think this could be a bug in the EKS resource?

PR #349, opened 2019-03-05
Fix service account secrets parsing

pablo-ruth:
Fix two errors introduced by PR #281

### DIFF
--- a/kubernetes/resource_kubernetes_service_account.go
+++ b/kubernetes/resource_kubernetes_service_account.go
@@ -106,12 +106,16 @@ func resourceKubernetesServiceAccountCreate(d *schema.ResourceData, meta interfa
 		for _, secret := range secretList.Items {
 			for _, svcSecret := range diff {
 				if secret.Name != svcSecret.Name {
-					break
+					continue
 				}
 				if secret.Type == api.SecretTypeServiceAccountToken {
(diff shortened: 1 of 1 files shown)


kubernetes/resource_kubernetes_service_account.go
```diff
@@ -106,14 +106,18 @@ func resourceKubernetesServiceAccountCreate(d *schema.ResourceData, meta interfa
 		for _, secret := range secretList.Items {
 			for _, svcSecret := range diff {
 				if secret.Name != svcSecret.Name {
-					break
+					continue
```
pablo-ruth: Here we want to continue to iterate over secrets to find the right one instead of breaking loop on first fail

kubernetes/resource_kubernetes_service_account.go
```diff
@@ -106,14 +106,18 @@ func resourceKubernetesServiceAccountCreate(d *schema.ResourceData, meta interfa
...
 				}
 				if secret.Type == api.SecretTypeServiceAccountToken {
 					svcAccTokens = append(svcAccTokens, secret)
 				}
 			}
 		}
 
+		if len(svcAccTokens) == 0 {
```
pablo-ruth: I missed to handle the case where there is secrets available but not the service account token

PR #355, opened 2019-03-09
Update link to Kubernetes blog for reference

sjentzsch:


alexsomesan:
Hi @sjentzsch! Thanks a lot for spotting this and posting the fix.
I did a quick search for that URL and it turns out there is one more mention of it in the datasource docs `website/docs/d/storage_class.html.markdown`. Would you mind updating it over there as well? Thanks!

alexsomesan:
Nevermind. I think it's faster if I make the change myself and save you the back-and-forth.

PR #357, opened 2019-03-11
Remove ValidateFunc functions on complex types

### LINKED ISSUE #34: Variables not correctly interpreted when configuring labels on a namespace
When working with the kubernetes namespace resource, I ran into an issue where it seems that variables that are fetched from remote state are not interpreted and seem to be treated as literals.
Only the labels seem to be affected, the `name` property does not have an issue.

### Terraform Version
This is with terraform version 0.10-rc1

### Affected Resource(s)
The labels within the kubernetes namespace

```hcl
data "terraform_remote_state" "globalvars" {
  backend = "local"
  config {
    path = "path/to/terraform.tfstate"
  }
}
resource "kubernetes_namespace" "k8s_namespace" {
  metadata {
    labels {
      application = "${data.terraform_remote_state.globalvars.application_name}"
    }
    name = "${data.terraform_remote_state.globalvars.application_name}"
  }
}
```
The remote state is filled through:
```hcl
output "application_name" { value = "myapp" }
```

### Debug Output
module.kubernetes-namespace.kubernetes_namespace.k8s_namespace: metadata.0.labels ("${data.terraform_remote_state.globalvars.application_name}") a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')

### Expected Behavior
Label should be set to `application = myapp`

### Actual Behavior
Get an error stating I try to create an invalid label (see debug output)

### Steps to Reproduce
First run code that outputs variable in a state file
Then create a tf file that includes the kubernetes namespace as a module

### Important Factoids
If you replace the variable with a local one (${var.application_name}) then there is no issue.
If you set the label to a hardcoded value, you'll see it works with interpolation for `name`.
I also tried to wrap the variable in a function (substr/length/chomp) but the error remains.
Could it be that it doesn't recognize the remote state variable as a valid variable?

### References
The globalvars trick as described in https://github.com/hashicorp/terraform/issues/5480 is used here.

pablo-ruth:
As said in [Terraform documentation](https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L220), ValidateFunc currently only works for primitive types. So we should not use it on complex types used by "annotations" and "labels" as it prevents the use of interpolated values on them.

Fix #34

### DIFF
--- a/kubernetes/schema_metadata.go
+++ b/kubernetes/schema_metadata.go
@@ -9,11 +9,10 @@ import (
 func metadataFields(objectName string) map[string]*schema.Schema {
 	return map[string]*schema.Schema{
 		"annotations": {
-			Type:         schema.TypeMap,
-			Description:  fmt.Sprintf("An unstructured key value map stored with the %s that may be used to store arbitrary metadata. More info: http://kubernetes.io/docs/user-guide/annotations", objectName),
-			Optional:     true,
-			Elem:         &schema.Schema{Type: schema.TypeString},
-			ValidateFunc: validateAnnotations,
+			Type:        schema.TypeMap,
+			Description: fmt.Sprintf("An unstructured key value map stored with the %s that may be used to store arbitrary metadata. More info: http://kubernetes.io/docs/user-guide/annotations", objectName),
+			Optional:    true,
+			Elem:        &schema.Schema{Type: schema.TypeString},
 		},
 		"generation": {
 			Type:        schema.TypeInt,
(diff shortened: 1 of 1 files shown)


alexsomesan:
Hi @pablo-ruth.

Validation for map attributes is in fact available and working. It was added to Terraform a while ago, via https://github.com/hashicorp/terraform/pull/3505. Unfortunately the documentation lagged behind, but it's on it's way to being fixed via https://github.com/hashicorp/terraform/pull/20851

Further proof of that is that this change actually causes an acceptance test which checks this validation to fail.
```
------- Stdout: -------
=== RUN   TestAccKubernetesNamespace_invalidLabelValueType
--- FAIL: TestAccKubernetesNamespace_invalidLabelValueType (6.57s)
    testing.go:519: Step 0, no error received, but expected a match to:
        
        Expected value to be string
   
FAIL
```

That said I'm going to close this PR and have a fresh look at the root cause of https://github.com/terraform-providers/terraform-provider-kubernetes/issues/34

PR #370, opened 2019-03-22
Test infra GKE: auto-detect available cluster versions

alexsomesan:
This change allows the GKE test infra template to query and pick the latest available Kubernetes version in that specific region.
By setting the `kubernetes_version` Terraform variable to a partial or specific GKE version the selected version can be controlled precisely.
This allows the templates to more robustly function in CI environments where daily runs have to keep up with the rolling releases of GKE.


alexsomesan:
Yes, tested in TC and locally.

PR #311, opened 2019-02-13
Possible fix for deployment strategy

xmackex:
Possible fix for #260
I'm not a experienced Go developer, bu I needed `strategy` to work properly.
This fix works for me, but was not able to make any verbose testing.

Issue #293, opened 2019-02-01
kubernetes_daemonset with strategy type=OnDelete has error "expected pointer, but got nil"


mbarrien:
Tried running from the head to test the upcoming daemonset support, ran into an issue with the update strategy of `OnDelete`. The strategy block has an optional field `rolling_update` that should only be filled in if the strategy is `RollingUpdate`, which this is not. If you omit it as is correct in this case, provider returns an error message.

### Terraform Version
Terraform v0.11.8, self-built kubernetes provider from head cef8ec04

### Affected Resource(s)
- kubernetes_daemonset

### Terraform Configuration Files
Very stripped down; won't actually launch as is, but will manifest the bug
```hcl
resource "kubernetes_daemonset" "kiam-agent" {
  metadata {
    labels {
      app = "kiam"
    }
    name = "kiam-agent"
  }

  spec {
    selector {
      match_labels {
        app  = "kiam"
      }
    }

    template {
      metadata {
        labels {
          app = "kiam"
        }
      }
      spec {
        container {
          name = "kiam-agent"
          image = "quay.io/uswitch/kiam:v3-release"
        }
      }
    }

    strategy {
      type  = "OnDelete"
    }
  }
}
```

### Expected Behavior
DaemonSet successfully creates.

### Actual Behavior
```
* module.k8s-core.kubernetes_daemonset.kiam-agent: 1 error(s) occurred:

* kubernetes_daemonset.kiam-agent: Failed to create daemonset: DaemonSet in version "v1" cannot be handled as a DaemonSet: expected pointer, but got nil
```

### Steps to Reproduce
terraform apply with a kubernetes_daemonset resource.

### Workaround
Make the strategy block say the following, explicitly setting an empty map for rolling_update:
```hcl

    strategy {
      type           = "OnDelete"
      rolling_update = {}
    }
```

hashibot:
This issue has been open 180 days with no activity. If this issue is reproducible with the latest version of the provider and with Terraform 0.12, please comment. Otherwise this issue will be closed in 30 days.

Issue #260, opened 2018-12-18
Kubernetes deployment: setting strategy.type not possible 


jeff1985:
I have a kubernetes deployment where I need to set the strategy type to "Recreate". See https://www.terraform.io/docs/providers/kubernetes/r/deployment.html#strategy 

I have specified the definition as following:
```
resource "kubernetes_deployment" "my-sftp" {
  metadata {
    name = "sftp"
    namespace = "${kubernetes_namespace.stage.metadata.0.name}"

    labels {
      app  = "my-sftp"
    }
  }


  spec {
    replicas = 1

    # because we are using azure disks, we can not attach the storage at the same time to multiple containers

    strategy {
      type = "Recreate" 
    }


    selector {
      match_labels {
        app = "my-sftp"
      }
    }

    template {
      metadata {
        name = "my-sftp"

        labels {
          app  = "my-sftp"
        }
      }

      spec {

        container {
          image = "${var.registry_host}/sftp-server:v1"
       ...
```

When I execute terraform, it recognizes a change is needed:
```
~ module.k8s-alpha.kubernetes_deployment.my-sftp
      spec.0.strategy.#:                "0" => "1"
      spec.0.strategy.0.type:           "" => "Recreate"
```

After execution the field is however not changed, which I can verify by displaying yaml definition directly in kubernetes. Also subsequent executions will try to update the field with no effect. The same is happening, if I try to edit rolling_update{} section.
 

### Terraform Version
```
Terraform v0.11.7
+ provider.azurerm v1.17.0
+ provider.kubernetes v1.4.0
+ provider.random v2.0.0
```

xpros:
Seeing same behavior with a slightly newer version of terraform:
```
Terraform v0.11.10
+ provider.kubernetes v1.4.0
```

ghamed:
I can also confirm same issue

```Terraform v0.11.11
+ provider.google v1.20.0
+ provider.kubernetes v1.4.0
```

Code snippet
```resource "kubernetes_deployment" "myapp" {
  metadata {
    name = "myapp"
    labels {
      app = "myapp"
    }
  }
  spec {
    replicas = 1
    strategy {
      type = "RollingUpdate"
      rolling_update {
        max_surge = "1"
        max_unavailable = "1"
      }
    }
...
```

I hope this helps.

jeff1985:
This issue is breaking my deployment on azure kubernetes as disks can only be bound to one container.

[...]

PR #372, opened 2019-03-22
fix update strategy for deployment

campbel:


campbel:
This is a fix to address https://github.com/terraform-providers/terraform-provider-kubernetes/issues/260

campbel:
All current unit tests are passing for me. Would like to write a test to cover this issue if anyone could provide guidance. Thank you.

campbel:
There was a bit more to do to read the correct state and set max_surge and max_unavailable. I manually tested 4 scenarios. 

1. setting no strategy, 
2. setting strategy type recreate, 
3. setting strategy type rolling update with max surge/unavailable = 25%
4. setting strategy type rolling update with max max surge / unavailable = 50%. 

They all correctly set the value and showed no changes on subsequent runs. Setting no value followed by explicitly setting the defaults also correctly saw no change.

I'll look into adding unit tests, if anyone is available to review I would appreciate it. Thanks!

pasali:
There is a another PR(#311) for this issue, waiting for review.

campbel:
I think this PR and the other are duplicates. Didn't see that one beforehand. I believe computed was taken from the Daemonset implementation or me just fiddling with settings. I can test without if needed, but happy to defer to the other PR.

Thanks!

xmackex:
@campbel I've made [PR](https://github.com/terraform-providers/terraform-provider-kubernetes/pull/311) with same functionality (have not took a look at yours). Anyway will be happy to take a look so we could find the best solution. 

About `Computed` field, same as you I've taken it from Daemonset (or somewhere else)

Note, as I've mentioned I'm not very experienced Go developer so some stuff may be not very good

kubernetes/structures_deployment.go
```diff
@@ -90,40 +114,37 @@ func expandPodTemplate(l []interface{}) (*corev1.PodTemplateSpec, error) {
...
-		return appsv1.DeploymentStrategy{}
+		obj.Type = appsv1.RollingUpdateDeploymentStrategyType
+		return obj
 	}
 	in := l[0].(map[string]interface{})
-	obj := appsv1.DeploymentStrategy{}
-	if v, ok := in["type"].(map[string]interface{}); ok && len(v) > 0 {
-		obj.Type = appsv1.DeploymentStrategyType(in["type"].(string))
+
+	if v, ok := in["type"]; ok {
```
xmackex: Probably we should check `in["type"]` if it is `string` and length of `v`
Like
```
if v, ok := in["type"].(string); ok && len(v) > 0 {
```

kubernetes/structures_deployment.go
```diff
@@ -90,40 +114,37 @@ func expandPodTemplate(l []interface{}) (*corev1.PodTemplateSpec, error) {
...
+		obj.Type = appsv1.DeploymentStrategyType(v.(string))
 	}
-	if v, ok := in["rolling_update"].([]interface{}); ok && len(v) > 0 {
-		obj.RollingUpdate = expandRollingUpdateDeployment(v)
+	if v, ok := in["rolling_update"]; ok {
```
xmackex: Probably we should check `in["type"]` if it is `string` and length of `v`
Like
```
if v, ok := in["rolling_update"].([]interface{}); ok && len(v) > 0 {
```

kubernetes/structures_deployment.go
pranas: This could be reduced to

```
if v, _ := in["type"].(string); v != "" {
```

campbel: So one difference of checking `ok` instead of `v != ""` is in the scenario below:
```
strategy {
  type = "RollingUpdate"

  rolling_update {
    max_surge       = ""
    max_unavailable = ""
  }
}
```
You'll get this error message
```
kubernetes_deployment.scenario5: Failed to update deployment: Deployment.apps "scenario5" is invalid: [spec.strategy.rollingUpdate.maxUnavailable: Invalid value: intstr.IntOrString{Type:1, IntVal:0, StrVal:""}: a valid percent string must be a numeric string followed by an ending '%' (e.g. '1%',  or '93%', regex used for validation is '[0-9]+%'), spec.strategy.rollingUpdate.maxSurge: Invalid value: intstr.IntOrString{Type:1, IntVal:0, StrVal:""}: a valid percent string must be a numeric string followed by an ending '%' (e.g. '1%',  or '93%', regex used for validation is '[0-9]+%'), spec.strategy.rollingUpdate.maxUnavailable: Invalid value: intstr.IntOrString{Type:1, IntVal:0, StrVal:""}: may not be 0 when `maxSurge` is 0]
```

if we check `v != ""` then we'll silently fail to set the `max_surge` and `max_unavailable` which will cause subsequent applies to identify a change like below:
```
~ kubernetes_deployment.default
      spec.0.strategy.0.rolling_update.0.max_surge:       "25%" => ""
      spec.0.strategy.0.rolling_update.0.max_unavailable: "25%" => ""
```
I don't have a strong preference other than to be consistent with other resources.

What do you think?

kubernetes/structures_deployment.go
campbel: I don't think there is any practical benefit to doing this check because `in["rolling_update"]` will either be empty or a slice of length 1 per the schema. Am I missing something?

campbel:
@alexsomesan 
> Also, one think I asked the author of the other one is: why was it necessary to mark the strategy and sub-attributes as `Computed`?

If the values aren't computed, when `strategy` or `rolling_update` aren't specified you'll get
```
~ kubernetes_deployment.default
      spec.0.strategy.#:                  "1" => "0"
      spec.0.strategy.0.rolling_update.#: "1" => "0"
```
or
```
~ kubernetes_deployment.default
      spec.0.strategy.0.rolling_update.#: "1" => "0"
```
On every apply.

kubernetes/structures_deployment.go
xmackex: In this case `change` message can be more readable
```
~ kubernetes_deployment.default
      spec.0.strategy.0.rolling_update.0.max_surge:       "25%" => ""
      spec.0.strategy.0.rolling_update.0.max_unavailable: "25%" => ""
```
Will provide more info on why `change` occurs.

kubernetes/structures_deployment.go
xmackex: You are right, my bad. Thank you for figuring out.
Only reason I see is to be consistent with other resources.

campbel:
after applying suggested changes I'm now seeing this issue https://github.com/terraform-providers/terraform-provider-kubernetes/issues/293

will try to fix

campbel:
https://github.com/terraform-providers/terraform-provider-kubernetes/issues/293 should now be fixed for both deployment and daemonset

kubernetes/structures_deployment.go
campbel: I ended up needing this for the scenario where `rolling_update` is missing and `type` is specified. Turns out there is a practical benefit :)! Thanks for the suggestion

alexsomesan:
One acceptance test is failing:

```
------- Stdout: -------
=== RUN   TestAccKubernetesDeployment_basic
--- FAIL: TestAccKubernetesDeployment_basic (0.55s)
    testing.go:538: Step 0 error: Check failed: 2 errors occurred:
        	* Check 19/20 error: kubernetes_deployment.test: Attribute 'spec.0.strategy.0.rolling_update.0.max.surge' not found
        	* Check 20/20 error: kubernetes_deployment.test: Attribute 'spec.0.strategy.0.rolling_update.0.max.unavailable' not found
        
        
FAIL
````

PR #375, opened 2019-03-25
Add support for reading service account tokens

wjam:
Although the kubernetes_secret data resource is capable reading service account tokens, this tends to run into problems such as being unable to reference the `ca.crt` value or having Terraform plan which both creates a service account and attempts to read the associated service account token.

### DIFF
--- a/kubernetes/data_source_kubernetes_service_account_token.go
+++ b/kubernetes/data_source_kubernetes_service_account_token.go
@@ -1,0 +1,88 @@
+package kubernetes
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+	core_v1 "k8s.io/api/core/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"log"
+)
+
+func dataSourceKubernetesServiceAccountToken() *schema.Resource {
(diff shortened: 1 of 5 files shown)


wjam:
This would provide a work around for #334 as well as provide a solution to my own problem where I'm unable to have a single plan which will:
1. Create a service account
2. Read the service account token for the account that was just created
3. Generate a template using that token for use in another resource such as an S3 object or cloud-init.

I found that attempting to reference `data.kubernetes_secret.service_account.data.token` led to problems running `terraform plan` if the service account hadn't been created yet and `data.kubernetes_secret.service_account.data["token"]` led to problems when attempting to run `terraform apply`

alexsomesan:
Hi @wjam,

Can you reference here some issues related to the problems you mentioned while fetching tokens via the existing `kubernetes_secret` datasource?

alexsomesan:
@wjam That is a strange error. I will look into it.

According to my testing, you should be able to access the `token` field by using this syntax:
```
output "demo" {
  value = "${lookup(data.kubernetes_secret.foo.data, "token")}"
}
```
Can you confirm whether that works for you?

wjam:
Yes, that does appear to be a workaround for my issue - although I haven't had a chance to run `terraform apply` on the codebase that is causing the issue.

I'll close this PR under the assumption that the workaround is sufficient for the moment.

PR #387, opened 2019-04-05
Update automount_service_account_token attribute to support unsetting the attribute

wjam:


wjam:
While this PR is an attempt to correctly support the automount_service_account_token attribute, I accept that it probably isn't suitable in its current form as the change of `automount_service_account_token` from boolean to string isn't backwards compatible.

The purpose of this PR is mainly to try and start a discussion around what is possible/acceptable. The problem is trying to determine whether the `false` value being seen is from the default boolean value, so unset meaning the field in the SA should be updated to nil, or from the attribute actually being set to false, meaning the field in the SA should be updated to false.

wjam:
@alexsomesan Any opinion?

wjam:
@alexsomesan Thoughts?

PR #396, opened 2019-04-12
Add support of client-go credential plugins in auth

### LINKED ISSUE #161: Feature Request: Support exec Authentication (client-go Credential Plugins)

dh-harald:
Closes #161

### DIFF
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -101,6 +101,34 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("KUBE_LOAD_CONFIG_FILE", true),
 				Description: "Load local kubeconfig.",
 			},
+			"exec": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"api_version": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"command": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"env": {
+							Type:     schema.TypeMap,
+							Optional: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+						},
+						"args": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+						},
+					},
+				},
+				Description: "",
+			},
 		},
 
 		DataSourcesMap: map[string]*schema.Resource{
@@ -179,6 +207,21 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	}
 	if v, ok := d.GetOk("token"); ok {
 		cfg.BearerToken = v.(string)
+	}
+
+	if v, ok := d.GetOk("exec"); ok {
+		exec := &clientcmdapi.ExecConfig{}
+		if spec, ok := v.([]interface{})[0].(map[string]interface{}); ok {
+			exec.APIVersion = spec["api_version"].(string)
+			exec.Command = spec["command"].(string)
+			exec.Args = expandStringSlice(spec["args"].([]interface{}))
+			for kk, vv := range spec["env"].(map[string]interface{}) {
+				exec.Env = append(exec.Env, clientcmdapi.ExecEnvVar{Name: kk, Value: vv.(string)})
+			}
+		} else {
+			return nil, fmt.Errorf("Failed to parse exec")
+		}
+		cfg.ExecProvider = exec
 	}
 
 	k, err := kubernetes.NewForConfig(cfg)
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -109,4 +109,8 @@ The following arguments are supported:
 * `config_context_cluster` - (Optional) Cluster context of the kube config (name of the kubeconfig cluster, `--cluster` flag in `kubectl`). Can be sourced from `KUBE_CTX_CLUSTER`.
 * `token` - (Optional) Token of your service account.  Can be sourced from `KUBE_TOKEN`.
 * `load_config_file` - (Optional) By default the local config (~/.kube/config) is loaded when you use this provider. This option at false disable this behaviour. Can be sourced from `KUBE_LOAD_CONFIG_FILE`.
-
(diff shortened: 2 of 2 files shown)


bflad:
This looks like this would close https://github.com/terraform-providers/terraform-provider-kubernetes/issues/161 so I have edited the PR description to automatically close it. I cannot provide a full review of this enhancement, but can provide some review of the Go/Terraform Provider SDK code. 👍

kubernetes/provider.go
```diff
@@ -101,6 +101,33 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("KUBE_LOAD_CONFIG_FILE", true),
 				Description: "Load local kubeconfig.",
 			},
+			"exec": {
+				Type:     schema.TypeSet,
```
bflad: With `MaxItems: 1` configurations blocks, it is generally preferred to use `Type: schema.TypeList` to simplify handling. 👍

kubernetes/provider.go
```diff
@@ -101,6 +101,33 @@ func Provider() terraform.ResourceProvider {
...
+						"api_version": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"command": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"env": {
+							Type:     schema.TypeMap,
```
bflad: While the Terraform Provider SDK does not explicitly error when not providing the `Elem` for `Type: schema.TypeMap`, we generally recommend adding it, e.g. `Elem: &schema.Schema{Type: schema.TypeString},`

kubernetes/provider.go
alexsomesan: Can you please make this a TypeList instead?
I've found that in Terraform 0.12+ a TypeList works best for modeling a complex block, such as this.

website/docs/index.html.markdown
```diff
@@ -109,4 +109,4 @@ The following arguments are supported:
 * `config_context_cluster` - (Optional) Cluster context of the kube config (name of the kubeconfig cluster, `--cluster` flag in `kubectl`). Can be sourced from `KUBE_CTX_CLUSTER`.
 * `token` - (Optional) Token of your service account.  Can be sourced from `KUBE_TOKEN`.
 * `load_config_file` - (Optional) By default the local config (~/.kube/config) is loaded when you use this provider. This option at false disable this behaviour. Can be sourced from `KUBE_LOAD_CONFIG_FILE`.
-
+* `exec` - (Optional) Exec-based client auth provider (https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins)
```
bflad: Since we are adding a new configuration block with nested arguments, we should appropriately document everything here. 👍 

```suggestion
* `exec` - (Optional) Configuration block to use an [exec-based credential plugin](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins), e.g. call an external command to receive user credentials. 
  * `api_version` - (Required) API version to use when decoding the ExecCredentials resource, e.g. `client.authentication.k8s.io/v1beta1`.
  * `command` - (Required) Command to execute.
  * `args` - (Optional) List of arguments to pass when executing the plugin.
  * `env` - (Optional) Map of environment variables to set when executing the plugin.
```

kubernetes/provider.go
```diff
@@ -181,6 +208,18 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		cfg.BearerToken = v.(string)
 	}
 
+	if v, ok := d.GetOk("exec"); ok {
+		exec := &clientcmdapi.ExecConfig{}
+		spec := v.(*schema.Set).List()[0].(map[string]interface{})
```
alexsomesan: Also, would you mind breaking down this line into multiple steps and check for errors and nils on the type assertions. This looks like it has the potential to panic.

dh-harald: In theory, it should be pass by the schema checker, but I've added an extra check to avoid the panic

kubernetes/provider.go
```diff
@@ -101,6 +101,34 @@ func Provider() terraform.ResourceProvider {
 				DefaultFunc: schema.EnvDefaultFunc("KUBE_LOAD_CONFIG_FILE", true),
 				Description: "Load local kubeconfig.",
 			},
+			"exec": {
+				Type:     schema.TypeList,
```
chancez: Why wasn't this done as `TypeMap`?

chancez:
Is this in a release? The code is in available at least in 1.10.0, but I'm getting:
```
Error: Unsupported argument

  on ../../modules/eks-cluster/main.tf line 70, in provider "kubernetes":
  70:   exec = [{

An argument named "exec" is not expected here. Did you mean to define a block
of type "exec"?
```

chancez:
Ah, it's a block. I see.

```
  exec {
    api_version = "client.authentication.k8s.io/v1alpha1"
    command     = "aws-iam-authenticator"
    args        = concat(["token", "-i", data.aws_eks_cluster.cluster.name], local.authenticator_additional_args)
    env         = local.authenticator_env_vars
  }

```

ballad89:
@chancez did the block above work for you ? what versions are you using ?
```
$ terraform version                                                                                         ✔  5506  17:22:08
Terraform v0.12.24
+ provider.aws v2.57.0
+ provider.external v1.2.0
+ provider.helm v1.1.1
+ provider.kubernetes v1.11.1
+ provider.local v1.4.0
+ provider.null v2.1.2
+ provider.random v2.2.1
+ provider.template v2.1.2
```

chancez:
It did.
```
$ terraform version
Terraform v0.12.24
+ provider.aws v2.54.0
+ provider.kubernetes v1.10.0
+ provider.local v1.4.0
+ provider.null v2.1.2
+ provider.random v2.2.1
+ provider.template v2.1.2
```

PR #398, opened 2019-04-13
Check if Generation is newer than the ObservedGeneration

jpreese:
When checking to see whether or not a deployment has succeeded, we should be checking if the generation as defined in the spec is newer than the last observed generation.

In its current form, the deployment is created and then immediately read from. This is resulting in every property (UpdatedReplicas, AvailableReplicas, etc) on the Status response to be zero, causing the `waitForDeploymentReplicasFunc()` to immediately return without waiting for the pods to become ready.

### DIFF
--- a/kubernetes/resource_kubernetes_deployment.go
+++ b/kubernetes/resource_kubernetes_deployment.go
@@ -367,7 +367,7 @@ func waitForDeploymentReplicasFunc(conn *kubernetes.Clientset, ns, name string) 
 			return resource.NonRetryableError(err)
 		}
 
-		if dply.Generation <= dply.Status.ObservedGeneration {
+		if dply.Generation >= dply.Status.ObservedGeneration {
 			cond := GetDeploymentCondition(dply.Status, appsv1.DeploymentProgressing)
 			if cond != nil && cond.Reason == TimedOutReason {
 				err := fmt.Errorf("Deployment exceeded its progress deadline")


jpreese:
Just checking in to see if there's any feedback in regards to this. Right now we're just sleeping between kubernetes_deployment resources, so there is a workaround. Though this seems small enough to merge in if it looks OK.

alexsomesan:
Does look like the right thing to do, on first glance.
I'll test it soon.

jpreese:
Sounds good @alexsomesan, thanks for looking into it!

jpreese:
Just wanted to ping on this as it's been awhile. Let me know if there's anything I can do.!

jpreese:
Any updates on this @alexsomesan ?

jpreese:
Just checking in again @alexsomesan. Let me know if there's anything I can do to help.

jpreese:
@alexsomesan ?

aareet:
Hi @jpreese thank you for your contributions and for keeping this PR alive. Just wanted to drop in and mention that this PR is in our queue for triage and review, so no need to ping us. It follows our Pull Request Lifecycle - https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/_about/PULL_REQUESTS.md

jpreese:
Apologies. I thought it got lost! It's been over a year and is a single character change. I'll keep an eye out!

jrhouston:
This was fixed in https://github.com/hashicorp/terraform-provider-kubernetes/commit/51c1bbd1f57f355817c97a8ea857c81fdaee2dc5

PR #407, opened 2019-04-19
kubernetes_service: Add support for externalTrafficPolicy field

dnikolai:


kubernetes/resource_kubernetes_service.go
```diff
@@ -65,6 +65,11 @@ func resourceKubernetesService() *schema.Resource {
 							Elem:        &schema.Schema{Type: schema.TypeString},
 							Set:         schema.HashString,
 						},
+						"external_traffic_policy": {
+							Type:        schema.TypeString,
+							Description: "Only applies to `type = NodePort` or `type = LoadBalancer`. Defines the policy for external traffic. Can be Cluster (default) or Local. Local adds proxy rules only for pods that exist on the same node.",
+							Optional:    true,
```
JordanP: @dnikolai It would be great to rebase this for good measure and add :+1: 

```
ValidateFunc: validation.StringInSlice([]string{"Cluster", "Local"}, false),
Default: "Cluster",
"
here

JordanP:
@dnikolai also, could you add a line in the documentation file https://github.com/terraform-providers/terraform-provider-kubernetes/blob/master/website/docs/r/service.html.markdown#spec

dnikolai:
Hi @JordanP, thanks for your comments.

There is no sense in doing that as support of external traffic policy was added in version v1.7.0

JordanP:
So ? Having code that enforces/reflects what the doc string says for sanity check is beneficial, I mean that's basic user input validation.

aareet:
Closing based on the above comments.

externalTrafficPolicy is implemented https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/kubernetes/resource_kubernetes_service.go

PR #417, opened 2019-04-26
Ingress Resource (closes #14)

### LINKED ISSUE #14: Feature request: Ingress resource
Hello

There is lack of [ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) resources support in Terraform. Please can you add it?

Regards
holoGDM

mbarrien:
More extraction of @sl1pm4t features, this time for the kubernetes_ingress resource. Rebased against 1.6.2 branch. Closes #14. Given the possibility of supporting some beta resources via #378, this should be something to consider merge in line with the new policy.

### DIFF
--- a/_examples/ingress/main.tf
+++ b/_examples/ingress/main.tf
@@ -1,0 +1,164 @@
+provider "kubernetes" {
+  config_context_auth_info = "minikube"
+  config_context_cluster   = "minikube"
+}
+
+resource "kubernetes_ingress" "example" {
+  metadata {
+    name = "example"
+
+    annotations {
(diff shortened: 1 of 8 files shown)


mdz:
> Second, small thing, but the `path_regex` attribute is actually called `path` in the Kubernetes API so it's be great if it followed that naming.

Not only that, the value of this attribute is not actually interpreted as a regex, at least in the gce ingress controller, so `path` would be less confusing!

alexsomesan:
I've made a bit of a rebasing mess here. Trying to fix it.

Issue #142, opened 2018-03-14
Allow config_path to be computed 


Ikradex:
I have a resource from a provider that creates a K8s cluster and another that downloads it's kube config. I can't seem to configure the provider with a `config_path` that will be computed at a later time ie. after the cluster is created.

### Terraform Version
v0.11.3

### Affected Resource(s)
- provider kubernetes

### Terraform Configuration Files
I've simplified the example below but this is the flow:

```hcl
resource "container_cluster" "test_cluster" {
  name         = "${var.cluster_name}"
  datacenter   = "${var.datacenter}"
  machine_type = "${var.machine_type}"
  public_vlan_id = "${var.public_vlan_id}"
  private_vlan_id = "${var.private_vlan_id}"
  no_subnet    = true
  workers = "${var.workers[var.num_workers]}"
}

data "container_cluster_config" "test_cluster_config" {
  depends_on = ["container_cluster.test_cluster"]
  cluster_name_id = "${container_cluster.test_cluster.name}"
}

provider "kubernetes" {
   config_path = "${data.container_cluster_config.test_cluster_config.config_file_path}"
}

resource "kubernetes_namespace" "test_namespace" {
  depends_on = ["data.container_cluster_config.test_cluster_config"]
  metadata {
    name = "test-ns"
  }
}
...
...
```

### Expected Behavior
Kubernetes provider should be instantiated and load `config_path` after the `data.container_cluster_config.test_cluster_config.config_file_path` is available

### Actual Behavior
Provider seems to be prematurely instantiating itself with no config.
```
$ terraform apply
...
...
Error: Error running plan: 1 error(s) occurred:

* provider.kubernetes: Failed to load config (; default context): invalid configuration: default cluster has no server defined
```

### Steps to Reproduce
The general flow above is the steps to reproduce. Basically try to instantiate the provider with a `config_path` that has yet to be generated ie. computed after the cluster resource is created.

### References
This is a slightly related issue: https://github.com/hashicorp/terraform/issues/2430

walterdolce:
I just experienced this in a similar scenario: I create a GKE cluster *first*, then I want to create a k8s resource within it (in this case s `kubernetes_secret`), but Terraform/the provider errors with:

```
module.core_application_sqlite_database_backup_generator_service_account_secrets.provider.kubernetes: Failed to load config (/path/to/my/user/.kube/config; overriden context; cluster: core-application): cluster "core-application" does not exist
```

[...]

PR #419, opened 2019-04-29
Allow specifying generated kubeconfig file

tennix:
This PR fixes https://github.com/terraform-providers/terraform-provider-kubernetes/issues/142. When failed to load kubeconfig, fallback to use `rest.Config{}`.


jhoblitt:
What is the result of this change?  Would  `KUBECONFIG` from the env be used even when the value of `load_config_file = true` and `config_path = "/dev/null"`?

aareet:
I'm closing this now since this is attempting to solve an upstream issue - https://github.com/hashicorp/terraform-provider-kubernetes/issues/142#issuecomment-518266024.

PR #427, opened 2019-05-07
Add type to volume hostPath (closes #358)

mbarrien:
As described above. Closes #358

PR #428, opened 2019-05-07
Add CustomResourceDefinition kind

mbarrien:
This PR adds support for the CustomResourceDefinition kind. Note that it does not support objects of the given kind, just allows defining the definition within Kubernetes. Due to inherent limitations in Terraform and its diff algorithm, there are several features that are not supported.

To support this, we need to extend the provider with more than just the kubernetes.Clientset, since apiextensions are not supported inside the default clientset. We create a new kubernetesProvider type, and stick both the kubernetes.Clientset and apiextensions.Clientset inside of it. We are forced to update all references to meta.(*kubernetes.Clientset) to the new type.

Unsupported features include:
* Supporting JSON Schema validation (e.g. the schema or validation fields): These would become JSONSchemaProps objects as defined in https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1beta1/types_jsonschema.go which would become a recursive Terraform schema, which according to https://github.com/hashicorp/terraform/issues/18616 is not supported. Supporting parsing an opaque JSON blob into the given type is left for a future PR.
* AdditionalPrinterColumn and Conversion fields, and Subresources inside a version: The behavior of these are dependent on the user enabling or disabling the CustomResourceWebhookConversion feature, and were causing spurious diffs. I left the code in but commented out to support this, in the event someone wants to tackle the spurious diffs, or for when this gets solved by Terraform support Kubernetes server side diffs.

Other quirks include:
* Kubernetes will autopopulate the listKind field in names, which could cause spurious diffs. A workaround is for the Terraform file to explicitly define list_kind.
* Kubernetes will autopopulate the versions list from the version field and vice versa, causing spurious diffs. Someone defining the CustomResourceDefinition kind can workaround this by explicitly defining both.

The motivation of this was getting Calico installed on an EKS cluster. With this functionality (and #427), Terraform can install Calico on an EKS cluster by converting the YAML files from https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/master/config/v1.4/calico.yaml (as linked from https://docs.aws.amazon.com/eks/latest/userguide/calico.html) into Terraform format, using the caveats described above.


sl1pm4t:
@mbarrien 
> Terraform can install Calico on an EKS cluster by converting the YAML files from...

Shameless plug but you might find this tool helpful: [https://github.com/sl1pm4t/k2tf](https://github.com/sl1pm4t/k2tf)

JulienBreux:
Any updates on this?

alexsomesan:
Quick update here: the recently released v0.3.1 of the Kubernetes Alpha provider has quite usable support for CRDs. I would encourage everyone to give that a try a provide us with feedback. That would help us to graduate that provider to GA as soon as possible.

PR #434, opened 2019-05-09
Add doc examples for container health probes.

alexsomesan:
Adds examples of how to specify probes like `liveness_probe` on `container` blocks.


alexsomesan:
> LGTM.
> 
> You may want to add the same to the pod doc:
> 
> https://github.com/terraform-providers/terraform-provider-kubernetes/blob/50ba01ad941cd65b3a71cf3419aace7a9842c06b/website/docs/r/pod.html.markdown#L18

I did consider it, but was at the same time doubting it really makes a lot of sense for a stand-alone Pod. I added it anyway.

Issue #418, opened 2019-04-29
Cannot set max_unavailable to 0


lpil:
### Terraform Version

```
Terraform v0.11.13
...
```

### Terraform Configuration Files

```hcl
resource "kubernetes_deployment" "app" {
  metadata {
    name      = "app"

    labels { workload = "app" }
  }

  spec {
    replicas = 3

    selector {
      match_labels {
        workload = "app"
        role     = "web_server"
      }
    }

    strategy {
      type = "RollingUpdate"

      rolling_update = {
        max_unavailable = 0 # Problem here
        max_surge       = "25%"
      }
    }

    template {
      metadata {
        labels {
          workload = "app"
          role     = "web_server"
        }
      }

      spec {
        container {
          image = "hashicorp/http-echo"
          name  = "app"
        }
      }
    }
  }
}


...
```

### Expected Behavior

0 should be accepted, as specified in the documentation: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-unavailable

### Actual Behavior
```
Error: module.infra.kubernetes_deployment.app: expected value of spec.0.strategy.0.rolling_update.0.max_unavailable to match regular expression "^([1-9][0-9]*|[1-9][0-9]%|[1-9]%|100%)$"

```

### Steps to Reproduce

1. `terraform init`
2. `terraform apply`

mbarrien:
Caused by #372 adding a ValidateFunc to max_unavailable (and max_surge) that incorrectly bans 0 as a legal value.

PR #443, opened 2019-05-13
Allow max_unavailable and max_surge to be 0

dylannz-sailthru:
This allows for the valid case where either max_surge or max_unavailable
is 0 but the other is non-0.

At least one of these must be non-0, which means this change will allow
both values to be set to 0 which could cause issues. Suggest some way of
validating a resource (potentially add some kind of validation function
parameter to schema.Resource?) to ensure dependencies between different
parameters are being met.

Fixes https://github.com/terraform-providers/terraform-provider-kubernetes/issues/418


alexsomesan:
Looks good. I'm running it in CI. Will merge if green.

alexsomesan:
Unfortunately there is no way to validate across several attributes at the moment. Mostly because the validation functions don't get a reference to the ResourceData struct. That would be a change to TF core that needs to happen first.

PR #497, opened 2019-06-06
Support in-cluster configuration with service accounts

MrSaints:
Prior to this commit, the provider supports authenticating through
a local `kubeconfig` or through various configuration provided directly
to the provider or from the environment.

The provider however, did not authenticate using a service account
when running inside a Kubernetes pod. This commit provides support
for loading in-cluster configuration so that the provider can
be used inside a Kubernetes pod (with a service account) without
much additional configuration.

The use-case for this is running Terraform via Atlantis inside
a Kubernetes pod. But, outside of Atlantis, there may be cases where
we want to run some Terraform automation inside a pod (e.g. GitLab runner).

### DIFF
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -150,7 +150,16 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		return nil, err
 	}
 	if cfg == nil {
-		cfg = &restclient.Config{}
+		// Attempt to load in-cluster config
+		cfg, err = restclient.InClusterConfig()
+		if err != nil {
+			// Fallback to standard config if we are not running inside a cluster
+			if err == restclient.ErrNotInCluster {
+				cfg = &restclient.Config{}
+			} else {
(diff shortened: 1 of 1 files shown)


aeneasr:
Can this please be merged, we need this urgently.

alexsomesan:
@aeneasr It's merged. Will be part of the next release.

MrSaints:
@alexsomesan Thanks for the review, and merge. This will help us tremendously 🎉

aeneasr:
Awesome, thanks!

sleenen:
When will we see a new release? 

I'm running into ```Error: Failed to configure: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory```  issues in my pipeline (bitbucket pipelines) too when using ```load_config_file = false``` 

With 1.9 it works, so downgraded for now, but something seems very off when telling the provider to not load the config file. 

I'm curious to see if the PR merged to solve this issue will also solve mine.

PR #502, opened 2019-06-12
Fix sometimes deployment is not waited

titilambert:
"Sometimes" the kubernetes provider doesn't wait for the deployment.
After debugging, I saw that's because the following condition is not respected:
```
dply.Generation <= dply.Status.ObservedGeneration
```
So the "wait" time is `0s`

When this condition is not respected, `dply.Generation = 1` and `dply.Status.ObservedGeneration = 0`.

So I just added an other test which handle when only `dply.Status.ObservedGeneration = 0`. Since then, I don't have this issue anymore.

### DIFF
--- a/kubernetes/resource_kubernetes_deployment.go
+++ b/kubernetes/resource_kubernetes_deployment.go
@@ -385,6 +385,8 @@ func waitForDeploymentReplicasFunc(conn *kubernetes.Clientset, ns, name string) 
 			if dply.Status.AvailableReplicas < dply.Status.UpdatedReplicas {
 				return resource.RetryableError(fmt.Errorf("Waiting for rollout to finish: %d of %d updated replicas are available...", dply.Status.AvailableReplicas, dply.Status.UpdatedReplicas))
 			}
+		} else if dply.Status.ObservedGeneration == 0 {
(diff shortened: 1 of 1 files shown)


kubernetes/resource_kubernetes_deployment.go
```diff
@@ -385,6 +385,8 @@ func waitForDeploymentReplicasFunc(conn *kubernetes.Clientset, ns, name string)
 			if dply.Status.AvailableReplicas < dply.Status.UpdatedReplicas {
 				return resource.RetryableError(fmt.Errorf("Waiting for rollout to finish: %d of %d updated replicas are available...", dply.Status.AvailableReplicas, dply.Status.UpdatedReplicas))
 			}
+		} else if dply.Status.ObservedGeneration == 0 {
```
alexsomesan: Can `dply.Generation` ever be `0` and thus also satisfy the first branch of the condition?

djsly: We haven’t seen it so far. 

It’s either 1 and 1 or 1 and 0 

We see this when we deploy tiller on a fresh AKS using k8s provider. The helm provider follows up with an error that tiller isn’t installed once in a while.

djsly:
thanks @alexsomesan ,  what is the typically tagging schedule for this provider ?

PR #511, opened 2019-06-19
accept PATH-style kubeconfig

cruwe:
The env var KUBECONFIG may be set in the style of $PATH, i.e., mutliple
files divided by a colon, so that the actual configuration may be spread
out over multiple files. THis may be desirable so that the non-secret
cluster configuration may be checked into a source code repo leaving
actual credentials separate.

### DIFF
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -3,17 +3,17 @@ package kubernetes
 import (
 	"bytes"
 	"fmt"
-	"log"
-	"os"
-
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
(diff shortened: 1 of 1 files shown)


cruwe:
First, thank you very much for developing the terraform kubernetes provider.

I yet have to spend some hours coding golang, so please bear with me should my code be ugly.

colinhoglund:
FWIW, I checked this out locally and it seems to work, but I suspect maintainers might want some unit testing. Would love to see this merged!

aareet:
We now support `config_paths` in version 2.x - https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#config_paths

dak1n1:
Closing since the equivalent feature has been released in version 2.x.

PR #513, opened 2019-06-20
Automountsa

tlmiller:
* Adds automountServiceAccountToken support to podspec. See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#podspec-v1-core


alexsomesan:
There is already an open PR addressing this issue. Let's concentrate on working on that one. It's almost there.

https://github.com/terraform-providers/terraform-provider-kubernetes/pull/261

PR #563, opened 2019-07-24
Migrate test infra to TF 0.12

alexsomesan:
This change is the result of running `terraform 0.12upgrade` command inside each of the test-infra configurations.

I will update with TC results as the runs complete.

### DIFF
--- a/kubernetes/test-infra/eks/node-config/config.tf
+++ b/kubernetes/test-infra/eks/node-config/config.tf
@@ -21,17 +21,20 @@ MAPPEDROLE
 
 resource "local_file" "cluster_ca" {
   content = base64decode(var.cluster_ca)
-  filename = "${path.module}/cluster_ca"
+  filename = "${path.root}/cluster_ca"
 }
 
 resource "null_resource" "wait_for_api" {
   depends_on = [local_file.cluster_ca]
 
   provisioner "local-exec" {
-
-    working_dir = path.module
+    working_dir = path.root
+    environment = {
+      K8S_CA_FILE = abspath(local_file.cluster_ca.filename)
+      K8S_ENDPOINT = var.cluster_endpoint
+    }
     command = <<CMDEOF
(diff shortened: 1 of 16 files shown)


kubernetes/test-infra/eks/node-config/config.tf
```diff
@@ -16,37 +16,42 @@ locals {
...
 resource "null_resource" "wait_for_api" {
-  depends_on = ["local_file.cluster_ca"]
+  depends_on = [local_file.cluster_ca]
 
   provisioner "local-exec" {
+
+    working_dir = path.module
     command = <<CMDEOF
-while ! curl -s --cacert ${local_file.cluster_ca.filename} ${var.cluster_endpoint}/version > /dev/null; do 
+while ! curl -s --cacert ./cluster_ca ${var.cluster_endpoint}/version > /dev/null; do
```
radeksimko: Is this intended? I think that we're leaking implementation details (i.e. filename) by doing this.

alexsomesan: The two failed tests are unrelated and were also failing before this change.
I'm working on fixing them next.

Screenshot 2019-07-25 at 12 23 07

alexsomesan: Also, that filename is set on line 24 above, so not exactly obscured :)

radeksimko: For now - code can be moved in the future an become obscured, so using references rather than hardcoded duplicated values in this specific case is beneficial IMO and if the previous/existing notation works in 0.12 I don't see a reason why not.

If we break it again in future releases, we should be the first to notice through the dog fooding here, before our users do in the wild 😜

alexsomesan: Found a solution to re-use the `filename` attribute from `local_file`. Thanks for the tip about `abspath()`.

radeksimko: Nice, that's so much cleaner! Thanks.

Issue #567, opened 2019-07-26
PVC storage_class_name empty sets default storage class


davidq2q:
Hello,

There is a critical bug for the kubernetes terraform provider that has been reported since 2017 but is still open.

See:
https://github.com/terraform-providers/terraform-provider-kubernetes/issues/102
https://github.com/terraform-providers/terraform-provider-kubernetes/issues/339

What happens is that terraform is not passing the empty storage class string instead somehow the pvc's are created with the default storage class. The work around mentioned by using label selectors is not working anymore on terraform 0.12.

Best regards,
David

ajchiarello:
I was able to work around this by creating a dummy storage class that I applied to the PV and the PVC - not ideal, but it did work.

davidq2q:
@ajchiarello how did you set up the dummy class?

lawliet89:
I opened #590 to fix this.

koxu1996:
Any updates on merging this? Bug is really annoying.

leonbreedt:
+1, this is a blocker for using PVCs in our scenario, and we have to do volume mounts directly on the pod template spec, which does not allow us to override mount options, which breaks our scenario.

dak1n1:
Hi, we have a few issues relating to this that are being consolidated here: https://github.com/terraform-providers/terraform-provider-kubernetes/issues/872 See the new issue for a work-around. Thanks!

PR #568, opened 2019-07-26
Still add to obj when value is empty

davidq2q:
See: https://github.com/terraform-providers/terraform-provider-kubernetes/issues/567


mbarrien:
Duplicate of #590 (and the other one also handles flattening)

Issue #401, opened 2019-04-18
feature request: kubernetes_namespace timeout block


jhoblitt:
### Terraform Version
```
Terraform v0.11.13
```

### Affected Resource(s)
Please list the resources as a list, for example:
- `kubernetes_namespace`

### (Desired) Terraform Configuration Files

```hcl
resource "kubernetes_namespace" "foo" {
  metadata {
    name = "bar"
  }

  timeouts {
    delete = "30m"
  }
}
```
### Use Case

k8s namespace termination ends up being a catch all for resources that haaven't already been torn down.  In some cases, such as with pvc termination, it can take a very, very, very long time.  It would be very convenient to be able to extend the default timeout.

gertjangaillet:
Would it be a solution to increase the default timeout? This is also done in for instance [the aws_eks_cluster resource](https://github.com/terraform-providers/terraform-provider-aws/tree/master/aws):
```
		Timeouts: &schema.ResourceTimeout{
			Create: schema.DefaultTimeout(30 * time.Minute),
			Update: schema.DefaultTimeout(60 * time.Minute),
			Delete: schema.DefaultTimeout(15 * time.Minute),
		},
```
What would be sensible defaults? Something like this:
* Create: 5 min
* Update: 10 min
* Destroy: 30 min

I'd be happy to update this and provide a PR myself, as we have lots of manual work because of the fact it can take a much longer time to destroy a namespace due to pvc's that are being deleted in the background, and running tf destroy after the namespace has been destroyed after a timeout simply doesn't recover...

gertjangaillet:
Related to #335

gertjangaillet:
@jhoblitt - I assume this can be closed now

Issue #335, opened 2019-02-23
kubernetes_namespace fails to destroy; stuck in Terminating state


appnojag:
### Terraform Version

 - v0.11.11

### Affected Resource(s)

 - kubernetes_namespace

### Terraform Configuration Files
```hcl

resource "kubernetes_namespace" "foo" {
  metadata {
    name = "foo"
  }
}

```

### Expected Behavior

Kubernetes namespace gets deleted [within hardcoded 5m timeout] during terraform destroy.

### Actual Behaviour

Kubernetes namespace is stuck in Terminating state; terraform destroy fails.

### Important Factoids

 - Google Kubernetes Engine 1.11.7-gke.4

### Steps to Reproduce

1. `terraform destroy`

### References

- GH-60807

appnojag:
https://github.com/kubernetes/kubernetes/issues/60807

markeissler:
The following reply in that issue thread suggests it's a terraform problem that's causing this, related to the order in which resources are destroyed:

https://github.com/kubernetes/kubernetes/issues/60807#issuecomment-463764193

alexsomesan:
@appnojag Can you provide of log of the destroy operation with TF_LOG=TRACE environment variable set? This will allow us to trace the communication between Terraform and the K8s API.
If this is cause by the problem described in the Kubernetes thread mentioned above, then there isn’t a whole lot Terraform can do.

gertjangaillet:
@appnojag - can you check if the configurable timeout resolves your issue?

alexsomesan:
This issue is easily reproduced on vanilla Kubernetes by configuring a namespace with a malfunctioning finalizer. This is not something that can be mitigated by the provider and as such there's not much we can do here but fail the destroy in the event of timeout.

Can you confirm if setting the `Delete` timeout on the namespace resource works for you as described?

chrisrot:
@alexsomesan Can elaoborate how to avoid a malfunctioning finalizer?
I have created the namespace like this:
```
resource "kubernetes_namespace" "foo" {
  metadata {
    annotations = {
      name = "foo"
    }
    name = "foo"
  }
}
```

dan-barrett:
@chrisrot, I was seeing the same issue you are. I'm realizing that this is an issue with the kubernetes APIs rather than a specific kubernetes_namespace you've deployed to.

Try running this: `kubectl api-resources`

[...]

PR #570, opened 2019-07-28
#401 configurable delete timeout for kubernetes_namespace resource

gertjangaillet:
Adding a configurable timeout for deleting the kubernetes_namespace resource.
Would resolve issues #401 and #335


gertjangaillet:
@alexsomesan I added a test, and also added an additional paragraph in the docs to indicate that kubernetes_namespace supports a delete timeout - however CI fails now while testing the website with the following error: `Error response from daemon: No such container: tf-website-kubernetes-temp`

Other tests are successful.

Can you have a look at the CI?

PR #607, opened 2019-09-05
Parametrise all timeouts

bartoszj:
Parametrised all resource timeouts.

In my case, creating a load balancer service is taking more than hardcoded 10 minutes. I decided to parametrise all timeouts which I've found in the code. I didn't change any value, so the provider is working as before, but we have more options to customise.


alexsomesan:
@bartoszj This is good stuff, thank you! I'll test it and come back with feedback soon.

alexsomesan:
CI ok

Issue #534, opened 2019-07-06
Resource `kubernetes_job` has a different behavioral from the documentation


DzeryCZ:
Hi there,

`Terraform v0.12.1 `
`provider.kubernetes v1.8.0`

I'm trying to play around with `kubernetes_job` resource in Terraform, but I spotted a different behavioral from the one described in the documentation.

---

1. **Waiting for `Completed` status**

Documentation says: 
>A Job creates one or more Pods and ensures that a specified number of them successfully terminate. As pods successfully complete, the Job tracks the successful completion. When a specified number of successful completion is reached, the task (ie, Job) is complete.

Let's say I have given terraform config:
```tf
resource "kubernetes_job" "migration" {
  metadata {
    name = "migration"
  }
  spec {
    template {
      metadata {}
      spec {
        container {
          name    = "migration"
          image   = "alpine:latest"
          command = ["sleep", "60"] 
        }
        restart_policy = "Never"
      }
    }
    backoff_limit = 4
    completions = 1
  }
}
```

When I do `terraform apply` it doesn't wait for the `Completed` status of the job. But it just creates the resource and then it doesn't care about the successful or unsuccessful result.

```bash
kubernetes_job.migration: Creating... 
kubernetes_job.migration: Creation complete after 3s [id=default/migration] 
```

... the job becomes to status `Completed` after 60 seconds or so.

---

2. **Destroy a job doesn't clean up the pods**
Documentation says:
> Deleting a Job will clean up the Pods it created.

But when I run `terraform destroy` it returns
```bash
kubernetes_job.migration: Destroying... [id=default/migration] 
kubernetes_job.migration: Destruction complete after 0s
```

But when I list pods, I can still see Pod used for the job.
```bash
$ kubectl get jobs                                                                                                                                          
No resources found. 

$ kubectl get pods
NAME              READY   STATUS      RESTARTS   AGE 
migration-42zs7   0/1     Completed   0          3h18m
```

I think this is a bug. It should work as described in the documentation, but unfortunately, it does not work like that.

Many thanks.

alexsomesan:
Hi DzeryCZ,

Thanks for starting the conversation on this behaviour. It's helpful for us to understand how users relate to it.
Let me first review the chain of events that happens in this case, which I think will help in understanding why things behave like you've described.

[...]

PR #625, opened 2019-09-22
Add Kubernetes Job - `wait_for_completion` functionality

ichekrygin:
### Overview
The current behavior of `kubernetes_job` resource is to return immediately once it successfully creates the Kubernetes job resource. This behavior, however, creates a problem when we are using Kubernetes Jobs as a dependency step. For example, one may use Kubernetes Job to prepare/update the environment, before advancing to using the environment (database migration, etc.) In such cases, it is essential to wait until the successful completion of Kubernetes Job to proceed to the next step.

### Changes
This commit adds `wait_for_completion = true/false` attribute to the
`kubernetes_job` resource. The default value (if not provided) is `false`,
thus `kubernetes_job` has the same functionality as in the original.

When enabled `wait_for_completion = true`, upon resource create or update
the process will wait until said Kubernetes Job reaches "Complete" state

### Tracking
#534

#### Default (current) Behavior 
```yaml
resource kubernetes_job test-pr {
  metadata {
    name = "job-with-wait"
    namespace = "default"
  }
  spec {
    completions = 1
    active_deadline_seconds = 60
    template {
      metadata {}
      spec {
        container {
          name = "sleep"
          image = "busybox:latest"
          command = ["sleep", "30"]
        }
        restart_policy = "Never"
      }
    }
  }
}
```
Results in:
```

  Enter a value: yes

kubernetes_job.test-pr: Creating...
kubernetes_job.test-pr: Creation complete after 0s [id=default/job-with-wait]

```
While the actual Job and Job's Pod is 
```
> kubectl get job
NAME            COMPLETIONS   DURATION   AGE
job-with-wait   0/1           7s         7s
⋊> kubectl get job job-with-wait -ojson | jq '.status'
{
  "active": 1,
  "startTime": "2019-09-22T21:19:42Z"
}
⋊> kubectl get pod
NAME                  READY   STATUS    RESTARTS   AGE
job-with-wait-k4wzm   1/1     Running   0          28s
```

#### With Wait
Same as the above, but with `wait_for_completion = true` 

```yaml
resource kubernetes_job test-pr {
  metadata {
    name = "job-with-wait"
    namespace = "default"
  }
  spec {
    completions = 1
    active_deadline_seconds = 60
    template {
      metadata {}
      spec {
        container {
          name = "sleep"
          image = "busybox:latest"
          command = ["sleep", "30"]
        }
        restart_policy = "Never"
      }
    }
  }
  wait_for_completion = true
}
```
Results in:
```

  Enter a value: yes

kubernetes_job.test-pr: Creating...
kubernetes_job.test-pr: Still creating... [10s elapsed]
kubernetes_job.test-pr: Still creating... [20s elapsed]
kubernetes_job.test-pr: Still creating... [30s elapsed]
kubernetes_job.test-pr: Creation complete after 37s [id=default/job-with-wait]
```

### DIFF
--- a/kubernetes/resource_kubernetes_job.go
+++ b/kubernetes/resource_kubernetes_job.go
@@ -85,7 +85,7 @@ func resourceKubernetesJobCreate(d *schema.ResourceData, meta interface{}) error
 	}
 	if d.Get("wait_for_completion").(bool) {
 		return resource.Retry(d.Timeout(schema.TimeoutCreate),
-			retryUntilJobCondition(conn, namespace, name, batchv1.JobComplete))
+			retryUntilJobIsFinished(conn, namespace, name))
 	}
 
 	return resourceKubernetesJobRead(d, meta)
@@ -121,7 +121,7 @@ func resourceKubernetesJobUpdate(d *schema.ResourceData, meta interface{}) error
 
 	if d.Get("wait_for_completion").(bool) {
 		return resource.Retry(d.Timeout(schema.TimeoutUpdate),
-			retryUntilJobCondition(conn, namespace, name, batchv1.JobComplete))
+			retryUntilJobIsFinished(conn, namespace, name))
 	}
 	return resourceKubernetesJobRead(d, meta)
 }
@@ -238,7 +238,8 @@ func resourceKubernetesJobExists(d *schema.ResourceData, meta interface{}) (bool
 	return true, err
 }
 
-func retryUntilJobCondition(conn *kubernetes.Clientset, ns, name string, condition batchv1.JobConditionType) resource.RetryFunc {
+// retryUntilJobIsFinished checks if a give job finished its execution and either in Complete or Failed state
+func retryUntilJobIsFinished(conn *kubernetes.Clientset, ns, name string) resource.RetryFunc {
 	return func() *resource.RetryError {
 		job, err := conn.BatchV1().Jobs(ns).Get(name, metav1.GetOptions{})
 		if err != nil {
@@ -246,12 +247,17 @@ func retryUntilJobCondition(conn *kubernetes.Clientset, ns, name string, conditi
 		}
 
 		for _, c := range job.Status.Conditions {
-			if c.Type == condition && c.Status == corev1.ConditionTrue {
-				return nil
+			if c.Status == corev1.ConditionTrue {
+				log.Printf("[DEBUG] Current condition of job: %s/%s: %s\n", ns, name, c.Type)
+				switch c.Type {
+				case batchv1.JobComplete:
+					return nil
+				case batchv1.JobFailed:
+					return resource.NonRetryableError(fmt.Errorf("job: %s/%s is in failed sate", ns, name))
+				}
 			}
-			log.Printf("[DEBUG] Current condition of job: %s/%s: %s\n", ns, name, c.Type)
 		}
 
-		return resource.RetryableError(fmt.Errorf("job: %s/%s is not in: %s condition", ns, name, condition))
+		return resource.RetryableError(fmt.Errorf("job: %s/%s is not in complete state", ns, name))
 	}
 }
--- a/kubernetes/resource_kubernetes_job.go
+++ b/kubernetes/resource_kubernetes_job.go
@@ -85,7 +85,7 @@ func resourceKubernetesJobCreate(d *schema.ResourceData, meta interface{}) error
 	}
 	if d.Get("wait_for_completion").(bool) {
 		return resource.Retry(d.Timeout(schema.TimeoutCreate),
-			waitForJobCondition(conn, namespace, name, batchv1.JobComplete))
+			retryUntilJobCondition(conn, namespace, name, batchv1.JobComplete))
 	}
 
 	return resourceKubernetesJobRead(d, meta)
@@ -121,7 +121,7 @@ func resourceKubernetesJobUpdate(d *schema.ResourceData, meta interface{}) error
 
 	if d.Get("wait_for_completion").(bool) {
 		return resource.Retry(d.Timeout(schema.TimeoutUpdate),
-			waitForJobCondition(conn, namespace, name, batchv1.JobComplete))
+			retryUntilJobCondition(conn, namespace, name, batchv1.JobComplete))
 	}
 	return resourceKubernetesJobRead(d, meta)
 }
@@ -238,7 +238,7 @@ func resourceKubernetesJobExists(d *schema.ResourceData, meta interface{}) (bool
 	return true, err
 }
 
-func waitForJobCondition(conn *kubernetes.Clientset, ns, name string, condition batchv1.JobConditionType) resource.RetryFunc {
+func retryUntilJobCondition(conn *kubernetes.Clientset, ns, name string, condition batchv1.JobConditionType) resource.RetryFunc {
 	return func() *resource.RetryError {
 		job, err := conn.BatchV1().Jobs(ns).Get(name, metav1.GetOptions{})
 		if err != nil {
@@ -246,13 +246,10 @@ func waitForJobCondition(conn *kubernetes.Clientset, ns, name string, condition 
 		}
 
 		for _, c := range job.Status.Conditions {
-			if c.Status != corev1.ConditionTrue {
-				continue
-			}
-			log.Printf("[DEBUG] Current condition of job: %s/%s: %s\n", ns, name, c.Type)
-			if c.Type == condition {
+			if c.Type == condition && c.Status == corev1.ConditionTrue {
 				return nil
 			}
+			log.Printf("[DEBUG] Current condition of job: %s/%s: %s\n", ns, name, c.Type)
 		}
 
 		return resource.RetryableError(fmt.Errorf("job: %s/%s is not in: %s condition", ns, name, condition))
--- a/kubernetes/resource_kubernetes_job.go
+++ b/kubernetes/resource_kubernetes_job.go
@@ -28,6 +28,7 @@ func resourceKubernetesJob() *schema.Resource {
 
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(1 * time.Minute),
+			Update: schema.DefaultTimeout(1 * time.Minute),
 			Delete: schema.DefaultTimeout(1 * time.Minute),
 		},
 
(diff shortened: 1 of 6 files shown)


JordanP:
Is this a functionality that K8s natively supports ?

ichekrygin:
@JordanP, not sure I fully understand the "natively" part, but, yes, you can use "wait" for job completion in Kubernetes via `kubectl`: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#wait

Example: https://stackoverflow.com/questions/44686568/tell-when-job-is-complete
```
kubectl wait --for=condition=complete job/myjob
```

For more context see #534.

alexsomesan:
Thanks a lot for this! I think it will make great user experience improvement.
Before merging I think we need to address the usage of timeout values a bit. 

First of all, in order to make the timeout values configurable by the user and not just rely on the defaults, they need to be declared in the `schema.Resource` block. You can find a good explanation of how to achieve that here: https://www.terraform.io/docs/extend/resources/retries-and-customizable-timeouts.html

Second, since this change in behaviour affects both Create and Update action I think both the Create and the Update timeout values should be respected in their specific contexts. Right now only the Create timeout has an effect.

Would you mind addressing these two things? 

Also please rebase. PR #635 also added some related functionality that you might want to be aware of.

kubernetes/resource_kubernetes_job.go
```diff
@@ -25,6 +27,7 @@ func resourceKubernetesJob() *schema.Resource {
 		},
 
 		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(1 * time.Minute),
```
alexsomesan:
```suggestion
			Create: schema.DefaultTimeout(1 * time.Minute),
			Update: schema.DefaultTimeout(1 * time.Minute),
```

kubernetes/resource_kubernetes_job.go
```diff
@@ -97,6 +111,11 @@ func resourceKubernetesJobUpdate(d *schema.ResourceData, meta interface{}) error
 	log.Printf("[INFO] Submitted updated job: %#v", out)
 
 	d.SetId(buildId(out.ObjectMeta))
+
+	if d.Get("wait_for_completion").(bool) {
+		return resource.Retry(d.Timeout(schema.TimeoutCreate),
```
alexsomesan: I think this should follow the value of `schema.TimeoutUpdate` instead. See my suggestion above about adding one to the schema.

website/docs/r/job.html.markdown
```diff
@@ -94,3 +121,14 @@ More info: http://kubernetes.io/docs/user-guide/labels
 These arguments are the same as the for the `spec` block of a Pod.
 
 Please see the [Pod resource](pod.html#spec-1) for reference.
+
+## Timeouts
+
+The following [Timeout](/docs/configuration/resources.html#operation-timeouts) configuration options are available for the `kubernetes_job` resource when used with `wait_for_completion = true`:
+
+* `create` - (Default `1 minute`) Used for creating new job and waiting for a successful job completion.
```
alexsomesan:
```suggestion
* `create` - (Default `1 minute`) Used for creating a new job and waiting for a successful job completion.
* `update - (Default `1 minute`) Used for updating an existing job and waiting for a successful job completion.
```

ichekrygin:
@alexsomesan, updated, PTAL

arturrez:
@alexsomesan, is it ready to be merged. We are looking forward for this functionality.

quentin9696:
Hi Guys, 

Where do you plan this feature ? It actually complicated to managed my dependencies with terraform

ichekrygin:
I am under impression that this PR review is "abandoned" by terraform-provider-kubernetes maintainers, @alexsomesan.

alexsomesan:
Apologies for dropping the ball here. I'll have a another look at this and run some tests on it.

quentin9696:
that's a good news ! I'll see keep looking this thread

thunder-spb:
Any updates here? really useful feature.

kubernetes/resource_kubernetes_job.go
```diff
@@ -207,3 +222,24 @@ func resourceKubernetesJobExists(d *schema.ResourceData, meta interface{}) (bool
...
+		job, err := conn.BatchV1().Jobs(ns).Get(name, metav1.GetOptions{})
+		if err != nil {
+			return resource.NonRetryableError(err)
+		}
+
+		for _, c := range job.Status.Conditions {
+			if c.Status != corev1.ConditionTrue {
+				continue
+			}
+			log.Printf("[DEBUG] Current contions of job: %s/%s: %s\n", ns, name, c.Type)
```
zuc:
```suggestion
			log.Printf("[DEBUG] Current conditions of job: %s/%s: %s\n", ns, name, c.Type)
```

zuc:
Hey @alexsomesan, any updates here? 🙏

dpkirchner:
It looks like 2e2a19295b6bbf34551dbe214c71e085fe1a6be9 is no longer necessary once you rebase this onto master. Removing it and rebasing should resolve the failed checks.

Thanks for this feature, btw!

ichekrygin:
@dpkirchner rebased, PTAL.

dpkirchner:
@ichekrygin LGTM, and it worked in my testing.

kubernetes/resource_kubernetes_job_test.go
```diff
@@ -63,6 +64,7 @@ func TestAccKubernetesJob_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("kubernetes_job.test", "spec.0.manual_selector", "true"),
 					resource.TestCheckResourceAttr("kubernetes_job.test", "spec.0.template.0.spec.0.container.0.name", "hello"),
 					resource.TestCheckResourceAttr("kubernetes_job.test", "spec.0.template.0.spec.0.container.0.image", "alpine"),
+					resource.TestCheckResourceAttr("kubernetes_job.test", "wait_for_completion", "true"),
```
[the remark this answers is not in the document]
jrhouston: An idea that springs to mind having the test create a job that takes duration _N_ to run, and asserting that the resource took _no less_ than duration _N_ to be created.

kubernetes/resource_kubernetes_job.go
```diff
@@ -73,6 +79,15 @@ func resourceKubernetesJobCreate(d *schema.ResourceData, meta interface{}) error
 
 	d.SetId(buildId(out.ObjectMeta))
 
+	namespace, name, err := idParts(d.Id())
+	if err != nil {
+		return err
+	}
+	if d.Get("wait_for_completion").(bool) {
+		return resource.Retry(d.Timeout(schema.TimeoutCreate),
+			waitForJobCondition(conn, namespace, name, batchv1.JobComplete))
```
jrhouston, line 88: Looking at this a bit more - if we run a job that fails immediately, Terraform will just wait it out and eventually error when the timeout is reached. The error message is not super helpful: 

```
Error: job: default/job-with-wait is not in: Complete condition
```

What do you think about also capturing condition type == Failed and returning a NonRetryableError telling the user that the Job failed? It will save people waiting around for a few minutes when we can tell them their job failed right away.

kubernetes/resource_kubernetes_job.go
```diff
@@ -222,3 +237,24 @@ func resourceKubernetesJobExists(d *schema.ResourceData, meta interface{}) (bool
 	}
 	return true, err
 }
+
+func waitForJobCondition(conn *kubernetes.Clientset, ns, name string, condition batchv1.JobConditionType) resource.RetryFunc {
+	return func() *resource.RetryError {
+		job, err := conn.BatchV1().Jobs(ns).Get(name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
ichekrygin: I think I agree in principle that `watch` could be a "more efficient" way of keeping track of state changes than "retry".

kubernetes/resource_kubernetes_job.go
ichekrygin: Yes, I agree, current job failure handling is less than ideal. 

The `retryUntilJobCondition` initially is intended to be generic so that it can be used to wait for job failures (if needed). 
With that said, I don't see an apparent use case when a user will run a job that intended to fail (doesn't mean that it isn't a valid use-case)

What if we provide a list of conditions, i.e.
```go
func retryUntilJobHasACondition(conn *kubernetes.Clientset, ns, name string, conditions []batchv1.JobConditionType) resource.RetryFunc {
...
```
This could allow us to use it as the following:
```go
retryUntilJobCondition(conn, namespace, name, []batchv1.JobConditionType{batchv1.JobComplete, batchv1.JobFailed}))
```

jrhouston: I can totally think of use-cases where you would _expect_ that something is going to fail, i.e you might use terraform to start a job that runs some kind of test as part of your CI and then report on it having failed. 

I was thinking more for users that might be running Jobs that contain initialization or migration steps or something like that - where it would be helpful for the user for Terraform to explicitly tell them that the job failed, rather than just letting it time out - it takes a way a little bit of uncertainty. The set of reasons for a Job failing is a lot smaller than for a Job having timed out.

The issue with just accepting a list of conditions is that the resource will say it was created successfully even if the job ends up with a failed condition. This might confuse users who use a Job as part of a `depends_on`.

What about supplying two lists of conditions, one list of "success" conditions that will consider the resource created, and another list of "failure" conditions that will bubble up as an error to the user?

ichekrygin: > The set of reasons for a Job failing is a lot smaller than for a Job having timed out.

I think you made a compelling argument and I am coming around on:

> What do you think about also capturing condition type == Failed and returning a NonRetryableError telling the user that the Job failed? It will save people waiting around for a few minutes when we can tell them their job failed right away.

ichekrygin: Thank you for the review and feedback, @jrhouston. 

I incorporated most of the feedback suggestions, except the:
> An idea that springs to mind having the test create a job that takes duration N to run, and asserting that the resource took no less than duration N to be created.

It has been a while since I initially drafted this PR (back in Sep) and TBH even back then I think I didn't understand well enough terraform provider unit tests, so I may need some additional help with that.

jrhouston: No problem, thanks for making those changes! I figured out a way doing this so I will add a commit to your branch with the test then we can merge.

kubernetes/resource_kubernetes_job_test.go
```diff
@@ -16,6 +17,44 @@ func ttlAfterDisabled() (bool, string) {
 	return os.Getenv("FEATURE_GATE_TTL_AFTER_FINISHED") != "enabled", "TTLAfterFinished is not enabled"
 }
 
+func TestAccKubernetesJob_wait_for_completion(t *testing.T) {
```
jrhouston, line 20: @ichekrygin I added a separate test just for this functionality to make it a bit easier to understand what is being tested.

PR #634, opened 2019-10-07
Bump random provider to v2.2.1

radeksimko:
This is in preparation for migrating this provider to the standalone plugin SDK.

Bumping of other providers (AWS & Google) will follow when they cut their next release which should in both cases already use the standalone SDK.


alexsomesan:
I've kicked off CI on this

PR #635, opened 2019-10-08
set option to cascade delete job resources

houqp:
This is also the default behavior from kubectl.

Currently, when a job is deleted from Terraform, all pods are left behind as orphan resources.

Relates to #534.


alexsomesan:
Thank you, this is really useful!
The necessity for such cleanups becomes apparent as users manage K8s more with Terraform.

alexsomesan:
I'm testing it and will merge when green.

houqp:
@alexsomesan I noticed other resources type like deployment also have this issue, do mind if I work on a PR to address all of them?

alexsomesan:
That would be really nice! By all means, please do.
Much appreciated!

PR #651, opened 2019-10-17
provider: Migrate to standalone plugin SDK v1.1.1

radeksimko:
I originally thought we could just bump dependencies in a separate PR, but it turns out it's easier to do everything in one PR.

TODO: Run acceptance tests


alexsomesan:
Testing this now.

PR #694, opened 2019-11-18
fix minor typo

npmaile:


aareet:
PR couldn't be merged because the upstream no longer exists. Will make this change in a separate PR.

PR #700, opened 2019-11-25
Fix GKE test infra attributes for Google provider v3.x.x

alexsomesan:
In the Google provider, attributes previously referring to `zone` are now referring to `location`.

Relates to https://github.com/GoogleCloudPlatform/magic-modules/pull/2436


alexsomesan:
Test infra again builds correctly.

PR #712, opened 2019-12-17
Bug in role binding patch

simonklb:
There is a bug when updating the subjects in a kubernetes_cluster_role_binding (and probably kubernetes_role_binding as well) resource in specific ways. I'm not totally sure what the actual corner cases are but they seem to be linked to the number of subjects you add and remove and the order they are in.

I've not had time to fix the bug yet, but I've added a test case that showcases the bug.

Anyone with better insight into the code base, feel free to commit a fix.

### DIFF
--- a/kubernetes/resource_kubernetes_cluster_role_binding_test.go
+++ b/kubernetes/resource_kubernetes_cluster_role_binding_test.go
@@ -157,6 +157,67 @@ func TestAccKubernetesClusterRoleBinding_importBasic(t *testing.T) {
 	})
 }
 
+func TestAccKubernetesClusterRoleBindingBug(t *testing.T) {
+	var conf api.ClusterRoleBinding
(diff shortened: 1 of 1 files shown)


simonklb:
So it looks like travis isn't configured to run the acceptance tests which makes the test I added being skipped. Here's the output from running the test I added:
```
% TF_ACC=1 go test -run TestAccKubernetesClusterRoleBindingBug ./kubernetes/ -v
=== RUN   TestAccKubernetesClusterRoleBindingBug
--- FAIL: TestAccKubernetesClusterRoleBindingBug (1.60s)
    testing.go:635: Step 1 error: Check failed: 1 error occurred:
        	* Check 16/17 error: kubernetes_cluster_role_binding.test: Attribute 'subject.1.name' expected "notauser4", got "notauser3"
        
        
FAIL
FAIL	github.com/terraform-providers/terraform-provider-kubernetes/kubernetes	1.638s
FAIL
```

alexsomesan:
Closing this as it was included in #714

PR #717, opened 2019-12-23
Fix modules for go 1.13

marcosnils:
When using go 1.13 to build this provider I was getting some
pseudo-version and ambiguous module definition errors.

This PR makes the proper updates to the provider builds in go 1.13


alexsomesan:
Fixed by #732

PR #720, opened 2019-12-31
host_aliases' ForceNew determined by isUpdatable

bradleybluebean:
In the context of things like a `kubernetes_deployment`, `host_aliases` are updatable and should not force the creation of a new resource. This removes the hard-coded `true` and lets the check for `isUpdatable` handle the setting.


aareet:
I believe this is already fixed in master -> https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/kubernetes/schema_pod_spec.go#L138

PR #731, opened 2020-01-09
Add kubernetes_service_account data source

andrein:
This PR adds support for the kubernetes service account datasource. 

This datasource makes it easy to get the auto-generated secret token from service accounts not directly managed by terraform (in our setup, we need to access a secret token created by a helm chart)

### DIFF
--- a/kubernetes/data_source_kubernetes_service_account_test.go
+++ b/kubernetes/data_source_kubernetes_service_account_test.go
@@ -19,21 +19,13 @@ func TestAccKubernetesDataSourceServiceAccount_basic(t *testing.T) {
 			{
 				Config: testAccKubernetesDataSourceServiceAccountConfig_basic(name),
 				Check: resource.ComposeAggregateTestCheckFunc(
-					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.annotations.%", "2"),
-					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.annotations.TestAnnotationOne", "one"),
-					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.annotations.TestAnnotationTwo", "two"),
-					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.labels.%", "3"),
-					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.labels.TestLabelOne", "one"),
-					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.labels.TestLabelTwo", "two"),
-					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.labels.TestLabelThree", "three"),
 					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.name", name),
-					resource.TestCheckResourceAttrSet("data.kubernetes_service_account.test", "metadata.0.generation"),
-					resource.TestCheckResourceAttrSet("data.kubernetes_service_account.test", "metadata.0.resource_version"),
-					resource.TestCheckResourceAttrSet("data.kubernetes_service_account.test", "metadata.0.self_link"),
-					resource.TestCheckResourceAttrSet("data.kubernetes_service_account.test", "metadata.0.uid"),
-					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "secret.#", "2"),
-					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "image_pull_secret.#", "2"),
+					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.annotations.TestAnnotation", "annotation"),
+					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.labels.TestLabel", "label"),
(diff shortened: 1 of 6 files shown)


andrein:
Hi guys, can someone take a look at this PR?

venky999:
can we have this PR merged.. very useful feature..

kubernetes/data_source_kubernetes_service_account_test.go
```diff
@@ -0,0 +1,51 @@
...
+					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.labels.TestLabelTwo", "two"),
+					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.labels.TestLabelThree", "three"),
+					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "metadata.0.name", name),
+					resource.TestCheckResourceAttrSet("data.kubernetes_service_account.test", "metadata.0.generation"),
+					resource.TestCheckResourceAttrSet("data.kubernetes_service_account.test", "metadata.0.resource_version"),
+					resource.TestCheckResourceAttrSet("data.kubernetes_service_account.test", "metadata.0.self_link"),
+					resource.TestCheckResourceAttrSet("data.kubernetes_service_account.test", "metadata.0.uid"),
+					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "secret.#", "2"),
+					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "image_pull_secret.#", "2"),
+					resource.TestCheckResourceAttr("data.kubernetes_service_account.test", "automount_service_account_token", "false"),
```
dak1n1, line 36: Nice! Looks like this test is passing.

```
[dakini@dax terraform-provider-kubernetes (review_731)]$ cd kubernetes ; TF_ACC=true TF_LOG=trace go test -v -run TestAccKubernetesDataSourceServiceAccount_basic ; cd -
--- PASS: TestAccKubernetesDataSourceServiceAccount_basic (0.92s)
PASS
ok  	github.com/terraform-providers/terraform-provider-kubernetes/kubernetes	0.944s
/home/dakini/go/src/github.com/hashicorp/terraform-provider-kubernetes
```

This file is also really similar to what we have throughout our repos. But I was hoping we can take it a step further to improve the process. I'll show you what I mean. I recently submitted a similar test in a PR and someone showed me a way to make it even better: https://github.com/terraform-providers/terraform-provider-kubernetes/pull/801/commits/c34cc422ad18813870d706d0f41d4adf64baf4eb

Basically we're taking this concept of "DAMP not DRY", and adding a little bit of duplicate code for the sake of readability and making the test more self-contained. This is a description of the concept, in case you're curious:  https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests/11837973#11837973

Could you update this file to have fewer tests and write out the kubernetes_service_account resource instead of referring to `testAccKubernetesServiceAccountConfig_basic`? That would help us in this goal. Thanks! (And you can refer to my PR in this comment for an example, if needed.)

kubernetes/provider.go
```diff
@@ -135,9 +135,10 @@ func Provider() terraform.ResourceProvider {
 		},
 
 		DataSourcesMap: map[string]*schema.Resource{
-			"kubernetes_secret":        dataSourceKubernetesSecret(),
-			"kubernetes_service":       dataSourceKubernetesService(),
-			"kubernetes_storage_class": dataSourceKubernetesStorageClass(),
+			"kubernetes_secret":          dataSourceKubernetesSecret(),
+			"kubernetes_service":         dataSourceKubernetesService(),
+			"kubernetes_service_account": dataSourceKubernetesServiceAccount(),
```
dak1n1, line 140: Looks good! I imagine you'll have to rebase since this file was updated recently.

website/docs/d/service_account.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+layout: "kubernetes"
+page_title: "Kubernetes: kubernetes_service_account"
+sidebar_current: "docs-kubernetes-data-source-service-account"
+description: |-
+  A service account provides an identity for processes that run in a Pod.
+---
+
+# kubernetes_service_account
+
+A service account provides an identity for processes that run in a Pod.
```
dak1n1, line 11: This looks good. It's consistent with what we have in a lot of the docs. Let's just add a little extra to let the users know the difference between a service account and what this data source does:

```suggestion
A service account provides an identity for processes that run in a Pod. This data source reads the service account and makes specific attributes available to Terraform.
```

andrein:
@dak1n1, thanks for the thorough review, I'll rebase and address the issues!

Andrei

website/docs/d/service_account.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+```hcl
+data "kubernetes_service_account" "example" {
+  metadata {
+    name = "terraform-example"
+  }
+}
+
+data "kubernetes_secret" "example" {
+  metadata {
+    name = "${kubernetes_service_account.example.default_secret_name}"
```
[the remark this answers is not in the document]
andrein: Whoopsie, missed that one.  I think I originally meant to use ```resource "kubernetes_service_account" "example"```. Shall I change that?

kubernetes/data_source_kubernetes_service_account_test.go
andrein: Thanks for the DAMP pointer. Indeed, the test looks a lot better now!

PR #735, opened 2020-01-15
Do not force base64 encoding for the ca_bundle in the API Service resource

Symbianx:
This PR removes the need for the `ca_bundle` property of the `kubernetes_api_service` resource to be in base64 format.

When we encode it before applying in terraform we see that the resource created has the bundle encoded 2 times.

According to the 14th point of the document [Setup an Extension API Server ](https://kubernetes.io/docs/tasks/access-kubernetes-api/setup-extension-api-server/#setup-an-extension-api-server-to-work-with-the-aggregation-layer) it is not necessary to encode the bundle when using the `kube-aggregator` API to create the APIService:
> If using the kube-aggregator API, only pass in the PEM encoded CA bundle because the base 64 encoding is done for you.


Symbianx:
@alexsomesan Thanks for merging the PR, any idea on when this will be released?

Issue #734, opened 2020-01-15
Startup probes missing from Pod spec, even though readiness and liveness probes work


flyte:
We're able to configure `readiness_probe` and `liveness_probe` but [startup_probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes) is conspicuously absent from the code and documentation. Presumably it didn't exist at time of liveness/readiness probe implementation.

> invalid or unknown key: startup_probe

Could this please be added?

### Terraform Version
Terraform v0.11.13
+ provider.kubernetes v1.10.0

### Affected Resource(s)
Please list the resources as a list, for example:
- kubernetes_pod

### Terraform Configuration Files
n/a

### Debug Output
n/a

### Panic Output
n/a

### Expected Behavior
Startup probes should be configurable in exactly the same way as liveness and readiness probes.

### Actual Behavior
Startup probes aren't configurable using this terraform provider.

### Steps to Reproduce
Try adding a `startup_probe` section to your Pod spec.

### Important Factoids
n/a

### References
n/a

palra:
It looks quite simple to add, adding a `startup_probe` field in the `containerFields` function in that file, following the same schema than `readiness_probe` : https://github.com/terraform-providers/terraform-provider-kubernetes/blob/3c00798739ec767475e92b6243f00ee9766848be/kubernetes/schema_container.go#L471-L477
I'd love to contribute, as I need the same feature, but I have no idea how to test that. Is there anything we could do to help on that?

flyte:
I did some initial work on this, but I couldn't figure out how to get it working. I built it, but still couldn't get it to work in my stack. https://github.com/flyte/terraform-provider-kubernetes/commit/feb30e15f4d5ecbfe36ed290b944eabfb90a4cd2

jrhouston:
I got this working using kind and minikube in the above PR by enabling the `StartupProbe` [feature_gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/).

Enabling feature gates on kind: https://kind.sigs.k8s.io/docs/user/quick-start#enable-feature-gates-in-your-cluster

Enabling feature gates on minikube: 

```
minikube start --feature-gates "StartupProbe=true"
```

PR #739, opened 2020-01-17
Add support for startup_probe 

jrhouston:
Resolves https://github.com/terraform-providers/terraform-provider-kubernetes/issues/734

`StartupProbe` is behind a [feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) and so has to be enabled on the cluster's apiserver to see this working. If the feature gate isn't enabled, the API will accept this field but ignore it.

See `StartupProbe` under [v1.Container](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#container-v1-core)


kubernetes/structures_container.go
```diff
@@ -1,7 +1,7 @@
 package kubernetes
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
alexsomesan: This line is equivalent to one it replaces. The convention is to rename imports when there is a conflict between multiple import paths with the same last element. In this case we only import one package path ending in v1, so there is no need to use the renamed import form.

jrhouston: Oops, yeah this was my editor.

kubernetes/provider_test.go
```diff
@@ -223,25 +259,7 @@ func skipIfNotRunningInGke(t *testing.T) {
...
-	k8sVersion, err := gversion.NewVersion(serverVersion.String())
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	v1_14_0, _ := gversion.NewVersion("1.14.0")
-	if k8sVersion.LessThan(v1_14_0) {
-		t.Skip("The Kubernetes version must be 1.14.0 or newer for this test to run - skipping")
-	}
+	skipIfClusterVersionLessThan(t, "1.14.0")
```
jrhouston: Just refactoring duplicated code here.

jrhouston:
Ready for re-review.

PR #741, opened 2020-01-20
calico also annotates pods automatically, triggering a diff

hughobrien:
```
  # kubernetes_pod.nginx will be updated in-place
  ~ resource "kubernetes_pod" "nginx" {
        id = "hugh/nginx-example"

      ~ metadata {
          ~ annotations      = {
              - "cni.projectcalico.org/podIP" = "10.1.10.200/32" -> null
            }
```

### DIFF
--- a/kubernetes/structures.go
+++ b/kubernetes/structures.go
@@ -158,7 +158,7 @@ func isKeyInMap(key string, d map[string]interface{}) bool {
 
 func isInternalKey(annotationKey string) bool {
 	u, err := url.Parse("//" + annotationKey)
-	if err == nil && strings.HasSuffix(u.Hostname(), "kubernetes.io") {
+	if err == nil && (strings.HasSuffix(u.Hostname(), "kubernetes.io") || strings.HasSuffix(u.Hostname(), "projectcalico.org")) {
 		return true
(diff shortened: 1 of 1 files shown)


jrhouston:
Something you could do here is use the [ignore_changes](https://www.terraform.io/docs/configuration/resources.html#ignore_changes) lifecycle feature to ignore the tag. A caveat of this feature is that you have to create a placeholder for the field for which you are ignoring non-terraform changes:

```hcl
resource "kubernetes_pod" "test_pod" {
    metadata {
        name = "test-pod"

        annotations = {
            "cni.projectcalico.org/podIP" = ""
        }
    }

    spec {
        container {
            name = "test-container"
            image = "nginx"
        }
    }

    lifecycle {
        ignore_changes = [metadata.0.annotations["cni.projectcalico.org/podIP"]]
    }
}
```

hughobrien:
That looks ideal actually, thank you for the quick feedback. Should I close the PR? An optional provider level ignore list might be the more general solution anyway.

jrhouston:
You can close this PR. You could open an issue to discuss an optional provider level ignore list and what that might look like.

PR #744, opened 2020-01-22
Update kubernetes provider

Ivasan7:
Update kubernetes provider config with dynamic usage of master credentials

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -87,7 +87,7 @@ provider "kubernetes" {
   password = "password"
 }
 ```
-or dinamically based on container cluster:
+or dynamically based on container cluster:
 
 ```hcl
 resource "google_container_cluster" "cluster" {
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -87,7 +87,29 @@ provider "kubernetes" {
   password = "password"
 }
 ```
+or dinamically based on container cluster:
 
+```hcl
+resource "google_container_cluster" "cluster" {
+  name     = "my-gke-cluster"
+  location = "europe-west1"
+  remove_default_node_pool = true
+  initial_node_count       = 1
+  master_auth {
+    username = ""
+    password = ""
+    client_certificate_config {
+      issue_client_certificate = true
+    }
+  }
+}
+provider "kubernetes" {
+  host = google_container_cluster.cluster.endpoint
+  client_certificate     = "${base64decode(google_container_cluster.cluster.master_auth.0.client_certificate)}"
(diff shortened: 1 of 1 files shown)


website/docs/index.html.markdown
```diff
@@ -87,7 +87,29 @@ provider "kubernetes" {
   password = "password"
 }
 ```
+or dinamically based on container cluster:
```
mewa: Got a typo here (dynamically)

Ivasan7:
@mewa bump

alexsomesan:
@Ivasan7 Thanks a lot for the effort to put this example together.

However, this illustrates a use-case we are actually advising against.
This is because of particularities in which Terraform core itself evaluates value interpolations. Provider blocks have their attributes evaluated much earlier in the `apply` run as the attributes of resources. This is counter intuitive at first and not immediately apparent but it leads to inconsistent behaviour when used exactly like your example shows. The keywords here are `same apply run`. This setup will work just fine as long as the `google_container_cluster` is created in a previous `apply` and already present in the state.

I'm happy to include such an example (in fact been meaning to) if it correctly illustrates this as an anti-pattern. Would you be willing to phrase it like that?

Ivasan7:
I understand, then indeed it is not something we would like to endorse. 

in summary if I understand correctly, then using static secrets would be a way to go and dynamic certificates and signing mechanism is advised against. Actually would not work in case if you apply it in the same step, contrary to how I have used it, added it later.

Good to know, thank you for you reply and explanation.

jaceq:
@alexsomesan This isn't right. This is what documentation says but I use it exactly like that (with GKE outputs used to configure kubernetes provider in the same state) and it works 100/100. I use this setup to build parallel environments couple times a month - never had a problem with such setup.

Ivasan7:
@alexsomesan @jaceq  indeed. Even in the official provider example they give this option:
https://github.com/GoogleCloudPlatform/terraform-google-examples/blob/master/example-gke-k8s-helm/helm.tf

aareet:
@Ivasan7 thank you for pointing out that example. We will reach out to the maintainers to improve the example to reflect our recommended method of using static credentials.

For this PR, would you be willing to rephrase this config as an example of an antipattern, as advised by our official documentation (https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#stacking-with-managed-kubernetes-cluster-resources)?

aareet:
Just jumping in here to follow up as it has been about 2 weeks. I'll close this for now. Please feel free to reopen if you'd like to contribute this as an antipattern. Thank you for this PR regardless.

Ivasan7:
@aareet for me this works, and the antipattern behaviour is not reproducible, therefore I am reluctant to add it as an antipattern.

PR #745, opened 2020-01-22
Add ttlSecondsAfterFinished to kubernetes_job and kubernetes_cron_job

### LINKED ISSUE #736: Support ttlSecondsAfterFinished in JobSpec
[ttlSecondsAfterFinished](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#ttl-mechanism-for-finished-jobs) was added to the job spec in 1.12.

### Terraform Version
0.12.19

### Affected Resource(s)
- kubernetes_job
- kubernetes_cron_job

jrhouston:
Resolves #736


alexsomesan:
Just ran into this: https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/
It looks like this feature is dependent on an admission controller being deployed and active in the cluster. We might need to figure out a way to detect its presence and add that to the skip function.
I just had the test fail on out-of-the-box minikube which doesn't run this controller.
WDYT?

PR #753, opened 2020-01-31
Honore load_config_file parameter

### LINKED ISSUE #752: load_config_file parameter not honored.
### Terraform Version
```
Terraform v0.12.12
+ provider.kubernetes v1.10.0
```

### Affected Resource(s)
All kubernetes resources that needs at some point to connect to a kubernetes cluster. 

```hcl
provider "kubernetes" {
  version          = "~> 1.5"
  load_config_file = false

  host                   = <my host for cluster A>
  token                  = <my token>
  cluster_ca_certificate = <my ca_cert>
}

resource "kubernetes_secret" "test-secret" {
    data = <REDACTED>
    type = "Opaque"

    metadata {
        name      = "a-secret"
        namespace = "a-namespace"
    }
}
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Expected Behavior
The local configuration should not be loaded.

### Actual Behavior
The local configuration is loaded.

#### Case 1:
if I run a `terraform plan` with an existing `~/.kube/config` file with the cluster A set as my current context, plan works.

#### Case 2:
If I change my context to cluster B through `~/.kube/config` (note: without changing the provider configuration nor my kubernetes resources), plan succeed but show me differences:

```
  + create

  # kubernetes_secret.test-secret will be created
  + resource "kubernetes_secret" "test-secret" {
      + data = (sensitive value)
      + id   = (known after apply)
      + type = "Opaque"

      + metadata {
          + generation       = (known after apply)
          + labels           = {
              + "jenkins.io/credentials-type" = "secretFile"
            }
          + name             = "a-secret"
          + namespace        = "a-namespace"
          + resource_version = (known after apply)
          + self_link        = (known after apply)
          + uid              = (known after apply)
        }
    }

```

That secret already exist in cluster A but not in cluster B.

#### Case 3:
If I remove my local config file (`mv ~/.kube/config ~/.kube/config.bk` for instance), the `terraform plan` fails with:

```
Error: Get http://localhost/api/v1/namespaces/a-namespace/secrets/a-secret: dial tcp [::1]:80: connect: connection refused
```

It tries to reach the local kubernetes cluster which is the default for this provider.

----------

Note that if you reuse _case 1_ context (a local configuration file and current-context pointing to the right cluster) and run the plan with `KUBE_LOAD_CONFIG_FILE=false`, you end up with the same error as case 3.

### Important Factoids
My takes here are:
- Setting `load_config_file` parameter in configuration file to `false` is not honored as shown in my examples.
- Defaulting the environment variable to true if it is not defined [here](https://github.com/terraform-providers/terraform-provider-kubernetes/blob/master/kubernetes/provider.go#L104) does not play well with the `load_config_file` being optional.

This default setting eases local development (you usually have kubernetes context on localhost, usually pointing to the kubernetes cluster your are working on) but hides issues: 

When you run the same terraform code on a different host, for instance in your CI, you will end up with the same error as in case 3. You do not necessarily have a kubeconfig on this other host and the provider will to contact a cluster on localhost because it is the default setting.

tgermain:
This parameter could be defined either:
- via provider configuration file
- via an environment variable

Previous behavior when the environment variable is not define is to
default it to true, completly by-passing the provider configuration
file.

fix #752

### DIFF
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -101,7 +101,7 @@ func Provider() terraform.ResourceProvider {
 			"load_config_file": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("KUBE_LOAD_CONFIG_FILE", true),
+				DefaultFunc: schema.EnvDefaultFunc("KUBE_LOAD_CONFIG_FILE", nil),
 				Description: "Load local kubeconfig.",
 			},
 			"exec": {
@@ -194,7 +194,8 @@ func providerConfigure(d *schema.ResourceData, terraformVersion string) (interfa
 
 	var cfg *restclient.Config
 	var err error
-	if d.Get("load_config_file").(bool) {
+
+	if v, ok := d.GetOk("load_config_file"); ok && v.(bool) {
 		// Config file loading
 		cfg, err = tryLoadingConfigFile(d)
 	}


alexsomesan:
Hello, 

I apologize if this particular configuration parameter was confusing. Please note that the documentation clearly explains the current defaulting behaviour, as seen here: https://www.terraform.io/docs/providers/kubernetes/index.html#load_config_file

For the moment, I think the best choice is to leave things as they are since this is change does not fix a malfunction, but rather alters a working feature. I estimate that introducing this would affect a lot more users who have build configurations around the current behaviour.

Let's use issue 752 to get to the bottom of the situation.

PR #762, opened 2020-02-14
Fix getting-started documentation code bloc formatting

thomas-riccardi:
The markdown rendering on terraform.io/docs/ seems to differ from the one in github: after `:\n` we need an extra empty line to start a code bock.


thomas-riccardi:
Current rendering on https://www.terraform.io/docs/providers/kubernetes/guides/getting-started.html :

image

js-timbirkett:
Looks like duplicate of #771 which is now merged. Nice to see so many of us caring about the docs 👀

alexsomesan:
Indeed this was fixed through #771

PR #765, opened 2020-02-15
Break line before code block

andredurao:
The code block right below was missing formatting because of that

You can check it on [https://www.terraform.io/docs/providers/kubernetes/guides/getting-started.html](https://www.terraform.io/docs/providers/kubernetes/guides/getting-started.html)

image


andredurao:
Woops, just realized thereś another PR for the same fix: https://github.com/terraform-providers/terraform-provider-kubernetes/pull/762

PR #796, opened 2020-03-16
Fix modules for Go version 1.13

dak1n1:
### Purpose
The purpose of this PR is to fix some errors that are seen when running `go mod download` on this repo for the first time.

### Steps to reproduce:

1. Move GOPATH to a different location temporarily, so we can start with a fresh, empty modules path. For example: `mv ~/go ~/go.backup; mkdir ~/go`

2. Create a fresh clone of this repo outside of the GOPATH.

3. `cd` into this repo and run `go mod download`.

Full output of the above 3 commands are here: https://gist.github.com/dak1n1/5a955dfd8f800978693195fcf6059c68

This is a summary of the error encountered:

```
github.com/terraform-providers/terraform-provider-aws@v2.32.0+incompatible: invalid version: +incompatible suffix not allowed: module contains a go.mod file, so semantic import versioning is required
github.com/terraform-providers/terraform-provider-google@v2.17.0+incompatible: invalid version: +incompatible suffix not allowed: module contains a go.mod file, so semantic import versioning is required
github.com/terraform-providers/terraform-provider-random@v2.2.1+incompatible: invalid version: +incompatible suffix not allowed: module contains a go.mod file, so semantic import versioning is required
k8s.io/client-go@v12.0.0+incompatible: invalid version: +incompatible suffix not allowed: module contains a go.mod file, so semantic import versioning is required
```

### Solution:

In this PR, I've removed the `+incompatible` suffix from repos that are no longer incompatible. Since these repos now contain a `go.mod`, there is no need to skip version validation (and actually the presence of the `go.mod` file in these repos is causing an error during `go mod download` because of Go 1.13's version validation).  https://golang.org/doc/go1.13#version-validation

When removing `+incompatible`, I chose pseudo-versions that point to the same commit as before in each respective git repo. This shouldn't impact any part of the code, and should make Go 1.13 happy.

https://tip.golang.org/cmd/go/#hdr-Pseudo_versions


dak1n1:
@jrhouston I just wanted to run this PR idea by you, since this is something I'm working on. There are still issues with the tests, so this PR is a work in progress. I might have to choose a different version for the k8s client-go and related dependencies. I just wanted to check in with you early to make sure I'm on the right path. Thanks,

go.mod
```diff
@@ -14,14 +14,11 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/pierrec/lz4 v2.3.0+incompatible // indirect
 	github.com/robfig/cron v1.2.0
-	github.com/terraform-providers/terraform-provider-aws v2.32.0+incompatible
```
alexsomesan: These providers are needed for the tests to work. 
Most likely why you are having issues with tests.

dak1n1: Thanks. I put these back and used the pseudo-version to point to the same commit as before.

go.mod
```diff
@@ -14,21 +15,24 @@ require (
...
+	github.com/terraform-providers/terraform-provider-aws v1.60.1-0.20191010190908-1261a98537f2
+	github.com/terraform-providers/terraform-provider-google v1.20.1-0.20191008212436-363f2d283518
+	github.com/terraform-providers/terraform-provider-random v1.3.2-0.20190925210718-83518d96ae4f // indirect
 	github.com/ulikunitz/xz v0.5.6 // indirect
 	github.com/vmihailenco/msgpack v4.0.4+incompatible // indirect
 	k8s.io/api v0.0.0-20191025225708-5524a3672fbb
 	k8s.io/apimachinery v0.0.0-20191025225532-af6325b3a843
-	k8s.io/client-go v12.0.0+incompatible
-	k8s.io/kube-aggregator v0.0.0-20191025230902-aa872b06629d
+	k8s.io/client-go v10.0.0+incompatible
```
dak1n1: `client-go` kept coming back when I ran `go mod tidy`, so I guess something is pulling it in? But I forced the right version in the `replace` statement below.

alexsomesan:
I tried running these changes in TC and the `test-compile` build target fails.
This means that there are likely still issues with the module dependencies of tests.

dak1n1:
Oops, I had forgotten to `git add` my local `vendor` dir. I think the tests will pass now. I'll have to wait until tomorrow when someone can add me to TC. I don't have access via oauth.

dak1n1:
TC tests are passing now.

dak1n1:
Rebased to fix merge conflicts, since go.mod was changed in master. I'll re-run the tests.

dak1n1:
TC test is passing.

dak1n1:
Oh, and I ran `go mod vendor` but there was nothing to update, so no changes were made. Probably because it's using the same module versions as before, and just referring to them by their pseudo-version tags.

PR #90, opened 2017-11-22
New datasource: kubernetes_config_map

sl1pm4t:
Fixes #76

Issue #76, opened 2017-10-16
Feature Request: ConfigMap data source


bonifaido:
I think it would be beneficial to read ConfigMaps via Terraform, in my use case I would like to read the configmap that Calico provides: `calico-config` to be able to create new resources with Terraform based on the data in that ConfigMap.

bluemalkin:
Any update on this ? Would love to see it released asap. Thanks !

bmcstdio:
I am not sure why #90 was close, but I could probably work on this as I'd also benefit from it 🙂

bonifaido:
Me neither :), it looks totally implemented (but closed).

justinm:
I have a scenario where I need to be able to read configMaps into Terraform. +1 for this feature.

jeremych1000:
Would it be possible to reopen the PR? We already have a `kubernetes_secret` data source, but no `kubernetes_config_map` data source.

dak1n1:
Hi, everyone! This feature has merged and will appear in the next release. Please let us know if it's still an issue after upgrading your provider. Thanks!

PR #801, opened 2020-03-28
Add data source for reading config maps

dak1n1:
Add data source for reading config maps.
Based on PR https://github.com/terraform-providers/terraform-provider-kubernetes/pull/90.
Solves https://github.com/terraform-providers/terraform-provider-kubernetes/issues/76.

### DIFF
--- a/kubernetes/data_source_kubernetes_config_map_test.go
+++ b/kubernetes/data_source_kubernetes_config_map_test.go
@@ -1,0 +1,49 @@
+package kubernetes
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+)
+
+func TestAccKubernetesDataSourceConfigMap_basic(t *testing.T) {
+	name := fmt.Sprintf("tf-acc-test-%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccKubernetesDataSourceConfigMapConfig_basic(name),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("data.kubernetes_config_map.test", "metadata.0.name", name),
+					resource.TestCheckResourceAttrSet("data.kubernetes_config_map.test", "metadata.0.generation"),
+					resource.TestCheckResourceAttrSet("data.kubernetes_config_map.test", "metadata.0.resource_version"),
+					resource.TestCheckResourceAttrSet("data.kubernetes_config_map.test", "metadata.0.self_link"),
+					resource.TestCheckResourceAttrSet("data.kubernetes_config_map.test", "metadata.0.uid"),
+					resource.TestCheckResourceAttr("data.kubernetes_config_map.test", "metadata.0.annotations.%", "2"),
+					resource.TestCheckResourceAttr("data.kubernetes_config_map.test", "metadata.0.annotations.TestAnnotationOne", "one"),
+					resource.TestCheckResourceAttr("data.kubernetes_config_map.test", "metadata.0.annotations.TestAnnotationTwo", "two"),
+					resource.TestCheckResourceAttr("data.kubernetes_config_map.test", "metadata.0.labels.TestLabelOne", "one"),
+					resource.TestCheckResourceAttr("data.kubernetes_config_map.test", "metadata.0.labels.TestLabelTwo", "two"),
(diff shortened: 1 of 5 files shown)


dak1n1:
TC test is still running. I did some some local tests using terraform 0.12.23 and a `kind` cluster.

One way I used the new data source was by creating a pod and placing one of the config map's values into an environment variable. It works for both binary data and regular data (even though I can't think of a usage for the binary data offhand). Since the field exists, it makes sense for the data source to be able to read it.

Usage example from my testing:

This populates the `data` field in the config map, which you can then read using the data source.
```
kubectl create configmap testcm --from-literal=key1=foo
```

This makes a `binaryData` field in the config map, if you pass it a file that's specifically not encoded in UTF-8. That can be read from the data source using the `binary_data` attribute.

```
kubectl create configmap testcm --from-file=key1=utf8test.txt
```

Example terraform config that reads from the `kubernetes_config_map` data source:

```
provider "kubernetes" {
  config_context_auth_info = "kind-kind"
  config_context_cluster   = "kind-kind"
}

data "kubernetes_config_map" "example" {
  metadata {
    name = "testcm"
  }
}

resource "kubernetes_pod" "test" {
  metadata {
    name = "testpod"
  }
  spec {
    container {

      env {
        name = "testvar"
        value = data.kubernetes_config_map.example.data["key1"]
      }

      image = "nginx:1.7.8"
      name  = "example"

      resources {
        limits {
          cpu    = "0.5"
          memory = "512Mi"
        }
        requests {
          cpu    = "250m"
          memory = "50Mi"
        }
      }
    }
  }
}
```

kubernetes/data_source_kubernetes_config_map_test.go
```diff
@@ -0,0 +1,49 @@
...
+					resource.TestCheckResourceAttr("data.kubernetes_config_map.test", "data.one", "first"),
+					resource.TestCheckResourceAttr("data.kubernetes_config_map.test", "data.two", "second"),
+				),
+			},
+		},
+	})
+}
+
+func testAccKubernetesDataSourceConfigMapConfig_basic(name string) string {
+	return testAccKubernetesConfigMapConfig_basic(name) + `
```
jrhouston: This tripped me up a little bit, as at first glance it seems like there's a bunch of checks in the test that aren't in the fixture.

I get that here we're trying to re-use the code from the `ResourceConfigMapConfig_basic` test. This makes sense to save writing code, but then we're creating a dependency between these two tests such that we have to keep their checks in sync when we update the underlying fixture. 

I find it's better to keep acceptance tests as explicit and isolated as possible, so that you can look at the test code and easily understand what is being tested and the expected behaviour without having to step through too much code. There is a pretty horrible acronym [DAMP](https://stackoverflow.com/a/11837973), as an antonym to DRY, that somewhat applies here.

What I would do here is explicitly code a `config_map` resource here, and maybe make it simpler so we don't have so many checks above. We are already testing all things that the `resourceKubernetesConfigMapRead` function does in that other test, so we don't need to re-test them here. 

edit: I just noticed you took this from the closed PR you linked, but I would have left the same comment on that PR 😄

website/docs/d/config_map.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+sidebar_current: "docs-kubernetes-resource-config map"
+description: |-
+  This data source reads configuration data from a config map.
+---
+
+# kubernetes_config map
+
+Config Maps are key-value pairs containing configuration data. The Config Map data source provides a mechanism for extracting these key-value pairs.
+
+~> **Note:** All arguments including the config map data will be stored in the raw state as plain-text. [Read more about sensitive data in state](/docs/state/sensitive-data.html).
```
jrhouston: I think this note only applies to the ConfigMap resource, not its data source.

dak1n1: Good point. I actually copied all this from the Secret data source, so I should probably also fix that in the Secret data source.

dak1n1: Oh actually, look at this: the data source stores the data in the tfstate.

```
    {
      "mode": "data",
      "type": "kubernetes_config_map",
      "name": "example",
      "provider": "provider.kubernetes",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "binary_data": {
              "key1": "VVRGLTggZGVjb2RlciBjYXBhYmlsaXR5IGFuZCBzdHJlc3MgdGV
...
```

So I think it makes sense to leave this warning here, so users are aware that any data they read in via the data source is going to be stored in plaintext. Even though I know it's bad practice to put sensitive data into a config map, I've seen it in production at another place I worked. What do you think?

jrhouston: Yep you're right, makes sense to leave it in.

kubernetes/data_source_kubernetes_config_map_test.go
dak1n1: Thanks, I really like this idea! I honestly had a pretty hard time understanding what this code did and wasn't happy with it. But I simplified it now.

Issue #791, opened 2020-03-10
Feature Request: MutatingWebhookConfiguration and ValidatingWebhookConfiguration resources


ibrahimalihc:
### Community Note

* Please vote on this issue by adding a 👍 reaction to the original issue to help the community and maintainers prioritize this request.
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Description
It is not currently possible to deploy applications such as the [Consul Connect injector](https://www.consul.io/docs/platform/k8s/connect.html#installation-and-configuration) or [Open Policy Agent](https://www.openpolicyagent.org/docs/latest/kubernetes-introduction/) using the Terraform Kubernetes provider because there are no resources for deploying a [MutatingWebhookConfiguration](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) and [ValidatingWebhookConfiguration](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook).

### References

* https://kubernetes.io/blog/2019/03/21/a-guide-to-kubernetes-admission-controllers/
* https://github.com/hashicorp/consul-helm/blob/master/templates/connect-inject-mutatingwebhook.yaml
* https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#mutatingwebhookconfiguration-v1beta1-admissionregistration-k8s-io
* https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#validatingwebhookconfiguration-v1beta1-admissionregistration-k8s-io

dak1n1:
:tada: These features have been added to the current release, 1.1.2. (Linking PR [here](https://github.com/terraform-providers/terraform-provider-kubernetes/pull/829) for reference).

PR #806, opened 2020-03-31
Add support for ValidatingWebhookConfiguration

jrhouston:
Partly solves #791 

**TODO**

- [x] Add update function
- [x] Add acceptance test
- [x] Add documentation


go.mod
```diff
@@ -36,7 +36,7 @@ require (
 // and https://github.com/Azure/go-autorest/issues/481
 replace (
 	github.com/Azure/go-autorest v11.1.2+incompatible => github.com/Azure/go-autorest v12.1.0+incompatible
-	k8s.io/client-go => k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
+	k8s.io/client-go => k8s.io/client-go v0.0.0-20190918160344-1fbdaa4c8d90
```
jrhouston, line 39: I had to update to a newer version of client-go to get the AdmissionregistratioV1() API. It took me a minute to realize the numbers before the commit SHA are the commit date 🤦‍♂

kubernetes/structure_validating_webhook_configuration.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	if v, ok := in["api_versions"].([]interface{}); ok {
+		obj.APIVersions = expandStringSlice(v)
+	}
+
+	if v, ok := in["operations"].([]interface{}); ok {
+		for _, op := range v {
+			if op != nil {
+				obj.Operations = append(obj.Operations, admissionregistrationv1.OperationType(op.(string)))
```
jrhouston, line 123: if only there was a way to have an expandSlice function that worked for any kubernetes type :trollface:

jrhouston:
Still to add documentation but the code is ready for review.

go.mod
dak1n1: I learned a helpful command to help get this pseudo-version right. You might already know it. Once you have the commit SHA you want, you can run `go get k8s.io/client-go 1fbdaa4c8d90` and it'll fill out the pseudo-version for you. https://tip.golang.org/cmd/go/#hdr-Pseudo_versions

jrhouston: Nice! Thanks

kubernetes/resource_kubernetes_validating_webhook_configuration.go
```diff
@@ -0,0 +1,234 @@
...
+			"webhook": {
+				Type:        schema.TypeList,
+				Description: "A list of webhooks and the affected resources and operations.",
+				Required:    true,
+				MinItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"admission_review_versions": {
+							Type:        schema.TypeList,
+							Description: `AdmissionReviewVersions is an ordered list of preferred AdmissionReview versions the Webhook expects. API server will try to use first version in the list which it supports. If none of the versions specified in this list supported by API server, validation will fail for this object. If a persisted webhook configuration specifies allowed versions and does not include any versions known to the API Server, calls to the webhook will fail and be subject to the failure policy.`,
```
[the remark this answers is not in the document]
jrhouston: You're right, I just put these descriptions in backticks because there's often a quote in there somewhere.

dak1n1:
Interestingly, the acceptance tests pass locally. It must be environment specific... I'm looking into it to see if I can get it to fail on my laptop like it did in TC.

```
[dakini@dax terraform-provider-kubernetes (validating-webhook=)]$ cd kubernetes ; TF_ACC=true go test -v -run TestAccKubernetesValidatingWebhookConfiguration_basic ; cd -
=== RUN   TestAccKubernetesValidatingWebhookConfiguration_basic
--- PASS: TestAccKubernetesValidatingWebhookConfiguration_basic (0.56s)
PASS
ok      github.com/terraform-providers/terraform-provider-kubernetes/kubernetes 0.578s
```

jrhouston:
> I'm still reading through a lot of this, but I wanted to leave a comment to say that the acceptance tests have failed in TC. https://ci-oss.hashicorp.engineering/viewLog.html?buildId=104951&buildTypeId=Kubernetes_ProviderKubernetesGke115
> 
> ```
> TestAccKubernetesValidatingWebhookConfiguration_basic
>  
> First failure: 	806	#1322569 	Changes (15) 	01 Apr 20 16:44
> ------- Stdout: -------
> === RUN   TestAccKubernetesValidatingWebhookConfiguration_basic
> --- FAIL: TestAccKubernetesValidatingWebhookConfiguration_basic (0.48s)
>     testing.go:654: Step 0 error: errors during apply:
>         
>         Error: the server could not find the requested resource
>         
>           on /opt/teamcity-agent/temp/buildTmp/tf-test917478516/main.tf line 2:
>           (source code not available)
> ```
> 
> ```
> TestAccKubernetesValidatingWebhookConfiguration_importBasic
>  
> Current failure:	806	#1322569 	Changes (15) 	01 Apr 20 16:44
>  	First failure: 	Calculating ...
> ------- Stdout: -------
> === RUN   TestAccKubernetesValidatingWebhookConfiguration_importBasic
> --- FAIL: TestAccKubernetesValidatingWebhookConfiguration_importBasic (0.37s)
>     testing.go:654: Step 0 error: errors during apply:
>         
>         Error: the server could not find the requested resource
>         
>           on /opt/teamcity-agent/temp/buildTmp/tf-test361925425/main.tf line 2:
>           (source code not available)
> ```

Ah I just realized, `v1` of this API is in kubernetes 1.16. It's `v1beta1` in 1.15, which is what the latest version on our CI.

jrhouston:
I have changed this PR to use `admissionregistration/v1beta1`. This is supported all the way to Kubernetes 1.18. We can add a shim to support the v1 client when `v1beta` is actually removed from the Kubernetes API completely in a future release.

kubernetes/resource_kubernetes_validating_webhook_configuration.go
```diff
@@ -0,0 +1,310 @@
...
+	cfg := admissionregistrationv1.ValidatingWebhookConfiguration{
+		ObjectMeta: expandMetadata(d.Get("metadata").([]interface{})),
+		Webhooks:   expandValidatingWebhooks(d.Get("webhook").([]interface{})),
+	}
+
+	log.Printf("[INFO] Creating new ValidatingWebhookConfiguration: %#v", cfg)
+
+	res := &admissionregistrationv1.ValidatingWebhookConfiguration{}
+
+	if useAdmissionregistrationV1beta1(conn) {
```
jrhouston, line 133: @dak1n1 @alexsomesan Can I get your thoughts on this approach to falling back to the old version of the API. I'm trying to avoid duplicating tons of code with the types from `v1beta1` like I had to do with the HPA. Alternatives here are to use unsafe.Pointer (😱) or duplicate all the code to support both types. If there is an easier way to do this that I've missed please let me know.


This code works because `v1` and `v1beta1` have identical struct fields, it is just the struct tags that have changed. Here is the diff:

```
~/d/k/a/admissionregistration ❯❯❯ diff v1/types.go v1beta1/types.go
2c2
< Copyright 2019 The Kubernetes Authors.
---
> Copyright 2017 The Kubernetes Authors.
17c17
< package v1
---
> package v1beta1
117a118
> // Deprecated in v1.16, planned for removal in v1.19. Use admissionregistration.k8s.io/v1 ValidatingWebhookConfiguration instead.
147a149
> // Deprecated in v1.16, planned for removal in v1.19. Use admissionregistration.k8s.io/v1 MutatingWebhookConfiguration instead.
195c197
< 	// allowed values are Ignore or Fail. Defaults to Fail.
---
> 	// allowed values are Ignore or Fail. Defaults to Ignore.
212c214
< 	// Defaults to "Equivalent"
---
> 	// Defaults to "Exact"
277c279
< 	// Acceptable values are: None, NoneOnDryRun (webhooks created via v1beta1 may also specify Some or Unknown).
---
> 	// Acceptable values are: Unknown, None, Some, NoneOnDryRun
281,282c283,285
< 	// sideEffects == Unknown or Some.
< 	SideEffects *SideEffectClass `json:"sideEffects" protobuf:"bytes,6,opt,name=sideEffects,casttype=SideEffectClass"`
---
> 	// sideEffects == Unknown or Some. Defaults to Unknown.
> 	// +optional
> 	SideEffects *SideEffectClass `json:"sideEffects,omitempty" protobuf:"bytes,6,opt,name=sideEffects,casttype=SideEffectClass"`
288c291
< 	// Default to 10 seconds.
---
> 	// Default to 30 seconds.
299c302,304
< 	AdmissionReviewVersions []string `json:"admissionReviewVersions" protobuf:"bytes,8,rep,name=admissionReviewVersions"`
---
> 	// Default to `['v1beta1']`.
> 	// +optional
> 	AdmissionReviewVersions []string `json:"admissionReviewVersions,omitempty" protobuf:"bytes,8,rep,name=admissionReviewVersions"`
324c329
< 	// allowed values are Ignore or Fail. Defaults to Fail.
---
> 	// allowed values are Ignore or Fail. Defaults to Ignore.
341c346
< 	// Defaults to "Equivalent"
---
> 	// Defaults to "Exact"
406c411
< 	// Acceptable values are: None, NoneOnDryRun (webhooks created via v1beta1 may also specify Some or Unknown).
---
> 	// Acceptable values are: Unknown, None, Some, NoneOnDryRun
410,411c415,417
< 	// sideEffects == Unknown or Some.
< 	SideEffects *SideEffectClass `json:"sideEffects" protobuf:"bytes,6,opt,name=sideEffects,casttype=SideEffectClass"`
---
> 	// sideEffects == Unknown or Some. Defaults to Unknown.
> 	// +optional
> 	SideEffects *SideEffectClass `json:"sideEffects,omitempty" protobuf:"bytes,6,opt,name=sideEffects,casttype=SideEffectClass"`
417c423
< 	// Default to 10 seconds.
---
> 	// Default to 30 seconds.
428c434,436
< 	AdmissionReviewVersions []string `json:"admissionReviewVersions" protobuf:"bytes,8,rep,name=admissionReviewVersions"`
---
> 	// Default to `['v1beta1']`.
> 	// +optional
> 	AdmissionReviewVersions []string `json:"admissionReviewVersions,omitempty" protobuf:"bytes,8,rep,name=admissionReviewVersions"`
```

dak1n1: How about making a helper function that will return the WebhookConfig for the appropriate API version? Something like this:

```
// createWebhookConfiguration will determine the kubernetes API version and generate a compatible webhook configuration.
func createWebhookConfiguration(d *schema.ResourceData, meta interface{}) (interface{}, error) {

	conn, err := meta.(KubeClientsets).MainClientset()
	if err != nil {
		return nil, err
	}

	if useAdmissionregistrationV1beta1(conn) {
		return &admissionregistrationv1beta1.ValidatingWebhookConfiguration{}, nil
	} else {
		return &admissionregistrationv1.ValidatingWebhookConfiguration{}, nil
	}
}
```

Then you can keep the CRUD functions agnostic of the API version. For example:

```
func resourceKubernetesValidatingWebhookConfigurationRead(d *schema.ResourceData, meta interface{}) error {

	conn, err := meta.(KubeClientsets).MainClientset()
	if err != nil {
		return err
	}

	name := d.Id()

	cfg, err := createWebhookConfiguration(d, meta)
	if err != nil {
		return err
	}
```

Then you'd use the variable `cfg` without needing to know its API version. If that isn't possible (for example, if you need to run an API version-specific operation on it), then we could have the helper function return an object that is version-specific to run the CRUD operations on. Just an idea.

jrhouston: I tried something like this initially, but I ran into the problem that all the expander and flattener functions are using `admissionregistration/v1` in their signatures, and you can't cast or convert types from one API version to the other, even if they are `interface{}`. The problem I'm trying to avoid is duplicating all those functions so we have signatures for both v1 and v1beta1. 😢

dak1n1: Today I was trying to figure out if we could solve this duplication issue by creating a kind of compatibility layer between the two API versions, similar to what I did in a different project.

My compatibility layer in that other project was basically a struct that contained an interface. And that interface could hold either of the two objects I was working with. That allowed my functions to accept either one of those similar objects interchangeably. The two interchangeable objects were an AWS S3 client and an AWS S3 mock client, but the same principle could apply here if we had a struct that holds an interface resembling the `ValidatingWebhookConfiguration`. That way you could squeeze either a `v1.ValidatingWebhookConfiguration` or a `v1beta.ValidatingWebhookConfiguration` into the same struct, and use that struct in version-agnostic functions.

Here's where I made that compatibility thing, in case it helps to see what I'm talking about. https://github.com/dak1n1/managed-velero-operator/blob/89f07878d7eebe4cfea721246a3ef635f8fca9f2/pkg/s3/client.go#L30-L55

I tried coding up something like that today, but it got pretty messy and I think I got in over my head with that. It looks like more trouble than it's worth, but I wanted to share it in case it sparked any ideas. Some code duplication might be more straightforward to maintain, compared to creating a compatibility layer abstraction. :thinking: 

It might even be neater to have different CRUD functions for the v1 API vs the v1beta API. That way it'll be really quick and easy to remove when the backwards-compatibility is no longer needed. I think as long as the code duplication is separated out into easy-to-understand (and easy-to-remove) blocks, it could be good to just duplicate it.

website/docs/r/validating_webhook_configuration.html.markdown
```diff
@@ -0,0 +1,129 @@
+---
+layout: "kubernetes"
+page_title: "Kubernetes: validating_webhook_configuration"
+sidebar_current: "docs-kubernetes-resource-validating-webhook-configuration"
+description: |-
+  Validating Webhook Configuration automatically scales the number of pods in a replication controller, deployment or replica set based on observed CPU utilization.
```
dak1n1, line 6: Is this the right description for the `validating_webhook_configuration` resource?

website/docs/r/validating_webhook_configuration.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+---
+
+# validating_webhook_configuration
+
+Validating Webhook Configuration configures a [validating admission webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks).
+
+## Example Usage
+
+```hcl
+resource "validating_webhook_configuration" "example" {
```
dak1n1, line 16: In my testing, this needed `kubernetes_` in front of it:

```suggestion
resource "kubernetes_validating_webhook_configuration" "example" {
```

website/docs/r/validating_webhook_configuration.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+        namespace = "example-namespace"
+        name = "example-service"
+      }
+    }
+
+    rule {
+      api_groups = ["apps"]
+      api_versions = ["v1"]
+      operations = ["CREATE"]
+      resources = "deployments"
```
dak1n1, line 37: `terraform validate` is telling me this should be a list of strings instead:

```suggestion
      resources = ["deployments"]
```

website/docs/r/validating_webhook_configuration.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+      scope = ["Namespaced"]
```
dak1n1, line 38: `terraform validate` is telling me this should be a string instead:
```suggestion
      scope = "Namespaced"
```

website/docs/r/validating_webhook_configuration.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+## Example Usage
+
+```hcl
+resource "validating_webhook_configuration" "example" {
+  metadata {
+    name = "test.terraform.io"
+  }
+
+  webhook {
+    name = "test.terraform.io"
```
dak1n1, line 22: I was unable to apply this configuration without specifying the `side_effect` attribute:
```suggestion
    name = "test.terraform.io"
    side_effects = "None"
```

website/docs/r/validating_webhook_configuration.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+* `rule` - (Optional) Describes what operations on what resources/subresources the webhook cares about. The webhook cares about an operation if it matches _any_ Rule. However, in order to prevent ValidatingAdmissionWebhooks and MutatingAdmissionWebhooks from putting the cluster in a state which cannot be recovered from without completely disabling the plugin, ValidatingAdmissionWebhooks and MutatingAdmissionWebhooks are never called on admission requests for ValidatingWebhookConfiguration and MutatingWebhookConfiguration objects.
+* `sideEffects` - (Optional) SideEffects states whether this webhook has side effects. Acceptable values are: None, NoneOnDryRun (webhooks created via v1beta1 may also specify Some or Unknown). Webhooks with side effects MUST implement a reconciliation system, since a request may be rejected by a future step in the admission change and the side effects therefore need to be undone. Requests with the dryRun attribute will be auto-rejected if they match a webhook with sideEffects == Unknown or Some.
```
dak1n1, line 92: I think this is supposed to be `side_effects`, since that's how I had to specify it in my main.tf.

website/docs/r/validating_webhook_configuration.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+* `timeoutSeconds` - (Optional) TimeoutSeconds specifies the timeout for this webhook. After the timeout passes, the webhook call will be ignored or the API call will fail based on the failure policy. The timeout value must be between 1 and 30 seconds. Default to 10 seconds.
```
dak1n1, line 93:
```suggestion
* `timeout_seconds` - (Optional) Specifies the timeout for this webhook. After the timeout passes, the webhook call will be ignored or the API call will fail based on the failure policy. The timeout value must be between 1 and 30 seconds. Default to 10 seconds.
```

website/docs/r/validating_webhook_configuration.html.markdown
dak1n1, line 92:
```suggestion
* `side_effects` - (Optional) States whether this webhook has side effects. Acceptable values are: None, NoneOnDryRun (webhooks created via v1beta1 may also specify Some or Unknown). Webhooks with side effects MUST implement a reconciliation system, since a request may be rejected by a future step in the admission change and the side effects therefore need to be undone. Requests with the dryRun attribute will be auto-rejected if they match a webhook with sideEffects == Unknown or Some.
```

website/docs/r/validating_webhook_configuration.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+* `self_link` - A URL representing this Validating Webhook Configuration.
+* `uid` - The unique in time and space value for this Validating Webhook Configuration. For more info see [Kubernetes reference](http://kubernetes.io/docs/user-guide/identifiers#uids)
+
+### `webhook`
+
+#### Arguments
+
+* `admission_review_versions` - (Optional) AdmissionReviewVersions is an ordered list of preferred `AdmissionReview` versions the Webhook expects. API server will try to use first version in the list which it supports. If none of the versions specified in this list supported by API server, validation will fail for this object. If a persisted webhook configuration specifies allowed versions and does not include any versions known to the API Server, calls to the webhook will fail and be subject to the failure policy.
```
dak1n1, line 84:
```suggestion
* `admission_review_versions` - (Optional) AdmissionReviewVersions is an ordered list of preferred `AdmissionReview` versions the Webhook expects. API server will try to use first version in the list which it supports. If none of the versions specified in this list are supported by the API server, validation will fail for this object. If a persisted webhook configuration specifies allowed versions and does not include any versions known to the API server, calls to the webhook will fail and be subject to the failure policy.
```

website/docs/r/validating_webhook_configuration.html.markdown
```diff
@@ -0,0 +1,129 @@
...
+* `client_config` - (Required) ClientConfig defines how to communicate with the hook. 
+* `failure_policy` - (Optional) FailurePolicy defines how unrecognized errors from the admission endpoint are handled - allowed values are Ignore or Fail. Defaults to Fail.
```
dak1n1, line 86:
```suggestion
* `failure_policy` - (Optional) FailurePolicy defines how unrecognized errors from the admission endpoint are handled. Allowed values are "Ignore" or "Fail". Defaults to "Fail".
```

PR #808, opened 2020-04-02
Add volume_binding_mode to docs for storageclass

chazu:
This puppy was missing from the docs but in the provider.


aareet:
Fixed in #883

Issue #296, opened 2019-02-03
Feature Request: Container Storage Interface Support for Persistent Volumes


myyra:
[Container Storage Interface (CSI) has been promoted to GA in the Kubernetes v1.13 release](https://kubernetes.io/blog/2019/01/15/container-storage-interface-ga/). This would add support for multiple cloud providers/custom solutions with one spec.

[The API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#csipersistentvolumesource-v1-core) is relatively simple since most of the work is handled by the external driver.

ipnextgen:
As this is the new recommend way to use AWS EFS with EKS. Would like to use it via terraform.

`apiVersion: storage.k8s.io/v1beta1
kind: CSIDriver
metadata:
  name: efs.csi.aws.com
spec:
  attachRequired: false`

ref: https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/deploy/kubernetes/base/csidriver.yaml

msnelling:
Supporting CSI is the only way AFAIK that we can use terraform with Longhorn volumes.
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: longhorn-vol-pv
spec:
  capacity:
    storage: 2Gi
  volumeMode: Filesystem
  accessModes:
    - ReadWriteOnce
  persistentVolumeReclaimPolicy: Delete
  storageClassName: longhorn
  csi:
    driver: driver.longhorn.io
    fsType: ext4
    volumeAttributes:
      numberOfReplicas: '3'
      staleReplicaTimeout: '2880'
    volumeHandle: existing-longhorn-volume
```

elkbullwinkle:
Hi, is there a plan to have support for this?

DrFaust92:
i opened up 2 PRs that are relevant for this:

1. support csi for persistent volumes - #817
2. support for creating a CSI driver - #825 (having issues with deletion, creation and testing works fine)

naushadh:
Would really appreciate if the above PRs were reviewed and accepted soon. Similar to what others have already noted, this blocks us from terraforming EKS-EFS integration: https://github.com/kubernetes-sigs/aws-efs-csi-driver

DrFaust92:
Both related PRs merged, thanks @jrhouston!

aareet:
Given #817 and #825 are merged, I believe this can now be closed. Thanks for your help @DrFaust92

PR #817, opened 2020-04-13
r/persistent_volume - add support for CSI

DrFaust92:
### Acceptance tests

```
--- PASS: TestAccKubernetesPersistentVolume_csi_basic (3.65s)
```

### References

Relates #296

### DIFF
--- a/kubernetes/schema_volume_source.go
+++ b/kubernetes/schema_volume_source.go
@@ -218,6 +218,41 @@ func commonVolumeSources() map[string]*schema.Schema {
 				},
 			},
 		},
+		"csi": {
+			Type:        schema.TypeList,
+			Description: "Represents a CSI Volume. More info: http://kubernetes.io/docs/user-guide/volumes#csi",
+			Optional:    true,
+			MaxItems:    1,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"driver": {
+						Type:        schema.TypeString,
+						Description: "the name of the volume driver to use. More info: https://kubernetes.io/docs/concepts/storage/volumes/#csi",
+						Required:    true,
+					},
+					"volume_handle": {
+						Type:        schema.TypeString,
+						Description: "A map that specifies static properties of a volume. More info: https://kubernetes.io/docs/concepts/storage/volumes/#csi",
+						Required:    true,
+					},
+					"volume_attributes": {
+						Type:        schema.TypeMap,
+						Description: "Attributes of the volume to publish.",
+						Optional:    true,
+					},
+					"fs_type": {
+						Type:        schema.TypeString,
+						Description: "Filesystem type to mount. Must be a filesystem type supported by the host operating system. Ex. \"ext4\", \"xfs\", \"ntfs\". Implicitly inferred to be \"ext4\" if unspecified.",
+						Optional:    true,
+					},
+					"read_only": {
+						Type:        schema.TypeBool,
+						Description: "Whether to set the read-only property in VolumeMounts to \"true\". If omitted, the default is \"false\". More info: http://kubernetes.io/docs/user-guide/volumes#csi",
+						Optional:    true,
+					},
+				},
+			},
+		},
 		"fc": {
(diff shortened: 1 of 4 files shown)


kubernetes/schema_volume_source.go
```diff
@@ -218,6 +218,41 @@ func commonVolumeSources() map[string]*schema.Schema {
...
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"driver": {
+						Type:        schema.TypeString,
+						Description: "the name of the volume driver to use. More info: https://kubernetes.io/docs/concepts/storage/volumes/#csi",
+						Required:    true,
+					},
+					"volume_handle": {
+						Type:        schema.TypeString,
+						Description: "A map that specifies static properties of a volume. More info: https://kubernetes.io/docs/concepts/storage/volumes/#csi",
```
benosman, line 235: This description is incorrect - should refer to a string like:
`A string value that uniquely identifies the volume. `

DrFaust92: Fixed

benosman:
I'm just an end user, but I tried out this pull request and with my initial testing, so far it's working as it should for me.

DrFaust92:
Thanks for the input @benosman

kubernetes/structure_persistent_volume_spec.go
```diff
@@ -1518,6 +1559,28 @@ func patchPersistentVolumeSource(pathPrefix, prefix string, d *schema.ResourceDa
 		}
 	}
 
+	if d.HasChange(prefix + "csi") {
+		oldIn, newIn := d.GetChange(prefix + "csi")
+		oldV, oldOk := oldIn.([]interface{})
+		newV, newOk := newIn.([]interface{})
+
+		if newOk && len(newV) > 0 {
+			if oldOk && len(oldV) > 0 {
```
jrhouston, line 1568: I think we could clean the duplication here by setting 

```go
path := pathPrefix + "/csi"
value := expandCSIPersistentDiskVolumeSource(newV)
```

before the conditional, and then just supply them when constructing AddOperation/ReplaceOperation

kubernetes/schema_volume_source.go
```diff
@@ -218,6 +218,41 @@ func commonVolumeSources() map[string]*schema.Schema {
 				},
 			},
 		},
+		"csi": {
+			Type:        schema.TypeList,
+			Description: "Represents a CSI Volume. More info: http://kubernetes.io/docs/user-guide/volumes#csi",
+			Optional:    true,
+			MaxItems:    1,
+			Elem: &schema.Resource{
```
jrhouston, line 226: I see in the [API documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#csipersistentvolumesource-v1-core) for this resource there are several additional attributes for referencing secrets - is there a reason we haven't added them here?

kubernetes/structure_persistent_volume_spec.go
DrFaust92: should i just do this for all other as well? maybe extract to a more general function.
id addressed the changes you mentioned, i can open another PR for a more general refactor

kubernetes/schema_volume_source.go
DrFaust92: they are a bit tricky but i added them.
while the secret_ref state that `namespace` is optional. the api call fails if it not set. i extracted the `secret_ref` block to a common func (maybe better to extract to provider wide later)
i kept the argument optional as stated in the docs (to also make the block common). its a non breaking change as the namespace is also computed.

DrFaust92:
addressed comments + some "minor" changes.

tests output:
```
--- PASS: TestAccKubernetesPersistentVolume_csi_basic (5.32s)
--- PASS: TestAccKubernetesPersistentVolume_csi_importBasic (2.89s)
--- PASS: TestAccKubernetesPersistentVolume_csi_secrets (2.69s)
```

DrFaust92:
@jrhouston, polite nudge

PR #830, opened 2020-04-27
Add wait_for_load_balancer field to kubernetes_ingress

### LINKED ISSUE #631: Kubernetes Ingress on Azure Kubernetes Service Doesn't Wait for Public IP
### Terraform Version
Terraform v0.12.9
+ provider.azurerm v1.34.0
+ provider.helm v0.10.2
+ provider.http v1.1.1
+ provider.kubernetes v1.9.0
+ provider.null v2.1.2
+ provider.random v2.2.0
+ provider.tls v2.1.0

### Affected Resource(s)
- kubernetes_ingress

```hcl
resource "kubernetes_ingress" "vault_ingress" {
  metadata {
    name = "vault-ui"
    annotations = {
      "kubernetes.io/ingress.class" = "addon-http-application-routing"
    }
  }

  spec {
    rule {
      host = "vault-ui.${azurerm_kubernetes_cluster.aks_cluster.addon_profile[0].http_application_routing[0].http_application_routing_zone_name}"
      http {
        path {
          backend {
            service_name = "vault-ui"
            service_port = 8200
          }

          path = "/"
        }
      }
    }

    tls {
      secret_name = "vault-ui-tls"
    }
  }
}

resource "azurerm_postgresql_firewall_rule" "allow_vault" {
  name                = "vault"
  resource_group_name = "${azurerm_resource_group.vaultdemo.name}"
  server_name         = "${azurerm_postgresql_server.psql_server.name}"
  start_ip_address    = "${kubernetes_ingress.vault_ingress.load_balancer_ingress[0].ip}"
  end_ip_address      = "${kubernetes_ingress.vault_ingress.load_balancer_ingress[0].ip}"
}
```

### Debug Output
https://gist.github.com/MattMencel/0948154a1835b8fcf4dcf5045add2ee8

### Expected Behavior
I expected the Ingress public IP address to be available to use in a postgres firewall rule. 

If I run `kubectl get ingress -w` during the terraform apply, the public IP takes up to a minute to become available and appear in the ingress.

### Actual Behavior
Terraform fails immediately after deploying the Ingress, when it tries to create the firewall rule.

> Error: Invalid index
> 
>   on DB.tf line 41, in resource "azurerm_postgresql_firewall_rule" "allow_vault":
>   41:   start_ip_address    = "${kubernetes_ingress.vault_ingress.load_balancer_ingress[0].ip}"
>     |----------------
>     | kubernetes_ingress.vault_ingress.load_balancer_ingress is empty list of object
> 
> The given key does not identify an element in this collection value.

If I wait a minute or two and re-run `terraform apply` it works fine because the public IP appears on the Ingress by then.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:

With Terraform
1. Deploy an AKS resource with `http_application_routing` enabled.
2. Deploy a Service and `kubernetes_ingress`... Azure should provision the public IP automatically and attach it to the Ingress.
3. Deploy a resource or output that references the public IP from the `kubernetes_ingress` resource.

jrhouston:
### Description

Adds a `wait_for_load_balancer` field that will wait until the load balancer status is not empty for the ingress.

Resolves #631 

### Acceptance tests

### DIFF
--- a/kubernetes/resource_kubernetes_ingress.go
+++ b/kubernetes/resource_kubernetes_ingress.go
@@ -8,8 +8,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"k8s.io/api/extensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func resourceKubernetesIngress() *schema.Resource {
@@ -146,7 +145,7 @@ func resourceKubernetesIngressCreate(d *schema.ResourceData, meta interface{}) e
 
 	log.Printf("[INFO] Waiting for load balancer to become ready: %#v", out)
 	return resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
-		res, err := conn.ExtensionsV1beta1().Ingresses(metadata.Namespace).Get(metadata.Name, v1.GetOptions{})
+		res, err := conn.ExtensionsV1beta1().Ingresses(metadata.Namespace).Get(metadata.Name, metav1.GetOptions{})
 		if err != nil {
 			return resource.NonRetryableError(err)
 		}
@@ -172,7 +171,7 @@ func resourceKubernetesIngressRead(d *schema.ResourceData, meta interface{}) err
 	}
 
 	log.Printf("[INFO] Reading ingress %s", name)
-	ing, err := conn.ExtensionsV1beta1().Ingresses(namespace).Get(name, meta_v1.GetOptions{})
+	ing, err := conn.ExtensionsV1beta1().Ingresses(namespace).Get(name, metav1.GetOptions{})
 	if err != nil {
 		log.Printf("[DEBUG] Received error: %#v", err)
 		return fmt.Errorf("Failed to read Ingress '%s' because: %s", buildId(ing.ObjectMeta), err)
@@ -242,7 +241,7 @@ func resourceKubernetesIngressDelete(d *schema.ResourceData, meta interface{}) e
 	}
 
 	log.Printf("[INFO] Deleting ingress: %#v", name)
-	err = conn.ExtensionsV1beta1().Ingresses(namespace).Delete(name, &meta_v1.DeleteOptions{})
+	err = conn.ExtensionsV1beta1().Ingresses(namespace).Delete(name, &metav1.DeleteOptions{})
 	if err != nil {
 		return fmt.Errorf("Failed to delete Ingress %s because: %s", d.Id(), err)
 	}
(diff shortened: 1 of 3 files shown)


kubernetes/resource_kubernetes_ingress.go
```diff
@@ -132,7 +139,24 @@ func resourceKubernetesIngressCreate(d *schema.ResourceData, meta interface{}) e
...
 
-	return resourceKubernetesIngressRead(d, meta)
+	if !d.Get("wait_for_load_balancer").(bool) {
+		return resourceKubernetesIngressRead(d, meta)
+	}
+
+	log.Printf("[INFO] Waiting for load balancer to become ready: %#v", out)
+	return resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
+		res, err := conn.ExtensionsV1beta1().Ingresses(metadata.Namespace).Get(metadata.Name, metav1.GetOptions{})
+		if err != nil {
```
alexsomesan, line 149: I suspect there can be occasions where this `err` could also be a `NotFound` caused by the API not being fast enough to register the changes from the previous `Create` call (possibly landing on another API instance)?

Although the K8s API doesn't normally behave as eventually consistent, in clusters with horizontally scaled API instances each API process maintains a local cache which can lead to brief discrepancies on the state of newly created resources depending on where the GET call lands. This is very rare, but if the Create / Get sequence is fast enough, we might still encounter it.

jrhouston: Good point! I'll catch NotFound and make it a Retryable error.

kubernetes/resource_kubernetes_ingress_test.go
```diff
@@ -147,6 +148,27 @@ func TestAccKubernetesIngress_InternalKey(t *testing.T) {
 	})
 }
 
+func TestAccKubernetesIngress_WaitForLoadBalancerGoogleCloud(t *testing.T) {
+	var conf api.Ingress
+	name := fmt.Sprintf("tf-acc-test-%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:      func() { testAccPreCheck(t); skipIfNoGoogleCloudSettingsFound(t) },
```
jrhouston, line 156: This will only run in Google Cloud for the moment. EKS and Azure have a bunch of extra steps to get the Load balancer integration working with Ingresses, whereas it's baked into GKE.

The test will work in `kind` if you uncomment this skip,

kubernetes/resource_kubernetes_ingress_test.go
```diff
@@ -334,3 +356,70 @@ resource "kubernetes_ingress" "test" {
 	}
 }`, name)
 }
+
+func testAccKubernetesIngressConfig_waitForLoadBalancer(name string) string {
+	return fmt.Sprintf(`
+resource "kubernetes_service" "test" {
```
jrhouston, line 362: For some reason this test wouldn't pass on gcloud if I didn't have a Service and Deployment behind the Ingress. I suspect because the load balancer needs an open port before it can become ready.

kubernetes/resource_kubernetes_ingress.go
```diff
@@ -132,7 +139,29 @@ func resourceKubernetesIngressCreate(d *schema.ResourceData, meta interface{}) e
...
+			// NOTE it possible in some HA apiserver setups that are eventually consistent
```
alexsomesan, line 150:
```suggestion
			// NOTE it is possible in some HA apiserver setups that are eventually consistent
```

PR #831, opened 2020-04-28
Update getting-started.html.markdown

szymon3:
Remove duplicated `load_config_file` lines in Kubernetes provider section

### Description

I'm not sure, but for me it seems like `load_config_file` instructions are just duplicated in docs. If I'm wrong, could you please explain to me why it needs to be like this?

### DIFF
--- a/website/docs/guides/getting-started.html.markdown
+++ b/website/docs/guides/getting-started.html.markdown
@@ -73,8 +73,6 @@ If you wish to configure the provider statically you can do so by providing TLS 
 
 ```hcl
 provider "kubernetes" {
-  load_config_file = "false"
-
   host = "https://104.196.242.174"
 
   client_certificate     = file("~/.kube/client-cert.pem")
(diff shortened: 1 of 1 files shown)


alexsomesan:
Hi @szymon3.

These examples actually need `load_config_file` to be set to false. This is to enable setting explicit values for `host` and / or credentials. Otherwise, the default value of `load_config_file` being `true`, it would lead to these values being also read for the `~/.kube/config` file.

You can review the meaning of `load_config_file` here: https://www.terraform.io/docs/providers/kubernetes/index.html#load_config_file

Let us know if you need additional guidance with configuring the provider.

szymon3:
@alexsomesan That's clear. Unfortunately change preview cut few lines. 
Look, `load_config_file` is set to _false_  *twice*:
image

johnchildren:
https://github.com/terraform-providers/terraform-provider-kubernetes/pull/690
https://github.com/terraform-providers/terraform-provider-kubernetes/pull/750

Looks like it might be due to these two PRs changing the same bit of documentation in different places and then being merged independently.

PR #840, opened 2020-05-06
Add an example for GKE + NFS + PVC

dak1n1:
# Description

This example shows how to create a GKE cluster and use Google Filestore (NFS) to create Persistent Volumes in Kubernetes. It also mounts the volume to an example Deployment (mariadb) using a Persistent Volume Claim.

### References

There is a limitation in the current version of the terraform-plugin-sdk, [which makes it difficult to differentiate between `null` and `""` values for attributes](https://discuss.hashicorp.com/t/differentiate-between-null-and-empty-value/8134). For example, the `storage_class_name` attribute of the `kubernetes_persistent_volume_claim` resource cannot accept a value of `""` using the current SDK. It will interpret the value as if the attribute had been omitted entirely, which generally will cause Kubernetes to use the default storage class.

This example demonstrates a workaround, which is to create a storageclass instead of attempting to use the blank value.

There are related issues that may be mitigated by this workaround.

* https://github.com/terraform-providers/terraform-provider-kubernetes/issues/567
* https://github.com/terraform-providers/terraform-provider-kubernetes/issues/102
* https://github.com/terraform-providers/terraform-provider-kubernetes/issues/339

### DIFF
--- a/_examples/google-gke-nfs-filestore/README.md
+++ b/_examples/google-gke-nfs-filestore/README.md
@@ -1,0 +1,48 @@
+# GKE (Google Container Engine) and Google Filestore (NFS)
+
+This example demonstrates building a Google GKE cluster and Google Filestore (NFS) to create persistent volumes for use in applications in Kubernetes. There is an example application Deployment included to demonstrate mounting the NFS volume using a Persistent Volume Claim.
+
+You will need the following environment variables to be set:
+
+ - `GOOGLE_CREDENTIALS`
+ - `GOOGLE_PROJECT`
+ - `GOOGLE_REGION`
+
+For example:
+```
+[myuser@linux ~]$ env |grep GOOGLE
+GOOGLE_REGION=us-west1
+GOOGLE_CREDENTIALS=/home/myuser/.config/gcloud/legacy_credentials/mygoogleuser/adc.json
+GOOGLE_PROJECT=my-gcp-project
+```
+
+See [Google Cloud Provider docs](https://www.terraform.io/docs/providers/google/index.html#configuration-reference) for more details about these variables.
+
+Install the example using the GKE default version of Kubernetes:
+```
+terraform init
+terraform apply
+```
+
+Or optionally specify a version to use:
+```
+terraform apply -var=kubernetes_version=1.15
+```
+
+## Versions
+
+Other available versions of Kubernetes can be found by running the [gcloud](https://cloud.google.com/sdk/docs#install_the_latest_cloud_tools_version_cloudsdk_current_version) tool. However, be aware that your chosen version must be present in the `validMasterVersions`, or else the GKE `defaultClusterVersion` will be used.
+
+```
+gcloud container get-server-config --region $GOOGLE_REGION
+```
+
+## Exporting K8S variables
+To access the cluster you need to export the `KUBECONFIG` variable pointing to the `kubeconfig` file for the current cluster.
+```
+export KUBECONFIG="$(terraform output kubeconfig_path)"
+export GOOGLE_ZONE=$(terraform output google_zone)
+```
+
+Now you can access the cluster via `kubectl`.
+
--- a/_examples/google-gke-nfs-filestore/README.md
+++ b/_examples/google-gke-nfs-filestore/README.md
@@ -10,7 +10,7 @@ You will need the following environment variables to be set:
 
 For example:
 ```
-[myuser@linux ~]$ env |grep GOOGLE
+[myuser@linux ~]$ env | grep GOOGLE
 GOOGLE_REGION=us-west1
 GOOGLE_CREDENTIALS=/home/myuser/.config/gcloud/legacy_credentials/mygoogleuser/adc.json
 GOOGLE_PROJECT=my-gcp-project
@@ -45,4 +45,3 @@ export GOOGLE_ZONE=$(terraform output google_zone)
 ```
 
 Now you can access the cluster via `kubectl`.
-
--- a/_examples/google-gke-nfs-filestore/main.tf
+++ b/_examples/google-gke-nfs-filestore/main.tf
@@ -1,0 +1,242 @@
+provider "google" {
+  // Provider settings to be provided via ENV variables
+}
+
+data "google_compute_zones" "available" {
+}
+
+resource "random_id" "cluster_name" {
+  byte_length = 10
+}
+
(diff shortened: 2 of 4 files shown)


_examples/google-gke-nfs-filestore/main.tf
```diff
@@ -0,0 +1,227 @@
...
+
+resource "kubernetes_persistent_volume" "example" {
+  metadata {
+    name = "nfs-volume"
+  }
+  spec {
+    capacity = {
+      storage = "1T"
+    }
+    storage_class_name = "filestore"
```
yu-iskw, line 139: Can we refer `kubernetes_storage_class.nfs.metadata.name` instead of the literal?

_examples/google-gke-nfs-filestore/main.tf
```diff
@@ -0,0 +1,227 @@
...
+  metadata {
+    name = "filestore"
+  }
+  reclaim_policy  = "Retain"
+  storage_provisioner = "nfs"
+}
+
+resource "kubernetes_persistent_volume" "example" {
+  metadata {
+    name = "nfs-volume"
```
yu-iskw, line 133: It would be better to put `labels` for PVC to identify what PV is used.

For example,
```
metadata {
  name = "nfs-volume"
  labels = {
    name = "nfs-volume"
  }
}
```

_examples/google-gke-nfs-filestore/main.tf
```diff
@@ -0,0 +1,227 @@
...
+    }
+  }
+}
+
+resource "kubernetes_persistent_volume_claim" "example" {
+  metadata {
+    name = "mariadb-data"
+    namespace = "test"
+  }
+  spec {
```
yu-iskw, line 155: It would be better to put `spec.selector` to identify what PVC is used.
For example,
```
spec {
  selector {
    match_labels = kubernetes_persistent_volume.example.metadata.labels
  }
}
```

_examples/google-gke-nfs-filestore/main.tf
```diff
@@ -0,0 +1,227 @@
...
+    access_modes = ["ReadWriteMany"]
+    storage_class_name = "filestore"
```
yu-iskw, line 157: ditto

_examples/google-gke-nfs-filestore/main.tf
```diff
@@ -0,0 +1,227 @@
...
+    volume_name = "nfs-volume"
```
yu-iskw, line 158: Can we refer `kubernetes_persistent_volume .example.metadata.name`?

_examples/google-gke-nfs-filestore/main.tf
```diff
@@ -0,0 +1,227 @@
...
+    resources {
+      requests = {
+        storage = "1T"
+      }
+    }
+  }
```
yu-iskw, line 164: It would be nicer to add `depends_on`.

For example,
```
  depends_on = [kubernetes_persistent_volume.example]
```

yu-iskw:
I got the error when executing `terraform plan` locally.

```
$ terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

data.google_compute_zones.available: Refreshing state...

Error: project: required field is not set

  on main.tf line 5, in data "google_compute_zones" "available":
   5: data "google_compute_zones" "available" {
```

yu-iskw:
I got the error when executing `terraform plan` locally.

```
Error: node_version and min_master_version must be set to equivalent values on create

  on main.tf line 41, in resource "google_container_cluster" "primary":
  41: resource "google_container_cluster" "primary" {



Error: Error waiting to create Instance: Error waiting for Creating Instance: error while retrieving operation: Get https://file.googleapis.com/v1/projects/kouzoh-p-yu/locations/us-west1-a/operations/operation-1588828160178-5a507e218b8a4-60a7d6b5-2de8328e?alt=json: dial tcp [2404:6800:4004:81b::200a]:443: connect: no route to host

  on main.tf line 71, in resource "google_filestore_instance" "test":
  71: resource "google_filestore_instance" "test" {
```

_examples/google-gke-nfs-filestore/main.tf
dak1n1: Good idea! Done.

_examples/google-gke-nfs-filestore/main.tf
dak1n1: Since this resource now has these references defined, there is an implicit dependency. So the PVC won't be created until the storageclass and the PV are created.

```
 storage_class_name = kubernetes_storage_class.nfs.metadata[0].name
 volume_name = kubernetes_persistent_volume.example.metadata[0].name
```

I looked it up in the [docs](https://www.terraform.io/docs/configuration/resources.html#depends_on-explicit-resource-dependencies) to be sure:

"Explicitly specifying a dependency is only necessary when a resource relies on some other resource's behavior but doesn't access any of that resource's data in its arguments."

dak1n1:
I'm glad you brought that up! I ran into that error too, with the node_version. It means there there is a mismatch in version availability between the master and node. The availability of versions varies by zone, and the version that you pass in via the CLI might not be available in your chosen zone for both the master and node.

I've made a change to the config that will install the latest version available for the master instead (since it looks like older versions are available for masters). And if no version is found, it'll use the GKE default. It worked in my testing when I passed in `1.15` and `1.16` for the kubernetes version.

_examples/google-gke-nfs-filestore/main.tf
yu-iskw: Sure. Let's keep the current implementation.

_examples/google-gke-nfs-filestore/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+You will need the following environment variables to be set:
+
+ - `GOOGLE_CREDENTIALS`
+ - `GOOGLE_PROJECT`
+ - `GOOGLE_REGION`
+
+For example:
+```
+[myuser@linux ~]$ env |grep GOOGLE
```
jrhouston, line 13:
```suggestion
[myuser@linux ~]$ env | grep GOOGLE
```

jrhouston:
This example is great @dak1n1 thanks for making this.

Issue #605, opened 2019-09-01
Wait for StatefulSet rollout status


yaman:
Hi,

Like **Deployment Resource**, **StatefulSet** rollout status should be successful before kubernetes provider decides that it is successfully deployed. Is there any obstacles no to use same kind of strategy used in Deployment Resource(**waitForDeploymentReplicasFunc**)? If not, I would gladly to contribute with a similar implementation.

### Terraform&Providers' Version
- Terraform v0.12.7
- terraform-providers/kubernetes: 1.9.0

### Affected Resource(s)
Please list the resources as a list, for example:
- **StatefulSet**

### Expected Behavior
Provider should wait before **StatefulSet** is deployed&running state.

### Actual Behavior
Provider successfully terminates saying that **StatefulSet** is created successfully.

### Steps to Reproduce
Please list the steps required to reproduce the issue;
1. `terraform apply` with a StatefulSet resource.

### Important Factoids
Running at GKE

LiuVII:
Alternatively and more generally it'd be nice to depend on the success of some script/command, like `pre-start` hook

dsymonds:
Related, I'd like a way to *avoid* waiting for rollouts. I have some Deployments and StatefulSets that can take 30m to fully update, but I don't want `terraform apply` to wait that long. In other words, I'd like either a per-resource property or some provider-wide option to skip the whole `waitForDeploymentReplicasFunc` step, and exit with the status of the Patch.

hashibot:
This issue has been open 180 days with no activity. If this issue is reproducible with the latest version of the provider and with Terraform 0.12, please comment. Otherwise this issue will be closed in 30 days.

dsymonds:
Yes, it is still relevant.

dsymonds:
I've proposed an approach here: https://github.com/terraform-providers/terraform-provider-kubernetes/pull/842

It doesn't do the StatefulSet waiting, but it does offer the control over this.

PR #842, opened 2020-05-07
Add wait_for_rollout attribute to kubernetes_deployment.

dsymonds:
This controls whether the provider will wait for a create or update
operation to be completely applied on the server side. For some
situations this is desirable behaviour, but in other situations where a
gradual rollout using the builtin Kubernetes controls can take hours
this is undesirable.

It is set to true by default (matching current behaviour).

This should be done for kubernetes_stateful_set too, but it wasn't clear
how to wait for a StatefulSet change to complete.

Updates #605.

### Acceptance tests
- [ ] Have you run the acceptance tests on this branch? (If so, please include the test log in a gist)

### DIFF
--- a/kubernetes/resource_kubernetes_deployment.go
+++ b/kubernetes/resource_kubernetes_deployment.go
@@ -31,13 +31,11 @@ func resourceKubernetesDeployment() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
-
 		Timeouts: &schema.ResourceTimeout{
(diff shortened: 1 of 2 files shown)


dsymonds:
Let me know if this is acceptable in principle, and I can update docs. Not sure how to add an acceptance test for something like this, but am happy to run the acceptance tests.

I tried to add this for kubernetes_stateful_set (the initial driver of #605), but couldn't figure out from the k8s API how to wait in the same way that this code waits on kubernetes_deployment.

dsymonds:
Taken over by #916.

Issue #693, opened 2019-11-15
Support enableServiceLinks in PodSpec


JPMoresmau:
### Terraform Version
Terraform v0.12.7
+ provider.kubernetes v1.10.0

### Affected Resource(s)
- kubernetes_deployment
- kubernetes_pod

### Expected Behavior
Have a way to use enable_service_links in a deployment/pod spec

### Actual Behavior
`An argument named "enable_service_links" is not expected here.`

### References
https://github.com/kubernetes/kubernetes/pull/68754

Since environment variables can wreak havoc with some programs (most notably Apache Java apps), turning them off was requested.
Note that I can see enableServiceLinks in https://github.com/terraform-providers/terraform-provider-kubernetes/blob/master/vendor/k8s.io/api/core/v1/types_swagger_doc_generated.go

abrzozowski-ITM:
StatefulSet is also affected.

MathyV:
@aareet any chance of getting that PR reviewed?

shd4:
Any updates on this? enableServiceLinks not being disabled produces a lot of pain...

aareet:
The PR is waiting on an acceptance test being added.

dak1n1:
It looks like we implemented this a while back but didn't update this issue. The `enable_service_links` option is present in the PodSpec now: https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/pod#enable_service_links

PR #853, opened 2020-05-25
Make enableServiceLinks configurable from the pod specification

MathyV:
### Description

I stumbled into resource problems on our cluster because enableServiceLinks is enabled by default, so I wanted to disable it but Terraform is not able to do so. I found issue #693 which talks about this lack of functionality. I created an implementation that seems to work by copying and adapting the host_ipc parameter.

One thing that I'm not sure I handled correctly is the default value. Might be a lack of comprehension from my side as this is the first time I take a dive into the terraform source. 

It seems if the value was never sent to k8s it's not in the pod spec and in.EnableServiceLinks == nil. In that case for k8s it means that the value is set to true. I thought that setting the Default in the schema would be enough to make sure that att["enable_service_links"] would be set to true elsewhere in the code but this did not work.

The only way I got it working was by adding the else clause in structures_pod.go. The code now works for me in all cases, I tested changing both pre-change pods and post-change pods, with true, false and no value specified.

### Acceptance tests

I did not create additional tests as host_ipc doesn't have them either.

### References

https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#accessing-the-service

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#pod-v1-core

Issue #693


DrFaust92:
@MathyV, can you rebase and resolve conflicts for this PR? really would like to have this feature

MathyV:
> @MathyV, can you rebase and resolve conflicts for this PR? really would like to have this feature

@DrFaust92 done

DrFaust92:
Thanks, looks good. Can you also add an acceptance test for this?

MathyV:
> Thanks, looks good. Can you also add an acceptance test for this?

I will have a look, not sure how to do it though, like I said I just copied stuff from host_ipc and that one didn't have tests so I figured this one might not need one too.

DrFaust92:
@MathyV, are you still looking into this? i can fork from your branch and add the needed tests

MathyV:
I didn't have time yet to look into it, won't have time until next week,
but feel free to do so if you want it faster.

DrFaust92:
Opened #975 that uses original branch + added tests

MathyV:
Closing this in favour of the new PR.

Issue #864, opened 2020-06-02
kubernetes_persistent_volume_claim resource overrides empty storage_class_name with "default" value


ragelo:
There is the difference in PVCs created using `terraform` and `kubectl`. `kubernetes_persistent_volume_claim` resource overrides defined `storage_class_name = ""` with "default" value. `kubernetes_persistent_volume` doesn't have such behavior and works correct.

### Terraform Version and Provider Version

```
> terraform -v
Terraform v0.12.24
```
Provider version: v1.11.2
### Affected Resource(s)

- kubernetes_persistent_volume_claim

### Terraform Configuration Files
```hcl
resource "kubernetes_persistent_volume_claim" "blobfuse" {
  metadata {
    name = "pvc-empty-storage-class"
  }

  spec {
    storage_class_name = ""
    access_modes = ["ReadWriteOnce"]

    resources {
      requests = {
        storage = "1Gi"
      }
    }
  }
}
```

#### Kubernetes equivalent that provides expected output
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-empty-storage-class-1
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  storageClassName: ""

```

### Debug Output

`terraform` PVC 
```
kubectl describe persistentvolumeclaim/pvc-empty-storage-class
Name:          pvc-empty-storage-class
Namespace:     default
StorageClass:  default
Status:        Bound
Volume:        pvc-9025ee73-59a3-48cf-9ebd-d0f935e6b2c1
Labels:        <none>
Annotations:   pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
               volume.beta.kubernetes.io/storage-provisioner: kubernetes.io/azure-disk
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      1Gi
Access Modes:  RWO
VolumeMode:    Filesystem
Mounted By:    <none>
Events:
  Type    Reason                 Age   From                         Message
  ----    ------                 ----  ----                         -------
  Normal  ProvisioningSucceeded  8s    persistentvolume-controller  Successfully provisioned volume pvc-9025ee73-59a3-48cf-9ebd-d0f935e6b2c1 using kubernetes.io/azure-disk
```

`kubectl` PVC
```
kubectl describe persistentvolumeclaim/pvc-empty-storage-class-1
[...]

PR #865, opened 2020-06-02
Handle empty PersistentVolumeClaimSpec storage_class_name

ragelo:
### Description

Pass empty string to `PersistentVolumeClaimSpec.StorageClassName ` instead of leaving attribute untouched (`nil`).

### Acceptance tests
- [X] Have you run the acceptance tests on this branch? (If so, please include the test log in a gist)
https://gist.github.com/ragelo/b988b0d4cbb8b22bf53d3ab8b6fdc76f
### References
#864


dak1n1:
Thanks for this PR! I wish I could have saved you the trouble... those tests look great! Unfortunately, we've already tested out this solution and it won't work with the current version of the SDK. Closing in favor of this issue: https://github.com/terraform-providers/terraform-provider-kubernetes/issues/872

PR #900, opened 2020-07-02
resourceKubernetesSecretCreate updated to support from_env_file

albertorm95:
### Acceptance tests
- [❌ ] Have you added an acceptance test for the functionality being added?
- [❌ ] Have you run the acceptance tests on this branch? (If so, please include the test log in a gist)

### References
https://github.com/hashicorp/terraform-provider-kubernetes/issues/889

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_secret.go
+++ b/kubernetes/resource_kubernetes_secret.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	api "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


kubernetes/resource_kubernetes_secret.go
```diff
@@ -48,9 +55,22 @@ func resourceKubernetesSecretCreate(d *schema.ResourceData, meta interface{}) er
...
+			secretData[lineSplitted[0]] = lineSplitted[1]
+		}
+	} else {
+		secretData = d.Get("data").(map[string]interface{})
+	}
+
 	secret := api.Secret{
 		ObjectMeta: metadata,
-		Data:       expandStringMapToByteMap(d.Get("data").(map[string]interface{})),
+		Data:       expandStringMapToByteMap(secretData),
```
jrhouston, line 73: The problem you're running into here is that you are setting this from the value from the `from_env_file` attribute, but then when the provider comes to read the resource after it is created [here](https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/kubernetes/resource_kubernetes_secret.go#L95) it reads this field into the `data` attribute, so you are always going to get a diff because `data` is empty at that point.

There's a couple of approaches you could take here:

1. Use the value of `from_env_file` to set the value for the `data` attribute at create time.
2. At read time, check which of `data` or `from_env_file` were set and populate accordingly. I can foresee some problems with this approach, because you then have to re-encode the secret data as the env_file format.

albertorm95: Ok! Thanks, I'll try that :)

PR #922, opened 2020-07-20
Add resource for CertificateSigningRequest

dak1n1:
### Description

This PR adds the `kubernetes_certificate_signing_request` resource to allow for automatic certificate generation within Kubernetes.

Example usage (I'll add more via examples/docs):

```
resource "kubernetes_certificate_signing_request" "example" {
  metadata {
    name = "example"
  }
  spec {
    request = tls_cert_request.example.cert_request_pem
    usages = ["client auth", "server auth"]
  }
  auto_approve = true # this can be set to false in case the user would prefer a custom controller approve the request
}
```

The resulting certificate is stored in Terraform state, making this resource a "logical resource" rather than persisting any objects in the Kubernetes API. The reason for this is that CSRs within Kubernetes are temporary objects that are [deleted by a garbage-collection controller](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#request-signing-process) usually within one hour. Since the most valuable part of the CertificateSigningRequest is the certificate itself, that is the only API-generated data that is kept persistently in this resource's state.

### Acceptance tests

### References

https://github.com/hashicorp/terraform-provider-kubernetes/issues/447#issuecomment-656199305

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_certificate_signing_request.go
+++ b/kubernetes/resource_kubernetes_certificate_signing_request.go
@@ -1,0 +1,200 @@
+package kubernetes
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"k8s.io/api/certificates/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+	"log"
+	"reflect"
+	"time"
+)
+
+func resourceKubernetesCertificateSigningRequest() *schema.Resource {
+	apiDoc := v1beta1.CertificateSigningRequest{}.SwaggerDoc()
+	apiDocSpec := v1beta1.CertificateSigningRequestSpec{}.SwaggerDoc()
+	apiDocStatus := v1beta1.CertificateSigningRequestStatus{}.SwaggerDoc()
+
+	return &schema.Resource{
+		Create: resourceKubernetesCertificateSigningRequestCreate,
+		Read:   resourceKubernetesCertificateSigningRequestRead,
+		Delete: resourceKubernetesCertificateSigningRequestDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(5 * time.Minute),
+		},
+		Schema: map[string]*schema.Schema{
+			"auto_approve": {
+				Type:        schema.TypeBool,
+				Description: "Automatically approve the CertificateSigningRequest",
+				Optional:    true,
+				ForceNew:    true,
+				Default:     true,
+			},
+			"certificate": {
+				Type:        schema.TypeString,
+				Description: apiDocStatus["certificate"],
+				Computed:    true,
+			},
+			"metadata": metadataSchemaForceNew(metadataSchema("certificate signing request", true)),
+			"spec": {
+				ForceNew:    true,
+				Type:        schema.TypeList,
+				Description: apiDoc[""],
+				Required:    true,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"request": {
+							Type:        schema.TypeString,
+							Description: apiDocSpec["request"],
+							Required:    true,
+							ForceNew:    true,
+						},
+						"signer_name": {
+							Type: schema.TypeString,
+							// no swagger doc available for signerName
+							Description: "Requested signer for the request. It is a qualified name in the form: `scope-hostname.io/name`." +
+								"If empty, it will be defaulted: 1. If it's a kubelet client certificate, it is assigned `kubernetes.io/kube-apiserver-client-kubelet`." +
+								"2. If it's a kubelet serving certificate, it is assigned `kubernetes.io/kubelet-serving`." +
+								"3. Otherwise, it is assigned `kubernetes.io/legacy-unknown`. Distribution of trust for signers happens out of band." +
+								"You can select on this field using `spec.signerName`.",
+							Optional: true,
+							ForceNew: true,
+						},
+						"usages": {
+							Type:        schema.TypeSet,
+							Description: apiDocSpec["usages"],
+							Set:         schema.HashString,
+							Elem:        &schema.Schema{Type: schema.TypeString},
+							Optional:    true,
+							ForceNew:    true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func resourceKubernetesCertificateSigningRequestCreate(d *schema.ResourceData, meta interface{}) error {
+	conn, err := meta.(KubeClientsets).MainClientset()
+	if err != nil {
+		return err
+	}
+
+	metadata := expandMetadata(d.Get("metadata").([]interface{}))
+	spec, err := expandCertificateSigningRequestSpec(d.Get("spec").([]interface{}))
+	if err != nil {
+		return err
+	}
+
+	csr := v1beta1.CertificateSigningRequest{
+		ObjectMeta: metadata,
+		Spec:       *spec,
+	}
+	log.Printf("[INFO] Creating new certificate signing request: %#v", csr)
+	newCSR, createErr := conn.CertificatesV1beta1().CertificateSigningRequests().Create(&csr)
+	if createErr != nil {
+		return fmt.Errorf("Failed to create certificate signing request: %s", err)
+	}
+
+	// Get the name, since it might have been randomly generated during create.
+	csrName := newCSR.ObjectMeta.Name
+
+	// Delete the remote CSR resource when this function exits, or when errors are encountered.
+	defer conn.CertificatesV1beta1().CertificateSigningRequests().Delete(csrName, &metav1.DeleteOptions{})
+
+	if d.Get("auto_approve").(bool) {
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			pendingCSR, getErr := conn.CertificatesV1beta1().CertificateSigningRequests().Get(csrName, metav1.GetOptions{})
+			if getErr != nil {
+				return getErr
+			}
+			approval := v1beta1.CertificateSigningRequestCondition{
+				Type:    "Approved",
+				Reason:  "TerraformAutoApprove",
+				Message: "This CSR was approved by Terraform auto_approve.",
+			}
+			pendingCSR.Status.Certificate = []byte{}
+			pendingCSR.Status.Conditions = append(pendingCSR.Status.Conditions, approval)
+			_, updateErr := conn.CertificatesV1beta1().CertificateSigningRequests().UpdateApproval(pendingCSR)
+			return updateErr
+		})
+		if retryErr != nil {
+			panic(fmt.Errorf("CSR auto-approve update failed: %v", retryErr))
+		}
+		fmt.Println("CSR auto-approve update succeeded")
+	}
+
+	log.Printf("[DEBUG] Waiting for certificate to be issued")
+	stateConf := &resource.StateChangeConf{
+		Target:  []string{"Issued"},
+		Pending: []string{"", "Approved"},
+		Timeout: d.Timeout(schema.TimeoutCreate),
+		Refresh: func() (interface{}, string, error) {
+			out, refreshErr := conn.CertificatesV1beta1().CertificateSigningRequests().Get(csrName, metav1.GetOptions{})
+			if refreshErr != nil {
+				log.Printf("[ERROR] Received error: %v", refreshErr)
+				return out, "Error", refreshErr
+			}
+			var csrStatus string
+			emptyStatus := v1beta1.CertificateSigningRequestStatus{}
+			emptyCSR := v1beta1.CertificateSigningRequest{}
+
+			// If the CSR is empty, check again later.
+			if reflect.DeepEqual(emptyCSR, out) {
+				return out, csrStatus, nil
+			}
+
+			// If the status is empty, check again later.
+			if reflect.DeepEqual(emptyStatus, out.Status) {
+				return out, csrStatus, nil
+			}
+
+			// Check to see if a certificate has been issued, and update status accordingly,
+			// since 'Issued' is not a state ever populated in the Status Conditions.
+			for _, condition := range out.Status.Conditions {
+				log.Printf("[DEBUG] Found Status.Condition.Type: %v", condition.Type)
+				if string(condition.Type) == "Approved" {
+					if string(out.Status.Certificate) != "" {
+						log.Printf("[DEBUG] Found non-empty Certificate field in Status")
+						csrStatus = "Issued"
+					}
+				}
+			}
+			log.Printf("[DEBUG] CertificateSigningRequest %s status received: %#v", csrName, csrStatus)
+			return out, csrStatus, nil
+		},
+	}
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return err
+	}
+	log.Printf("[INFO] Certificate issued for request: %s", csrName)
+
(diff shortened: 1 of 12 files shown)


kubernetes/structures.go
```diff
@@ -190,10 +190,6 @@ func ptrToString(s string) *string {
 	return &s
 }
 
-func ptrToInt(i int) *int {
```
dak1n1, line 193: My IDE was complaining that this function is unused, so I removed it.

kubernetes/structures.go
```diff
@@ -218,7 +214,7 @@ func base64EncodeStringMap(m map[string]interface{}) map[string]interface{} {
 	result := make(map[string]interface{})
 	for k, v := range m {
 		value := v.(string)
-		result[k] = (base64.StdEncoding.EncodeToString([]byte(value)))
+		result[k] = base64.StdEncoding.EncodeToString([]byte(value))
```
dak1n1, line 217: Another minor linting fix.

kubernetes/validators.go
```diff
@@ -181,25 +181,6 @@ func validatePositiveInteger(value interface{}, key string) (ws []string, es []e
 	return
 }
 
-func validateDNSPolicy(value interface{}, key string) (ws []string, es []error) {
```
dak1n1, line 184: These are more functions that are unused in the code base, so I cleaned them up.

kubernetes/resource_kubernetes_certificate_signing_request.go
```diff
@@ -0,0 +1,196 @@
...
+				out, refreshErr := conn.CertificatesV1beta1().CertificateSigningRequests().Get(metadata.Name, metav1.GetOptions{})
+				if refreshErr != nil {
+					log.Printf("[ERROR] Received error: %#v", refreshErr)
+					return out, "Error", err
+				}
+				statusConditions := ""
+				emptyStatus := v1beta1.CertificateSigningRequestStatus{}
+
+				if !reflect.DeepEqual(emptyStatus, out.Status) {
+					statusConditions = fmt.Sprintf("%v", out.Status.Conditions[0].Type)
```
dak1n1, line 149: I want to see if I can search for the condition in this array instead. WIP

kubernetes/resource_kubernetes_certificate_signing_request.go
```diff
@@ -0,0 +1,196 @@
...
+// resourceKubernetesCertificateSigningRequestRead does not return any data, because Read functions exist to
+// sync the local state with the remote state. Since this data is local-only, there is nothing to read.
+func resourceKubernetesCertificateSigningRequestRead(d *schema.ResourceData, meta interface{}) error {
+	_ = d
+	_ = meta
+	return nil
+}
+
+func resourceKubernetesCertificateSigningRequestDelete(d *schema.ResourceData, meta interface{}) error {
+	_ = meta
```
dak1n1, line 192: I put this `_ = meta` here because otherwise the parameter `meta` is unused. But the SDK requires the parameter even if you're not using it.

dak1n1:
The code is tested and working. TC test running now. I'll start working on docs for the website.

kubernetes/resource_kubernetes_certificate_signing_request.go
jrhouston: You should be able to do a linear search on the array using the `type` field.

kubernetes/resource_kubernetes_certificate_signing_request.go
```diff
@@ -0,0 +1,208 @@
...
+			Create: schema.DefaultTimeout(5 * time.Minute),
+		},
+		Schema: map[string]*schema.Schema{
+			"auto_approve": {
+				Type: schema.TypeBool,
+				Description: "By choosing this option, Terraform will automatically approve the CertificateSigningRequest" +
+					" and a certificate will be issued. Note that this will save the certificate in Terraform's state file, " +
+					"which makes the certificate readable in plaintext on the system where Terraform was run. " +
+					"It is recommended to use a remote backend when selecting this option, to prevent local users " +
+					"from reading the certificate data. https://www.terraform.io/docs/state/remote.html",
```
dak1n1, line 37: @jrhouston I'm going to remove the warning about saving the certificate in state. Does that sound ok? Kubernetes only generates a public cert, which is not sensitive data (it is freely given to remote servers by design). I originally had this warning here because at a previous employer, they liked to keep the certificates hidden.

dak1n1:
TC tests passed. Here is the latest acceptance test run locally.

```
[dakini@dax terraform-provider-kubernetes (csr $)]$ for acctest in TestAccKubernetesCertificateSigningRequest_basic TestAccKubernetesCertificateSigningRequest_generateName; do cd kubernetes ; TF_ACC=true go test -v -run $acctest ; cd - ;
done
=== RUN   TestAccKubernetesCertificateSigningRequest_basic
CSR auto-approve update succeeded
--- PASS: TestAccKubernetesCertificateSigningRequest_basic (1.07s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   1.144s
/home/dakini/go/src/github.com/hashicorp/terraform-provider-kubernetes
=== RUN   TestAccKubernetesCertificateSigningRequest_generateName
CSR auto-approve update succeeded
--- PASS: TestAccKubernetesCertificateSigningRequest_generateName (1.03s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   1.100s
/home/dakini/go/src/github.com/hashicorp/terraform-provider-kubernetes
```

kubernetes/resource_kubernetes_certificate_signing_request.go
dak1n1: Oops, I just saw this comment. Let me know if you want me to solve this a different way. The code snippet there is outdated.

kubernetes/resource_kubernetes_certificate_signing_request.go
jrhouston: Couldn't you make `_` the name of the parameter in the function signature, instead of doing this assignment?

kubernetes/schema_metadata.go
```diff
@@ -79,6 +79,11 @@ func metadataSchema(objectName string, generatableName bool) *schema.Schema {
 	}
 }
 
+func metadataSchemaForceNew(s *schema.Schema) *schema.Schema {
+	s.ForceNew = true
```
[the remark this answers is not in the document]
jrhouston: Something else you could do here is use the [CustomizeDiff](https://www.terraform.io/docs/extend/resources/customizing-differences.html) feature and use `ForceNewIfChange("metdata", ... `

I'm not sure which is better tbh.

kubernetes/resource_kubernetes_certificate_signing_request.go
```diff
@@ -0,0 +1,204 @@
...
+			if refreshErr != nil {
+				log.Printf("[ERROR] Received error: %v", refreshErr)
+				return out, "Error", refreshErr
+			}
+			var csrStatus string
+			emptyStatus := v1beta1.CertificateSigningRequestStatus{}
+			emptyCSR := v1beta1.CertificateSigningRequest{}
+
+			// Ensure the CSR fields have been populated before checking them.
+			if !reflect.DeepEqual(emptyCSR, out) {
```
jrhouston, line 149: I'd kind of like to invert this logic so the nesting doesn't get so deep, i.e:

```
// Ensure the CSR fields have been populated before checking them.
if reflect.DeepEqual(emptyCSR, out) {
    return out, "", nil 
}

if reflect.DeepEqual(emptyStatus, out.Status) {
    return out, "", nil
}

for _, condition := range out.Status.Conditions {
    log.Printf("[DEBUG] Found Status.Condition.Type: %v", condition.Type)
    if string(condition.Type) == "Approved" {
        // Check to see if a certificate has been issued, and update status accordingly,
        // since 'Issued' is not a state ever populated in the Status Conditions.
        if string(out.Status.Certificate) != "" {
            log.Printf("[DEBUG] Found non-empty Certificate field in Status")
            csrStatus = "Issued"
        }
    }
}
```

kubernetes/schema_metadata.go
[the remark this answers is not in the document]
alexsomesan: I don't think that the underscoring thing is necessary at all for function arguments.
If I'm not mistaken, Go doesn't account the function arguments as "unused variables" when they aren't actually user.

Did a quick test here: https://play.golang.com/p/hRJXvvbx_GJ

kubernetes/resource_kubernetes_certificate_signing_request.go
alexsomesan: I don't think that the underscoring thing is necessary at all for function arguments.
If I'm not mistaken, Go doesn't account the function arguments as "unused variables" when they aren't actually user.

Did a quick test here: https://play.golang.com/p/G6NaZdO6x0O

kubernetes/resource_kubernetes_certificate_signing_request.go
```diff
@@ -0,0 +1,204 @@
...
+						}
+					}
+				}
+			}
+			log.Printf("[DEBUG] CertificateSigningRequest %s status received: %#v", csrName, csrStatus)
+			return out, csrStatus, nil
+		},
+	}
+	_, err = stateConf.WaitForState()
+	if err != nil {
```
alexsomesan, line 169: I hope I understand this workflow correctly. If I do, I think I see an issue here.

So if `auto_approve` is set to `false` then a human operator (or a manually triggered tool) will have to sign the request on the cluster cluster side. In that case, the timeout in the StateChangeConf is likely to expire and WaitForState will return an error. This will in turn fail the Create action and Terraform will fail the apply.

At this point, it seems to me there isn't any mechanism available to actually retrieve the (supposedly) eventually signed certificate. As I see it, the next possible action is only to run apply again, which will cause a new certificate request to be created (since the entire resource is marked ForceNew) and the entire timeout cycle to potentially repeat itself.

I wonder if actually implementing Read and Import can help here. Having a Read function would allow the Terraform operator to run a `refresh` operation and eventually retrieve the approved cert and reconcile the state to something stable.

Also, analyzing the chain of responsibility from the point of view of Terraform (and this provider), the only thing asked through the CSR resource is to issue a signing request (it's in the name of it too). Waiting for the actual signing to happen should be a best effort IMHO and not on the critical path with a failure mode attached.

I hope I didn't misunderstand how this works, but if I did please do set me on the straight path.

kubernetes/resource_kubernetes_certificate_signing_request.go
dak1n1: Yeah, the code runs without `_`, but IDEs will flag it as an unused variable, depending on the IDE's linter. Mine underlines unused variables. This is what I see: https://i.imgur.com/BcxGmQd.png

But I've renamed the unused variables to `_` like John suggested and I think that looks better.

dak1n1: I looked up some information about why Go allows code to compile when there is an unused parameter. The purpose is to allow better readability in the function when you have optional parameters. I guess it's also required if you have more than one optional parameter too.

But since this isn't a parameter that will ever be used in these functions, I think it makes sense to leave it as `_`. (Reference https://stackoverflow.com/a/22549439/9231732 )

kubernetes/schema_metadata.go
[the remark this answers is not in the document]
dak1n1: Thanks for the suggestion. Since I'm omitting the `Update` function for this resource, all attributes need to have `ForceNew` set. Otherwise the SDK throws an error. I'm not sure that a `CustomizeDiff` function would satisfy that requirement. Plus it seems to be more of a standard practice to specify the `ForceNew` at the attribute level when possible. `CustomizeDiff` is most useful for cases where you need more complex logic to determine when an attribute should use `ForceNew`.

kubernetes/resource_kubernetes_certificate_signing_request.go
dak1n1: Good idea, that's simpler to follow.

kubernetes/resource_kubernetes_certificate_signing_request.go
dak1n1: Thanks for pointing this out! It actually showed me that I was making an assumption which turns out not to be true. So I'll fix this and test the timeout to make sure it works as expected. Here are some responses to specific parts of your comment to explain how it works:

> 
> So if `auto_approve` is set to `false` then a human operator (or a manually triggered tool) will have to sign the request on the cluster cluster side. In that case, the timeout in the StateChangeConf is likely to expire and WaitForState will return an error. This will in turn fail the Create action and Terraform will fail the apply.

The `auto_approve = false` setting is one that I tested by running `kubectl certificate approve` while the `terraform apply` was still running. This simulates the way a Kubernetes custom controller would quickly approve the request without manual intervention. But if needed, someone could also manually approve the CSR like I did. The most common use case though is a Kubernetes operator/controller.

> 
> At this point, it seems to me there isn't any mechanism available to actually retrieve the (supposedly) eventually signed certificate. 

This is why the `terraform apply` will wait until someone approves the request. If they don't approve it, then the apply will fail, and the remote resource will be deleted (or so I thought! but this is not true... yet).

I had never let it run for the full 5 minutes to see what would happen if it failed. I assumed if it failed then nothing would be saved in state, and so it would just try the create again... But when it does that, we get an error because the resource already exists in Kubernetes. This is the part I'll fix. What I need to do is defer the deletion of the API resource so that *even if it times out* the resource will be deleted, and the re-apply will be successful.

dak1n1: It works now. Thanks!

```
kubernetes_certificate_signing_request.example: Still creating... [4m40s elapsed]
kubernetes_certificate_signing_request.example: Still creating... [4m50s elapsed]
kubernetes_certificate_signing_request.example: Still creating... [5m0s elapsed]

Error: timeout while waiting for state to become 'Issued' (timeout: 5m0s)

  on main.tf line 16, in resource "kubernetes_certificate_signing_request" "example":
  16: resource "kubernetes_certificate_signing_request" "example" {


[dakini@dax 20200806]$
[dakini@dax 20200806]$
[dakini@dax 20200806]$ tfapply
tls_private_key.example: Refreshing state... [id=413b4f1762fa9d7431cad39b15fccb7f131b4e72]
tls_cert_request.example: Refreshing state... [id=eeabc2dd57703ea3ba1002a511e60bf108f6f0ee]
kubernetes_certificate_signing_request.example: Creating...
kubernetes_certificate_signing_request.example: Still creating... [10s elapsed]
kubernetes_certificate_signing_request.example: Still creating... [20s elapsed]
kubernetes_certificate_signing_request.example: Creation complete after 23s [id=example]
kubernetes_secret.example: Creating...
kubernetes_secret.example: Creation complete after 0s [id=default/test-secret]
kubernetes_pod.main: Creating...
kubernetes_pod.main: Creation complete after 6s [id=default/test-pod]

Apply complete! Resources: 3 added, 0 changed, 0 destroyed.
```

dak1n1:
I think I addressed all comments so this is ready for a final review. I'm running TC tests on it one more time.

dak1n1:
TC tests look good.

dak1n1:
`auto_approve` was optional but didn't actually work when left unspecified. I fixed that by adding a default behavior for it, and ran the acceptance test locally.

dak1n1:
Travis found an issue with the website:

```
/home/travis/gopath/src/github.com/hashicorp/terraform-website/content/scripts/check-links.sh "http://127.0.0.1:4567/docs/providers/kubernetes/"
Crawling site...
http://127.0.0.1:4567/docs/providers/kubernetes/:
Remote file does not exist -- broken link!!!
Found 1 broken link.
http://127.0.0.1:4567/docs/providers/kubernetes/
make[1]: *** [website-provider-test] Error 8
make[1]: Leaving directory `/home/travis/gopath/src/github.com/hashicorp/terraform-website'
make: *** [website-test] Error 2
The command "make website-test" exited with 2.
```

So I'll work on that next.

dak1n1:
It turns out this test has been broken for 4+ days. I'm going to merge since everything looks ok in this PR.

PR #923, opened 2020-07-20
Feature - Allow empty string for storage class name

josephmilla:
### Description

Current implementation doesn't allow empty strings for storage_class_names when it should've been allowed

### Acceptance tests

```
➜  kubernetes git:(feature/allow-empty-string-for-storage-class-name) go test
2020/07/20 10:56:33 [DEBUG] Trying to load configuration from file
2020/07/20 10:56:33 [DEBUG] Configuration file is: test-fixtures/kube-config.yaml
2020/07/20 10:56:33 [DEBUG] Using custom current context: "gcp"
2020/07/20 10:56:33 [DEBUG] Using overidden context: api.Context{LocationOfOrigin:"", Cluster:"", AuthInfo:"", Namespace:"", Extensions:map[string]runtime.Object(nil)}
2020/07/20 10:56:33 [INFO] Successfully initialized config
PASS
ok  	github.com/hashicorp/terraform-provider-kubernetes/kubernetes	1.546s
```

### References
- https://github.com/hashicorp/terraform-provider-kubernetes/issues/872

### Community Note

### DIFF
--- a/kubernetes/structure_persistent_volume_claim.go
+++ b/kubernetes/structure_persistent_volume_claim.go
@@ -79,7 +79,7 @@ func expandPersistentVolumeClaimSpec(l []interface{}) (*v1.PersistentVolumeClaim
 	if v, ok := in["volume_name"].(string); ok {
 		obj.VolumeName = v
 	}
-	if v, ok := in["storage_class_name"].(string); ok && v != "" {
+	if v, ok := in["storage_class_name"].(string); ok {
 		obj.StorageClassName = ptrToString(v)
 	}
 	return obj, nil


dak1n1:
Thanks for this PR. Unfortunately, this is an approach we have looked at in the past (such as in https://github.com/hashicorp/terraform-provider-kubernetes/pull/865) and we found that while this will add support for the use case of the *empty* storageclass name, it removes support for *omitting* the storageclass name (that is, using the [defaultStorageClass](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#dynamic) instead). Since it's a pretty common workflow to create a PVC without specifying a storageclass, we don't want to break that existing functionality. Instead, we recommend creating a new storageclass to use with your PVC. There is an example of creating a storageclass for NFS [here](https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/_examples/google-gke-nfs-filestore/main.tf#L130-L172).

PR #949, opened 2020-08-07
r/deployment - add support for readiness_gate

### LINKED ISSUE #811: support for pod readinessGates when using kubernetes_deployment
### Community Note

* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request

### Description

We use terraform resource "kubernetes_deployment" to deploy our pods. Our pods have readiness probes, but these probes are not good enough because we need external feedback to decide if a pod is ready. In our case, a pod is ready only after external program creates a file in aws S3 bucket, and this is a manual step, which may be completed at random time (can be several days/weeks) so readiness probe is not good, becuase it will fail and leave our pod in "unready" state. We understand kubernetes 1.14 introduced something called readiness-gate. See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate However it seems terraform resource "kubernetes_deployment" does not support pod readiness-gates.

Note that we prefer using kubernetes deployment (rather than defining pods directly) because we need rolling update strategy.

It would be great if resource "kubernetes_deployment" supported readinessGates for pods

DrFaust92:
### Acceptance tests

```
--- PASS: TestAccKubernetesPod_readinessGate (38.91s)
```

### References
Closes #811

### Community Note


DrFaust92:
Any ideas how i can achieve adding conditions to the pod in a simple manner? i cant even find a way to hack something like this.

dak1n1:
I think you would make an API call to edit the pod object on the server. Similar to what I did here, to update the status.Conditions field of the CSR object: https://github.com/dak1n1/terraform-provider-kubernetes/blob/2946fdd0b08bb4187bf8aeee94979d9ebd420974/kubernetes/resource_kubernetes_certificate_signing_request.go#L114-L125

Looking in my IDE, I see this function exists for updating a Pod's status: `conn.CoreV1().Pods(metadata.Namespace).UpdateStatus()`

That should let you set the status to any value. It would probably have to be a function inside of the test file that sets a test value.  But if it ends up being too much trouble, I can probably get something added, or talk to @jrhouston to see if we can skip adding this test.

In this example, I use one TestStep to create the resource (for example, create the Pod). Then the second TestStep makes a change to that resource, which can be tested using a custom function.

https://github.com/dak1n1/terraform-provider-kubernetes/blob/c245d421be2e2070e8cb61a03a6cc4f026d5c07c/kubernetes/resource_kubernetes_persistent_volume_claim_test.go#L459-L482

That's a start, anyway. But yeah, just let us know if it doesn't work out.

DrFaust92:
Thanks so much for the detailed explanation @dak1n1, i'll give this a go.

DrFaust92:
@dak1n1, this seems to have done the trick!

```
--- PASS: TestAccKubernetesPod_readinessGate (38.91s)

```

kubernetes/structures_pod.go
```diff
@@ -1025,6 +1049,22 @@ func expandVolumes(volumes []interface{}) ([]v1.Volume, error) {
 	return vl, nil
 }
 
+func expandReadinessGates(gates []interface{}) ([]v1.PodReadinessGate, error) {
+	if len(gates) == 0 {
```
alexsomesan, line 1053: Here you also need to check for `gates[0] = nil` (see the other TypeList expanders).
I know it's not intuitive but if you set it to empty block in configuration, you get a nil value and it will crash the provider.
```
readiness_gate {}
```

DrFaust92: Done

DrFaust92:
Rebased + Addressed comments 

`--- PASS: TestAccKubernetesPod_readinessGate (27.72s)`

alexsomesan (approved):
Tests pass locally.

PR #950, opened 2020-08-08
r/namespace - add support for finalizers

### LINKED ISSUE #855: Support spec with finalizers in namespace resource
### Community Note

* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request

### Description

In #701 , namespace data sources were added that can handle finalizers. The resource doesn't have support for it yet as far as I can tell from the code and the documentation.

### Potential Terraform Configuration

```hcl
resource "kubernetes_namespace" "test" {
  metadata {
    name = "mytest"
  }
  spec {
    finalizers = ['myfinalizer0','myfinalizer1']
  }
}
```

### References

- #701 
- https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers
- https://github.com/kubernetes/kubernetes/issues/77086

DrFaust92:
### Acceptance tests
- [X] Have you run the acceptance tests on this branch? (If so, please include the test log in a gist)

```
--- PASS: TestAccKubernetesNamespace_importSpec (9.56s)
--- PASS: TestAccKubernetesNamespace_spec (12.72s)
--- PASS: TestAccKubernetesNamespace_basic (11.09s)
```

### References
Closes #855 

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_namespace.go
+++ b/kubernetes/resource_kubernetes_namespace.go
@@ -37,7 +37,7 @@ func resourceKubernetesNamespace() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"finalizers": {
-							Type:        schema.TypeSet,
+							Type:        schema.TypeList,
 							Description: "Finalizers is an opaque list of values that must be empty to permanently remove object from storage.",
 							Optional:    true,
 							Elem: &schema.Schema{
@@ -113,18 +113,26 @@ func resourceKubernetesNamespaceUpdate(d *schema.ResourceData, meta interface{})
 	ctx := context.TODO()
 
 	ops := patchMetadata("metadata.0.", "/metadata/", d)
+
+	if d.HasChange("spec.0.finalizers") {
+		r := d.Get("spec.0.finalizers")
+
+		finalizers := make([]v1.FinalizerName, 0)
+		for _, f := range r.([]interface{}) {
+			finalizer := v1.FinalizerName(f.(string))
+			finalizers = append(finalizers, finalizer)
+		}
+
+		log.Printf("haha %#v", r)
+		ops = append(ops, &ReplaceOperation{
+			Path:  "/spec/finalizers",
+			Value: finalizers,
+		})
+	}
+
 	data, err := ops.MarshalJSON()
 	if err != nil {
 		return fmt.Errorf("Failed to marshal update operations: %s", err)
-	}
-
-	if d.HasChange("spec") {
-		spec := expandNamespaceSpec(d.Get("spec").([]interface{}))
-
-		ops = append(ops, &ReplaceOperation{
-			Path:  "/spec",
-			Value: spec,
-		})
 	}
 
 	log.Printf("[INFO] Updating namespace: %s", ops)
--- a/kubernetes/resource_kubernetes_namespace.go
+++ b/kubernetes/resource_kubernetes_namespace.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	api "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	pkgApi "k8s.io/apimachinery/pkg/types"
@@ -27,6 +28,25 @@ func resourceKubernetesNamespace() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"metadata": metadataSchema("namespace", true),
+			"spec": {
+				Type:        schema.TypeList,
+				Description: "Spec defines the behavior of the Namespace.",
+				Computed:    true,
+				Optional:    true,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"finalizers": {
+							Type:        schema.TypeList,
+							Description: "Finalizers is an opaque list of values that must be empty to permanently remove object from storage.",
+							Optional:    true,
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+					},
+				},
+			},
 		},
 		Timeouts: &schema.ResourceTimeout{
 			Delete: schema.DefaultTimeout(5 * time.Minute),
@@ -44,6 +64,7 @@ func resourceKubernetesNamespaceCreate(d *schema.ResourceData, meta interface{})
 	metadata := expandMetadata(d.Get("metadata").([]interface{}))
 	namespace := api.Namespace{
 		ObjectMeta: metadata,
+		Spec:       expandNamespaceSpec(d.Get("spec").([]interface{})),
 	}
 	log.Printf("[INFO] Creating new namespace: %#v", namespace)
 	out, err := conn.CoreV1().Namespaces().Create(ctx, &namespace, metav1.CreateOptions{})
@@ -76,6 +97,11 @@ func resourceKubernetesNamespaceRead(d *schema.ResourceData, meta interface{}) e
 		return err
 	}
 
+	err = d.Set("spec", flattenNamespaceSpec(&namespace.Spec))
+	if err != nil {
+		return fmt.Errorf("error setting spec: %w", err)
+	}
+
 	return nil
 }
 
@@ -90,6 +116,15 @@ func resourceKubernetesNamespaceUpdate(d *schema.ResourceData, meta interface{})
 	data, err := ops.MarshalJSON()
 	if err != nil {
 		return fmt.Errorf("Failed to marshal update operations: %s", err)
+	}
+
+	if d.HasChange("spec") {
+		spec := expandNamespaceSpec(d.Get("spec").([]interface{}))
+
+		ops = append(ops, &ReplaceOperation{
+			Path:  "/spec",
+			Value: spec,
+		})
 	}
 
 	log.Printf("[INFO] Updating namespace: %s", ops)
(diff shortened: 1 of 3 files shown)


kubernetes/resource_kubernetes_namespace.go
```diff
@@ -88,6 +114,15 @@ func resourceKubernetesNamespaceUpdate(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("Failed to marshal update operations: %s", err)
 	}
 
+	if d.HasChange("spec") {
```
alexsomesan, line 117: This way of constructing the patch operation by reusing the expand.. helper won't work. It seems that patch operations need to be constructed at the primitive attribute level, for each attribute in turn.

Please have a look at an existing implementation like `func patchPodSpec(...)` for inspiration on how to implement here.

dak1n1: I had to do something like this recently too. You can check out my example if you want. https://github.com/dak1n1/terraform-provider-kubernetes/blob/c245d421be2e2070e8cb61a03a6cc4f026d5c07c/kubernetes/resource_kubernetes_persistent_volume_claim.go#L197

You'll also want to move the `ops.MarshalJSON()` a few lines down to make sure you don't marshall it until all the operations have been appended. Otherwise the Update is performed without your changes.

kubernetes/resource_kubernetes_namespace_test.go
```diff
@@ -352,6 +406,20 @@ resource "kubernetes_namespace" "test" {
 `, nsName)
 }
 
+func testAccKubernetesNamespaceSpecConfig(nsName string) string {
+	return fmt.Sprintf(`
+resource "kubernetes_namespace" "test" {
+  metadata {
+    name = %[1]q
```
dak1n1, line 413: Usually we use `%s` for the name. Is there a reason to use this other notation? I'm not familiar with it offhand.

```suggestion
    name = "%s"
```

jrhouston: We use a mixture of `%q` and `%s` in the test fixtures, so we aren't consistent about it. The value of using the `%q` verb is just to save a few characters, particularly when you have to escape the quotes around it. The square brackets denote which argument to use. In this case it is superfluous as there is only one argument.

Side note: I like to use [gore](https://github.com/motemen/gore) to figure out language features quickly when I don't want to read through documentation:

```go
gore> :import fmt
fmt.Sprintf("%s %q %[1]q", "first", "second", "third")
"first \"second\" \"first\""
```

kubernetes/resource_kubernetes_namespace.go
```diff
@@ -26,6 +27,25 @@ func resourceKubernetesNamespace() *schema.Resource {
...
+			"spec": {
+				Type:        schema.TypeList,
+				Description: "Spec defines the behavior of the Namespace.",
+				Computed:    true,
+				Optional:    true,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"finalizers": {
+							Type:        schema.TypeList,
```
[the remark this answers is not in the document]
DrFaust92: actually i went with how its implemented in the namespace data source, should i switch this in both?

kubernetes/resource_kubernetes_namespace.go
DrFaust92: ive changed it to update according to the comments above, thanks for that. the body of the request seem to be fine now but the update doesnt fail but doesnt really work...

any advice?

kubernetes/resource_kubernetes_namespace.go
```diff
@@ -87,6 +113,23 @@ func resourceKubernetesNamespaceUpdate(d *schema.ResourceData, meta interface{})
...
+			finalizer := v1.FinalizerName(f.(string))
+			finalizers = append(finalizers, finalizer)
+		}
+
+		log.Printf("haha %#v", r)
+		ops = append(ops, &ReplaceOperation{
+			Path:  "/spec/finalizers",
+			Value: finalizers,
+		})
+	}
```
dak1n1, line 131: Hmm, this looks correct offhand. But I see what you mean, it's not actually working. Here are logs from applying this with `TF_LOG=trace`.

https://gist.githubusercontent.com/dak1n1/915bc7e682a14e5c8a838fd8b3dc40a6/raw/db9d6c16048ee0a040801892f9fca6a34451e71f/gistfile1.txt

You can see the request was submitted successfully, but the object remains unchanged (just one finalizer listed instead of two).

This seems to be enforced by the Kubernetes API itself, as you can see by this manual test:

```
[dakini@dax 20201005]$ kube edit namespace test -v7
I1005 12:10:58.003446   38864 loader.go:375] Config loaded from file:  /home/dakini/.kube/config
I1005 12:10:58.004348   38864 cert_rotation.go:137] Starting client certificate rotation controller
I1005 12:10:58.007521   38864 round_trippers.go:420] GET https://192.168.39.49:8443/api/v1/namespaces/test
I1005 12:10:58.007532   38864 round_trippers.go:427] Request Headers:
I1005 12:10:58.007537   38864 round_trippers.go:431]     Accept: application/json
I1005 12:10:58.007540   38864 round_trippers.go:431]     User-Agent: kubectl/v1.19.0 (linux/amd64) kubernetes/e199641
I1005 12:10:58.016199   38864 round_trippers.go:446] Response Status: 200 OK in 8 milliseconds
I1005 12:10:58.016799   38864 editor.go:128] Opening file with editor [vi /tmp/kubectl-edit-as522.yaml]
I1005 12:11:15.031662   38864 editoptions.go:298] User edited:
# Please edit the object below. Lines beginning with a '#' will be ignored,
# and an empty file will abort the edit. If an error occurs while saving this file will be
# reopened with the relevant failures.
#
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: "2020-10-05T19:04:17Z"
  name: test
  resourceVersion: "1097"
  selfLink: /api/v1/namespaces/test
  uid: 6aa1777e-f0ef-464c-886a-81aa2af0a622
spec:
  finalizers:
  - kubernetes
  - foregroundDeletion
status:
  phase: Active
I1005 12:11:15.032998   38864 round_trippers.go:420] GET https://192.168.39.49:8443/openapi/v2?timeout=32s
I1005 12:11:15.033043   38864 round_trippers.go:427] Request Headers:
I1005 12:11:15.033072   38864 round_trippers.go:431]     Accept: application/com.github.proto-openapi.spec.v2@v1.0+protobuf
I1005 12:11:15.033097   38864 round_trippers.go:431]     User-Agent: kubectl/v1.19.0 (linux/amd64) kubernetes/e199641
I1005 12:11:15.062323   38864 round_trippers.go:446] Response Status: 200 OK in 29 milliseconds
I1005 12:11:15.187899   38864 round_trippers.go:420] PATCH https://192.168.39.49:8443/api/v1/namespaces/test?fieldManager=kubectl-edit
I1005 12:11:15.187912   38864 round_trippers.go:427] Request Headers:
I1005 12:11:15.187916   38864 round_trippers.go:431]     Content-Type: application/strategic-merge-patch+json
I1005 12:11:15.187920   38864 round_trippers.go:431]     User-Agent: kubectl/v1.19.0 (linux/amd64) kubernetes/e199641
I1005 12:11:15.187923   38864 round_trippers.go:431]     Accept: application/json
I1005 12:11:15.191600   38864 round_trippers.go:446] Response Status: 200 OK in 3 milliseconds
namespace/test edited
```
```
[dakini@dax 20201005]$ kube get namespace test -o json |jq
{
  "apiVersion": "v1",
  "kind": "Namespace",
  "metadata": {
    "creationTimestamp": "2020-10-05T19:04:17Z",
    "managedFields": [
      {
        "apiVersion": "v1",
        "fieldsType": "FieldsV1",
        "fieldsV1": {
          "f:status": {
            "f:phase": {}
          }
        },
        "manager": "HashiCorp",
        "operation": "Update",
        "time": "2020-10-05T19:04:43Z"
      },
      {
        "apiVersion": "v1",
        "fieldsType": "FieldsV1",
        "fieldsV1": {
          "f:spec": {
            "f:finalizers": {}
          }
        },
        "manager": "kubectl-edit",
        "operation": "Update",
        "time": "2020-10-05T19:11:15Z"
      }
    ],
    "name": "test",
    "resourceVersion": "1371",
    "selfLink": "/api/v1/namespaces/test",
    "uid": "6aa1777e-f0ef-464c-886a-81aa2af0a622"
  },
  "spec": {
    "finalizers": [
      "kubernetes"
    ]
  },
  "status": {
    "phase": "Active"
  }
}
```

Based on this, I think this PR won't work after all. It probably makes more sense to pursue a metadata-level finalizer instead. https://github.com/hashicorp/terraform-provider-kubernetes/issues/855#issuecomment-676732693

dak1n1 (commented):
I found a problem that seems to make this PR impossible to actually implement. The Kubernetes API doesn't seem to support editing the namespace spec. Sorry that we didn't catch this sooner!

dak1n1:
It might actually be possible to delete and re-create the namespace... I'm not sure if it would be worth the trouble though. There is an explanation here about using a different API endpoint to "finalize" the namespace so that it will delete properly.

https://github.com/kubernetes/kubernetes/issues/77086#issuecomment-486840718

DrFaust92:
@dak1n1 No sweat. I was starting to go crazy why this doesn't update.

I'll take a look at what you sent later this week and try to make sense of it.

DrFaust92:
I don't see a way around updating, I can add force new for finalisers and call it a day. WDYT @dak1n1?

alexsomesan:
@DrFaust92 I suspect adding `ForceNew` is not best option here. It will translate into a sequence of Delete and Create operations. The Delete is problematic though, since we're dealing with a namespace here which might have dependent resources in the cluster. You could try falling back to an `Update` type of operation (in the sense of client-go Update). However, I'm not sure that would work - I haven't tested it. However, I think a complete solution here will require more complex logic than we have now.

Before doing that, I think it's worth taking a step back and considering if this change is actually worth doing at this point in time because `spec.finalizers` already have a successor in `metadata.finalizers`. @dak1n1 rightly pointed out the conversation in the Kubernetes repo that sheds some light on these. I haven't used finalizers in practice so far so I can't offer authoritative advice on the difference between the two. Please have a look at `metadata.finlalizers` and let's discuss the opportunity to implement those and how they suit your use-case.

DrFaust92:
Cool,

Regarding using the Update func, tried that   and it doesn't work.

dak1n1:
Agreed @DrFaust92, the only type of `Update` we can do on this field involves sending a command directly to the namespace's `/finalize` API endpoint, which will allow *removal* of the finalizer only, which then would allow for a `Delete` (ForceNew). All that would have to be done inside a CustomizeDiff function so that a ForceNew could be possible on that attribute. And at that point, the user might have unwittingly wiped out all the resources inside that namespace (they could potentially lose application data).

It's a lot of work for a feature that I think shouldn't actually be used in practice. If someone comes to us with a compelling use case, I would consider it, but offhand I think users would get more value from the metadata level finalizer.

The metadata finalizer is easily updatable and more broadly usable (once it's implemented, it should work on every resource automatically, since every resource pulls in a shared metadata schema).

Since Alex and I are in agreement about using the metadata finalizer instead, I'm going to close this PR. We'll track the issue here https://github.com/hashicorp/terraform-provider-kubernetes/issues/1040

PR #952, opened 2020-08-11
Use service account annotations for secret import

tustvold:
### Description

The initial implementation of service account import used a 1 second timestamp delta - see [here](https://github.com/hashicorp/terraform-provider-kubernetes/pull/377). This was raised to 3 seconds by #748, however, I have encountered a service account where the delta is 4 seconds preventing import with the current logic.

This PR therefore changes the logic to use the annotations put in place by the tokens_controller [see here](https://github.com/kubernetes/kubernetes/blob/release-1.13/pkg/controller/serviceaccount/tokens_controller.go#L384). This should not only be more accurate, but also less fragile.

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_service_account.go
+++ b/kubernetes/resource_kubernetes_service_account.go
@@ -328,8 +328,8 @@ func resourceKubernetesServiceAccountImportState(d *schema.ResourceData, meta in
 func findDefaultServiceAccount(sa *api.ServiceAccount, conn *kubernetes.Clientset) (string, error) {
 	/*
 		The default service account token secret would have:
-		- been created either at the same moment as the service account or _just_ after (Kubernetes controllers appears to work off a queue)
 		- have a name starting with "[service account name]-token-"
+		- annotations linking back to the service account name and uid
 
 		See this for where the default token is created in Kubernetes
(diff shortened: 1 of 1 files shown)


dak1n1:
Thanks for this contribution. It is a good solution to the problem! However, there was a new feature added last week that I think has solved this same problem (if I understand your use case correctly). The new resource is called `kubernetes_default_service_account`. It "adopts" the default service account into Terraform, similar to importing, but with the added benefit of allowing users to specify changes to the default service account in their Terraform config before the SA exists yet.

There will be a release later this week that will contain the feature. https://github.com/hashicorp/terraform-provider-kubernetes/pull/876

Let us know if this covers your use case, thanks!

aareet:
I'm going to close this since it has been open for 2 months without response since #876 was released.

tustvold:
Hi, sorry I missed the comment on this PR. This issue still stands for importing service accounts created outside of terraform into terraform. Whilst the default_service_account resource works for the default service account, this isn't the only use case nor the use case for which this PR was actually created

aareet:
No problem, thanks for responding @tustvold , I'll reopen so the team can reconsider.

aareet:
Test note: Need to make sure this method is reliable in all supported environments

PR #959, opened 2020-08-20
Fix url attribute in admissionregistration client_config.service block

### LINKED ISSUE #882: mutating_webhook_configuration forces service block in client config when url is used
Terraform version: v0.12.26
Provider: v1.11.3

This is easily reproducable, just use the example in the docs as foundation:

```
resource "kubernetes_mutating_webhook_configuration" "example" {
  metadata {
    name = "test.terraform.io"
  }

  webhook {
    name = "test.terraform.io"

    admission_review_versions = ["v1", "v1beta1"]

    client_config {
      url = "my-url"  
      #service {
      #   namespace = "example-namespace"
      #    name = "example-service"
      # }
    }

    rule {
      api_groups = ["apps"]
      api_versions = ["v1"]
      operations = ["CREATE"]
      resources = ["deployments"]
      scope = "Namespaced"
    }

    reinvocation_policy = "IfNeeded"
    side_effects = "None"
  }
}
```

As can be seen above i have commented out the service block and only want to use the url which should be valid configuration according to the docs:

> url - (Optional) Gives the location of the webhook, in standard URL form (scheme://host:port/path). Exactly one of url or service must be specified

The current output i'm seeing when run with only the url specified:

> "webhook.0.client_config.0.service": required field is not set

Funny enough if i set both the url and service terraform passes but of course if will fail when deploying to Kubernetes since this is an invalid state. I've also tried a couple of workarounds that almost got me there, for instance configuring the service block with _name_ and _namespace_ set to empty strings which seems to make the fields go away but there will be set a default port anyway so this won't work either. That is probably also a bug since i end up with a service block with only a port defined as:

```
service {
  port = 443
}
```

If there are any known workarounds that would be appreciated.

jrhouston:
### Description

Fixes #882 

The problem here was that we weren't checking list length before expanding the service block. I have update the acceptance test to capture this case. I also noticed that the API does some validation on the URL, so I have added that to the schema too.

### Acceptance tests

### Community Note

### DIFF
--- a/kubernetes/schema_webhook_client_config.go
+++ b/kubernetes/schema_webhook_client_config.go
@@ -1,6 +1,9 @@
 package kubernetes
 
 import (
+	"errors"
+	"net/url"
+
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 )
@@ -53,6 +56,30 @@ func webhookClientConfigFields() map[string]*schema.Schema {
 			Type:        schema.TypeString,
 			Description: apiDoc["url"],
 			Optional:    true,
+			ValidateFunc: func(v interface{}, k string) ([]string, []error) {
+				url, err := url.Parse(v.(string))
+				if err != nil {
+					return nil, []error{err}
+				}
+
+				if url.Scheme != "https" {
+					return nil, []error{errors.New("url: scheme must be https")}
+				}
+
+				if url.Host == "" {
+					return nil, []error{errors.New("url: host must be provided")}
+				}
+
+				if url.User != nil {
+					return nil, []error{errors.New("url: user info is not permitted")}
+				}
+
+				if url.Fragment != "" || url.RawQuery != "" {
+					return nil, []error{errors.New("url: fragments and query parameters are not permitted")}
+				}
+
+				return nil, nil
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


kubernetes/schema_webhook_client_config.go
```diff
@@ -53,6 +56,30 @@ func webhookClientConfigFields() map[string]*schema.Schema {
 			Type:        schema.TypeString,
 			Description: apiDoc["url"],
 			Optional:    true,
+			ValidateFunc: func(v interface{}, k string) ([]string, []error) {
+				url, err := url.Parse(v.(string))
```
dak1n1, line 60: The variable `url` collides with the imported package name here. While it doesn't cause any runtime or compile errors, it is advisable to change either the import name or the variable name to avoid issues if the code is updated in the future (for example, if someone tries to use a method from the `net/url` library further on in this file).

kubernetes/schema_webhook_client_config.go
```diff
@@ -53,6 +56,32 @@ func webhookClientConfigFields() map[string]*schema.Schema {
 			Type:        schema.TypeString,
 			Description: apiDoc["url"],
 			Optional:    true,
+			ValidateFunc: func(v interface{}, k string) ([]string, []error) {
+				u, err := url.Parse(v.(string))
+				if err != nil {
+					return nil, []error{err}
+				}
+
+				errs := []error{}
```
dak1n1, line 65: Good idea! This will create a more helpful error message. My IDE is being nit-picky and saying that nil slices should be created like this instead `var errs []error`, but I don't mind it either way.

jrhouston: I know that in this case the validator checks the length rather than checking if it's `nil` so it is a matter of taste here. 

Interestingly, a nil slice will behave the same as an empty slice in every way except that will also return true on comparison to `nil`. See this Stackoverflow post: https://stackoverflow.com/questions/44305170/nil-slices-vs-non-nil-slices-vs-empty-slices-in-go-language

PR #967, opened 2020-08-25
Fix crash when deferencing nil pointer in v1beta1.IngressRule

alexsomesan:
### Description

Fixes https://github.com/hashicorp/terraform-provider-kubernetes/issues/966

In `v1beta1.IngressRule` the `HTTP` attribute is a pointer type and this implies it potentially can be `nil`.
Before this change, the `flattenIngressRule` function was assuming this will never be the case.
Issue #966 demonstrates that `HTTP` can actually be null (see attached debug log). 

This change address the potential for `nil` values.

### Acceptance tests

### Community Note

### DIFF
--- a/kubernetes/structure_ingress_spec.go
+++ b/kubernetes/structure_ingress_spec.go
@@ -10,29 +10,34 @@ import (
 
 func flattenIngressRule(in []v1beta1.IngressRule) []interface{} {
 	att := make([]interface{}, len(in), len(in))
-	for i, n := range in {
-		// rulePrefix := fmt.Sprintf("rule.%d.")
+	for i, r := range in {
 		m := make(map[string]interface{})
 
-		m["host"] = n.Host
-
-		httpAtt := make(map[string]interface{})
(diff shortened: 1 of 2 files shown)


kubernetes/structure_ingress_spec.go
```diff
@@ -10,29 +10,34 @@ import (
...
-			path := map[string]interface{}{
-				"path":    p.Path,
-				"backend": flattenIngressBackend(&p.Backend),
-			}
-			pathAtts[i] = path
-		}
-		httpAtt["path"] = pathAtts
-		m["http"] = []interface{}{
-			httpAtt,
+func flattenIngressRuleHttp(in *v1beta1.HTTPIngressRuleValue) []interface{} {
```
jrhouston, line 23: Thanks for breaking this out 👍

jrhouston:
Looks good, but we should add a regression test for the config that actually surfaced this bug as well as the unit test for the flattener code that was buggy.

jrhouston (approved):

PR #837, opened 2020-05-05
[kubernetes_persistent_volume] fix azure pv disk kind

pchalamet:
### Description

This is an up-to-date fix for a long standing issue: inability to create a persistent volume from a managed disk on Azure.

### Acceptance tests
- [X] Have you added an acceptance test for the functionality being added?
- [X] Have you run the acceptance tests on this branch? (If so, please include the test log in a gist)

### References

Original issue is here: https://github.com/terraform-providers/terraform-provider-kubernetes/issues/202

People attempted to provide fix but never reach the approved state:
* https://github.com/terraform-providers/terraform-provider-kubernetes/pull/342
* https://github.com/terraform-providers/terraform-provider-kubernetes/pull/586

So it's basically a merge from master from PR https://github.com/terraform-providers/terraform-provider-kubernetes/pull/586 with cleanup.
I've mostly done nothing - prior work from aforementioned PRs.

PR includes:
* Support for kubernetes_persistent_volume.spec.persistent_volume_source.azure_disk.kind
* Units tests
* Documentation of new feature
* No change to current behavior if kind is not provided

I've tested this on my production, this works perfectly and all volumes are now created this way.

Let me know if some more work is necessary. I will try to do my best to make this into master.

PR #968, opened 2020-08-27
Update Go dependencies

dak1n1:
### Description

This PR updates Go dependencies as a prerequisite to adding the Azure provider into the Kubernetes provider (importing for acceptance tests). This also updates our other Go modules and specifically Kubernetes related libraries.

### Acceptance tests

### References

https://github.com/hashicorp/terraform-provider-kubernetes/pull/837

### Community Note


dak1n1:
This is still WIP. I just made it a regular PR so I can run TC tests on it.

kubernetes/provider_test.go
```diff
@@ -346,16 +363,6 @@ func clusterVersionLessThan(vs string) bool {
 	return cv.LessThan(v)
 }
 
-func skipCheckIf(skip func() (bool, string), check resource.TestCheckFunc) resource.TestCheckFunc {
```
dak1n1, line 349: This function was unused, so I removed it.

dak1n1:
So far, acceptance tests are passing. This is ready for review.

https://gist.github.com/dak1n1/a6addda90b1f741c5b3b41a42ec3fc81

dak1n1:
This is ready for review. I'm hoping to get it merged soon so I can continue working on the Azure disk PR (#837).

aareet:
👋 I'll take a look

kubernetes/provider.go
```diff
@@ -225,6 +228,13 @@ func (k kubeClientsets) MainClientset() (*kubernetes.Clientset, error) {
 	return k.mainClientset, nil
 }
 
+func (k kubeClientsets) GetContext() (*context.Context, error) {
```
jrhouston, line 231: I'm questioning the value of having this method, and storing a shared context here alongside the clients and passing down a pointer to it. I know that at this point it is academic as the context doesn't actually do anything yet.

It more makes sense that each CRUD operation would happen in its own context, or receive its context from the calling function in the case where a Create immediately then does a Read. This is actually what's going to happen when we migrate to SDK v2 - each CRUD func is going to have a context aware version that will supply a `context.Context` as the first argument.

See here: https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#more-support-for-context-context

What I would do here is add a `ctx := context.TODO()` to the top of every CRUD func, then we would simply remove that line when we do the SDK migration as `ctx`will become the first argument of the function.

cc @alexsomesan for his $0.02 on this.

Sorry, I know you've added this in like 200 places 😱

kubernetes/provider_test.go
```diff
@@ -214,6 +215,18 @@ func skipIfNoAwsSettingsFound(t *testing.T) {
 	}
 }
 
+func skipIfNoAzureCloudSettingsFound(t *testing.T) {
+	if os.Getenv("TF_VAR_aks_client_id") == "" && os.Getenv("ARM_CLIENT_ID") == "" {
```
jrhouston, line 219: The other skip functions in the tests don't use the `TF_VAR_` convention - why have we decided to use it for azure?

dak1n1: It was included in the Azure PR, and I remember testing it and thinking it was necessary. But I'll double check to be sure.

kubernetes/provider.go
dak1n1: Yeah, I was feeling uncertain about that function too. I know we don't want to share a context too widely. Here is some information on it, for reference:

https://github.com/golang/go/issues/22602
```
Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx: [...]"

"While we've told people not to add contexts to structs, I think that guidance is over-aggressive. The real advice is not to store contexts. They should be passed along like parameters. But if the struct is essentially just a parameter, it's okay. I think this concern can be addressed with package-level documentation and examples."
```

I thought that by storing it in that struct, it would allow us to share a context when needed, and we would have the option to declare new contexts where those are required too. So +1 to not using it everywhere. I'm just not sure where to draw that line. Some functions should probably share a context, while others should not. Maybe adding one per CRUD operation is the way to go, like you said.

kubernetes/provider_test.go
alexsomesan: My 2 cents here is to move away entirely from checking environment type based on credentials being present. It's not as precise as checking node labels, which we already do for some cluster types. IMO we should move towards using node labels across the board.

The reason why I'm advising against it is because they are easily producing false positives in runtime environments where more than one set of credentials is present (like cloning a TC config for an existing cloud provider and forgetting to remove the old credentials or more commonly local runtime environments where we might have all creds available).

dak1n1: That sounds good to me. I'll remove it from this PR and do the work in the Azure PR. I already put a node check in that branch anyway. Thanks!

kubernetes/provider.go
alexsomesan: I think John has a good point here. Taking a step back to think about the role of the context in our case, it's to coordinate the cancelation of upstream connections in case the "parent" operation (apply, destroy) is stopped prematurely for whatever reason. So we should always be passing around the context corresponding to that specific operation. Passing any other (older) context isn't adding value for the effort required to use them.

That being said, I second what John suggested about just using `ctx := context.TODO()` in every CRUD until we gain access to the new SDKs context aware interface.

dak1n1:
Ok, I think this is finally done. All the makefile targets pass locally, except I didn't run `make testacc` because that takes a couple hours. Instead, I queued up these acceptance tests to cover a variety of cloud providers and kubernetes versions:

Kubernetes (AKS @ 1.17)
https://ci-oss.hashicorp.engineering/viewLog.html?buildId=147049

Kubernetes (GKE @ 1.16) (Terraform @ 0.13)
https://ci-oss.hashicorp.engineering/viewLog.html?buildId=147047

Kubernetes (GKE @ 1.15)
https://ci-oss.hashicorp.engineering/viewLog.html?buildId=147048

Kubernetes (Typhoon AWS @ 1.18)
https://ci-oss.hashicorp.engineering/viewLog.html?buildId=147046

I'll come back in a couple hours to check out the results.

aareet (approved):
Thanks for working through this @dak1n1! Did a best-effort review of the changes, LGTM.

PR #998, opened 2020-09-09
Fix reported replica count when waiting for Deployment rollout

alexsomesan:
### Description

Provider reports incorrect replica count in error message after waiting for rollout times out.
```
Error: Waiting for rollout to finish: 0 out of 824644526080 new replicas have been updated...
```
After this change, the correct replica count will be reported:
```
Error: Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
```

### Acceptance tests
This changes only concerns user-facing output. It cannot be captured in acceptance tests.

### Release Note

```release-note
* Fix replica count reported in error message after timeout while waiting for rollout
```

### DIFF
--- a/kubernetes/resource_kubernetes_deployment.go
+++ b/kubernetes/resource_kubernetes_deployment.go
@@ -408,6 +408,11 @@ func waitForDeploymentReplicasFunc(conn *kubernetes.Clientset, ns, name string) 
 			return resource.NonRetryableError(err)
 		}
 
+		var specReplicas int32 = 1 // default, acording to API docs
+		if dply.Spec.Replicas != nil {
+			specReplicas = *dply.Spec.Replicas
+		}
+
 		if dply.Generation <= dply.Status.ObservedGeneration {
 			cond := GetDeploymentCondition(dply.Status, appsv1.DeploymentProgressing)
 			if cond != nil && cond.Reason == TimedOutReason {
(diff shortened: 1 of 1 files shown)


kubernetes/resource_kubernetes_deployment.go
```diff
@@ -408,15 +408,20 @@ func waitForDeploymentReplicasFunc(conn *kubernetes.Clientset, ns, name string)
 			return resource.NonRetryableError(err)
 		}
 
+		var specReplicas int32 = 1 // default, acording to API docs
```
dak1n1, line 411: According to Go docs, int32 is the default size of `int` on a 32-bit system. But on 64-bit systems, it's the size of int64. Why specify the size?

jrhouston:
We should probably change this function to use the helper from kubectl instead of reimplementing the logic. I did this in the statefulset resource: https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/kubernetes/resource_kubernetes_stateful_set.go#L232

kubernetes/resource_kubernetes_deployment.go
alexsomesan: It's the type of `*dply.Spec.Replicas`. Figured I should go with the same.

dak1n1 (approved):
LGTM

alexsomesan:
> We should probably change this function to use the helper from kubectl instead of reimplementing the logic. I did this in the statefulset resource: https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/kubernetes/resource_kubernetes_stateful_set.go#L232

Agree. This is just a UX hot fix done in 5 minutes. I'll add an issue for switching it to the helper.

PR #1000, opened 2020-09-09
Script for cleaning up orphaned ELBs after test run

jrhouston:
### Description

This adds a bash script to find and delete ELBs that are hanging around from the test cluster after the tests have run.

### DIFF
--- a/kubernetes/test-infra/eks/scripts/delete_orphan_elbs.sh
+++ b/kubernetes/test-infra/eks/scripts/delete_orphan_elbs.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash +e
+
+ORPHAN_ELBS=$(aws elb describe-tags \
+    --load-balancer-names $(aws elb describe-load-balancers | jq -r '.LoadBalancerDescriptions[] | .LoadBalancerName') | \
+    jq -r ".TagDescriptions[] | select(.Tags[] | .Key == \"kubernetes.io/cluster/$(terraform output cluster_name)\") | .LoadBalancerName")
+
+for ELB in $ORPHAN_ELBS; do
+    aws elb delete-load-balancer --load-balancer-name $ELB
+done
--- a/kubernetes/test-infra/eks/output.tf
+++ b/kubernetes/test-infra/eks/output.tf
@@ -2,3 +2,6 @@ output "kubeconfig_path" {
   value = abspath("${local.kubeconfig_path}/${local.kubeconfig_name}")
 }
 
+output "cluster_name" {
+  value = module.vpc.cluster_name
+}


kubernetes/test-infra/eks/scripts/delete_orphan_elbs.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash +e
```
dak1n1, line 1: I like that you're using the errexit flag here. However, the `+e` on this line isn't running on my local system (Fedora 32). It gives me this error:

```
[dakini@dax terraform-provider-kubernetes (eks-cleanup *$%=)]$  kubernetes/test-infra/eks/scripts/delete_orphan_elbs.sh
/usr/bin/env: ‘bash +e’: No such file or directory
/usr/bin/env: use -[v]S to pass options in shebang lines
```

How about something like this instead? You could even add on a pipefail, to catch errors that otherwise would be missed due to ` | ` pipes.

```suggestion
#!/usr/bin/env bash
set -o errexit
set -o pipefail
```

dak1n1: Could you also make this file executable? I see our other scripts have execute permissions.

```
[dakini@dax terraform-provider-kubernetes (eks-cleanup *$%=)]$ ls -l scripts/
total 20
-rwxrwxr-x. 1 dakini dakini 294 Apr 30 13:16 build.sh
-rwxrwxr-x. 1 dakini dakini 951 Aug 31 14:06 changelog-links.sh
-rwxrwxr-x. 1 dakini dakini 692 Sep  9 16:14 errcheck.sh
-rwxrwxr-x. 1 dakini dakini 367 Feb 11  2020 gofmtcheck.sh
-rwxrwxr-x. 1 dakini dakini 590 Apr 30 13:16 gogetcookie.sh
```

jrhouston: Good call, thanks!

dak1n1:
It looks like it runs! There are a couple areas where we could add some guardrails to prevent user error, but only if you think it's worthwhile:

1. Does `terraform output` work? If not, the user is probably in the wrong directory (like I was).

2. Does `aws elb...` find any load balancers? If not, you get an error that doesn't make much sense. I didn't write it down, but if you run the script and there's nothing for it to do, it errors.

I ran the first two commands here to ensure that my environment was going to work before running the script. The output after that is just to show that it works.

```
[dakini@dax eks (eks-cleanup *$%=)]$ terraform output cluster_name
k8s-acc-74fc
[dakini@dax eks (eks-cleanup *$%=)]$ aws elb describe-load-balancers | jq -r '.LoadBalancerDescriptions[] | .LoadBalancerName'
a0c0f7877815949a8a00759766fde727

[dakini@dax eks (eks-cleanup *$%=)]$ ./scripts/delete_orphan_elbs.sh 
/usr/bin/env: ‘bash +e’: No such file or directory
/usr/bin/env: use -[v]S to pass options in shebang lines
[dakini@dax eks (eks-cleanup *$%=)]$ bash ./scripts/delete_orphan_elbs.sh 
[dakini@dax eks (eks-cleanup *$%=)]$ aws elb describe-load-balancers
{
    "LoadBalancerDescriptions": []
}
[dakini@dax eks (eks-cleanup *$%=)]$ pwd
/home/dakini/go/src/github.com/hashicorp/terraform-provider-kubernetes/kubernetes/test-infra/eks
[dakini@dax eks (eks-cleanup *$%=)]$ 
```

dak1n1:
This is kind of interesting... even though the ELB was deleted, I'm still seeing `terraform destroy` fail the first 2 times I ran it:

```
module.vpc.aws_vpc.k8s-acc: Still destroying... [id=vpc-0082ebfc85049be3b, 4m50s elapsed]

Error: Error deleting VPC: DependencyViolation: The vpc 'vpc-0082ebfc85049be3b' has dependencies and cannot be deleted.
        status code: 400, request id: cf9f65cd-dc76-4f9d-b428-46d501990db5
```

When I log into the AWS web console, I can see the VPC is unable to delete because there is a security group that was associated with the deleted ELB. Since the SG is still around, the VPC can't delete. But when I delete the SG manually, the `terraform destroy` succeeds.

I wonder if you could add something that detects the security group associated with the ELB, and delete that too. Hmm. But that might not work if the cluster is using that SG still.

jrhouston:
Don't need this anymore.

PR #1052, opened 2020-11-02
Credentials changes for v2.0.0

jrhouston:
### Description

Removes the load_config_file attribute and drops implicit support for `KUBECONFIG`. 

### Acceptance tests
- [ ] ~Have you added an acceptance test for the functionality being added?~

### Release Note

```release-note
Remove `load_config_file` attribute from provider block
Remove explicit support for KUBECONFIG
Remove config_path default to ~/.kube/config
Add config_paths attribute to provider block
```

### Community Note

### DIFF
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -4,9 +4,10 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"log"
 	"net/http"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/logging"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -68,15 +69,10 @@ func Provider() *schema.Provider {
 				Description: "PEM-encoded root certificates bundle for TLS authentication.",
 			},
 			"config_path": {
-				Type:     schema.TypeString,
-				Optional: true,
-				DefaultFunc: schema.MultiEnvDefaultFunc(
-					[]string{
-						"KUBE_CONFIG",
-						"KUBECONFIG",
-					},
-					"~/.kube/config"),
-				Description: "Path to the kube config file, defaults to ~/.kube/config",
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("KUBE_CONFIG_PATH", ""),
+				Description: "Path to the kube config file. Can be set with KUBE_CONFIG_PATH environment variable.",
 			},
 			"config_context": {
 				Type:        schema.TypeString,
@@ -100,12 +96,6 @@ func Provider() *schema.Provider {
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("KUBE_TOKEN", ""),
 				Description: "Token to authenticate an service account",
-			},
-			"load_config_file": {
-				Type:        schema.TypeBool,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("KUBE_LOAD_CONFIG_FILE", true),
-				Description: "Load local kubeconfig.",
 			},
 			"exec": {
 				Type:     schema.TypeList,
@@ -270,40 +260,37 @@ func initializeConfiguration(d *schema.ResourceData) (*restclient.Config, error)
 	overrides := &clientcmd.ConfigOverrides{}
 	loader := &clientcmd.ClientConfigLoadingRules{}
 
-	if d.Get("load_config_file").(bool) {
-		log.Printf("[DEBUG] Trying to load configuration from file")
-		if configPath, ok := d.GetOk("config_path"); ok && configPath.(string) != "" {
-			path, err := homedir.Expand(configPath.(string))
-			if err != nil {
-				return nil, err
+	if configPath, ok := d.GetOk("config_path"); ok && configPath.(string) != "" {
+		path, err := homedir.Expand(configPath.(string))
+		if err != nil {
+			return nil, err
+		}
+		log.Printf("[DEBUG] Configuration file is: %s", path)
+		loader.ExplicitPath = path
+
+		ctxSuffix := "; default context"
+
+		kubectx, ctxOk := d.GetOk("config_context")
+		authInfo, authInfoOk := d.GetOk("config_context_auth_info")
+		cluster, clusterOk := d.GetOk("config_context_cluster")
+		if ctxOk || authInfoOk || clusterOk {
+			ctxSuffix = "; overriden context"
+			if ctxOk {
+				overrides.CurrentContext = kubectx.(string)
+				ctxSuffix += fmt.Sprintf("; config ctx: %s", overrides.CurrentContext)
+				log.Printf("[DEBUG] Using custom current context: %q", overrides.CurrentContext)
 			}
-			log.Printf("[DEBUG] Configuration file is: %s", path)
-			loader.ExplicitPath = path
 
-			ctxSuffix := "; default context"
-
-			kubectx, ctxOk := d.GetOk("config_context")
-			authInfo, authInfoOk := d.GetOk("config_context_auth_info")
-			cluster, clusterOk := d.GetOk("config_context_cluster")
-			if ctxOk || authInfoOk || clusterOk {
-				ctxSuffix = "; overriden context"
-				if ctxOk {
-					overrides.CurrentContext = kubectx.(string)
-					ctxSuffix += fmt.Sprintf("; config ctx: %s", overrides.CurrentContext)
-					log.Printf("[DEBUG] Using custom current context: %q", overrides.CurrentContext)
-				}
-
-				overrides.Context = clientcmdapi.Context{}
-				if authInfoOk {
-					overrides.Context.AuthInfo = authInfo.(string)
-					ctxSuffix += fmt.Sprintf("; auth_info: %s", overrides.Context.AuthInfo)
-				}
-				if clusterOk {
-					overrides.Context.Cluster = cluster.(string)
-					ctxSuffix += fmt.Sprintf("; cluster: %s", overrides.Context.Cluster)
-				}
-				log.Printf("[DEBUG] Using overidden context: %#v", overrides.Context)
+			overrides.Context = clientcmdapi.Context{}
+			if authInfoOk {
+				overrides.Context.AuthInfo = authInfo.(string)
+				ctxSuffix += fmt.Sprintf("; auth_info: %s", overrides.Context.AuthInfo)
 			}
+			if clusterOk {
+				overrides.Context.Cluster = cluster.(string)
+				ctxSuffix += fmt.Sprintf("; cluster: %s", overrides.Context.Cluster)
+			}
+			log.Printf("[DEBUG] Using overidden context: %#v", overrides.Context)
 		}
 	}
 
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -7,7 +7,7 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"strings"
+	"path/filepath"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 
@@ -279,7 +279,7 @@ func initializeConfiguration(d *schema.ResourceData) (*restclient.Config, error)
 	} else if v := os.Getenv("KUBE_CONFIG_PATHS"); v != "" {
 		// NOTE we have to do this here because the schema
 		// does not yet allow you to set a default for a TypeList
-		configPaths = strings.Split(v, ":")
+		configPaths = filepath.SplitList(v)
 	}
 
 	if len(configPaths) > 0 {
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -203,12 +203,19 @@ type kubeClientsets struct {
 	config              *restclient.Config
 	mainClientset       *kubernetes.Clientset
 	aggregatorClientset *aggregator.Clientset
+
+	configData *schema.ResourceData
 }
 
 func (k kubeClientsets) MainClientset() (*kubernetes.Clientset, error) {
 	if k.mainClientset != nil {
 		return k.mainClientset, nil
 	}
+
+	if err := checkConfigurationValid(k.configData); err != nil {
+		return nil, err
+	}
+
 	if k.config != nil {
 		kc, err := kubernetes.NewForConfig(k.config)
 		if err != nil {
@@ -280,10 +287,6 @@ See our documentation at: %s`, authDocumentationURL)
 }
 
 func providerConfigure(ctx context.Context, d *schema.ResourceData, terraformVersion string) (interface{}, diag.Diagnostics) {
-	if err := checkConfigurationValid(d); err != nil {
-		return nil, diag.FromErr(err)
-	}
-
 	// Config initialization
 	cfg, err := initializeConfiguration(d)
 	if err != nil {
@@ -310,6 +313,7 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData, terraformVer
 		config:              cfg,
 		mainClientset:       nil,
 		aggregatorClientset: nil,
+		configData:          d,
 	}
 	return m, diag.Diagnostics{}
 }
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -79,7 +79,7 @@ func Provider() *schema.Provider {
 			"config_path": {
 				Type:          schema.TypeString,
 				Optional:      true,
-				DefaultFunc:   schema.EnvDefaultFunc("KUBE_CONFIG_PATH", ""),
+				DefaultFunc:   schema.EnvDefaultFunc("KUBE_CONFIG_PATH", nil),
 				Description:   "Path to the kube config file. Can be set with KUBE_CONFIG_PATH.",
 				ConflictsWith: []string{"config_paths"},
 			},
@@ -238,6 +238,8 @@ func inCluster() bool {
 	return host != "" && port != ""
 }
 
+var authDocumentationURL = "https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#authentication"
+
 func checkConfigurationValid(d *schema.ResourceData) error {
 	if inCluster() {
 		log.Printf("[DEBUG] Terraform appears to be running inside the Kubernetes cluster")
@@ -263,7 +265,9 @@ func checkConfigurationValid(d *schema.ResourceData) error {
 	}
 
 	return fmt.Errorf(`provider not configured: you must configure a path to your kubeconfig
-or explicitly supply credentials in the provider block`)
+or explicitly supply credentials via the provider block or environment variables.
+
+See our documentation at: %s`, authDocumentationURL)
 }
 
 func providerConfigure(ctx context.Context, d *schema.ResourceData, terraformVersion string) (interface{}, diag.Diagnostics) {
@@ -326,6 +330,10 @@ func initializeConfiguration(d *schema.ResourceData) (*restclient.Config, error)
 			if err != nil {
 				return nil, err
 			}
+			if _, err := os.Stat(path); err != nil {
+				return nil, fmt.Errorf("could not open kubeconfig %q: %v", p, err)
+			}
+
 			log.Printf("[DEBUG] Using kubeconfig: %s", path)
 			expandedPaths = append(expandedPaths, path)
 		}
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -77,10 +77,11 @@ func Provider() *schema.Provider {
 				Description: "A list of paths to kube config files. Can be set with KUBE_CONFIG_PATHS environment variable.",
 			},
 			"config_path": {
-				Type:        schema.TypeString,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("KUBE_CONFIG_PATH", ""),
-				Description: "Path to the kube config file. Can be set with KUBE_CONFIG_PATH.",
+				Type:          schema.TypeString,
+				Optional:      true,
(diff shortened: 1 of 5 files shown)


kubernetes/provider.go
```diff
@@ -270,40 +262,50 @@ func initializeConfiguration(d *schema.ResourceData) (*restclient.Config, error)
...
 
-	if d.Get("load_config_file").(bool) {
-		log.Printf("[DEBUG] Trying to load configuration from file")
-		if configPath, ok := d.GetOk("config_path"); ok && configPath.(string) != "" {
-			path, err := homedir.Expand(configPath.(string))
+	configPaths := []string{}
+	if v, ok := d.Get("config_paths").([]string); ok && len(v) > 0 {
+		configPaths = v
+	} else if v := os.Getenv("KUBE_CONFIG_PATHS"); v != "" {
+		// NOTE we have to do this here because the schema
```
jrhouston, line 269: There's an open issue for this https://github.com/hashicorp/terraform-plugin-sdk/issues/142

kubernetes/provider.go
```diff
@@ -67,16 +70,11 @@ func Provider() *schema.Provider {
...
-				Type:     schema.TypeString,
-				Optional: true,
-				DefaultFunc: schema.MultiEnvDefaultFunc(
-					[]string{
-						"KUBE_CONFIG",
-						"KUBECONFIG",
-					},
-					"~/.kube/config"),
-				Description: "Path to the kube config file, defaults to ~/.kube/config",
+			"config_paths": {
```
jrhouston, line 73: Now that I've implemented this and I looked at it I still kind of want to support `config_path` too because it feels like a slightly nicer user experience for the average user of this. 

It's slightly clunky and surprising to have set an array if you just have one kubeconfig. It's also kind of clunky to have to set `KUBE_CONFIG_PATHS` to a single path if you're using an environment variable. 

There would be a small overhead to having both in the provider block, but I think it's a more intuitive UX for someone just picking up the tool. Leaving the list option for the subset of users that have more complicated environments.

What do you think? @redeux @alexsomesan @dak1n1

jrhouston: I've updated to support both, as per our discussion this week.

kubernetes/provider_test.go
```diff
@@ -86,91 +87,33 @@ func TestProvider_configure(t *testing.T) {
...
-	}
-	if err := os.Unsetenv("KUBE_CLUSTER_CA_CERT_DATA"); err != nil {
-		t.Fatalf("Error unsetting env var KUBE_CLUSTER_CA_CERT_DATA: %s", err)
-	}
-	if err := os.Unsetenv("KUBE_INSECURE"); err != nil {
-		t.Fatalf("Error unsetting env var KUBE_INSECURE: %s", err)
-	}
-	if err := os.Unsetenv("KUBE_LOAD_CONFIG_FILE"); err != nil {
-		t.Fatalf("Error unsetting env var KUBE_LOAD_CONFIG_FILE: %s", err)
+	envVars := map[string]string{
```
jrhouston, line 90: Mopped this up because it was tedious. 🧹

kubernetes/provider.go
```diff
@@ -270,40 +268,60 @@ func initializeConfiguration(d *schema.ResourceData) (*restclient.Config, error)
...
+	if v, ok := d.Get("config_path").(string); ok && v != "" {
+		configPaths = []string{v}
+	} else if v, ok := d.Get("config_paths").([]interface{}); ok && len(v) > 0 {
+		for _, p := range v {
+			configPaths = append(configPaths, p.(string))
+		}
+	} else if v := os.Getenv("KUBE_CONFIG_PATHS"); v != "" {
+		// NOTE we have to do this here because the schema
+		// does not yet allow you to set a default for a TypeList
+		configPaths = strings.Split(v, ":")
```
dak1n1, line 282: Splitting the string by `:` will work on Linux and MacOS, but Windows uses a semi-colon as the delimiter for its PATH. Maybe `os.PathListSeparator` would be better to use here than `:`?

jrhouston: Good call – I didn't know about os.PathListSeparator. I just had a poke at the kubectl code and it uses [filepath.SplitList](https://godoc.org/path/filepath#SplitList), which uses os.PathListSeparator underneath so I'm going to update to use that.

Just realized I haven't owned a Windows machine in about 10 years 😅.

kubernetes/provider_test.go
dak1n1: Nice, I actually did the same thing [here](https://github.com/hashicorp/terraform-provider-kubernetes/pull/1074/files#diff-d05e30b029577be42f0bfff95d7b8a6df42c291154a35736ffd33e766d99db3d), but I removed the whole section. I found it wasn't being used, or at least, it had no visible effect when I removed it. That's ok though, I'll remove this from my PR and test again once yours merges.

jrhouston: It seems to be used in this one test that doesn't actually run as part of the acceptance tests: https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/kubernetes/provider_test.go#L70

dak1n1: Yeah, I saw that it didn't run during acceptance tests and that nothing was calling it... but I think I understand its purpose now. Thanks for showing me. I was misinterpreting it as a "TestProvider" configure function. But it actually tests the "provider configure" code, which does serve a purpose. So that's worth keeping.

dak1n1:
Would it be possible to add an error message to let users know why the provider can no longer find their `KUBECONFIG`? When I run the provider using this branch, I get these errors, neither of which tell me I need to set `KUBE_CONFIG_PATH`:

```
kubernetes_pod.main: Creating...

Error: Post "http://localhost/api/v1/namespaces/default/pods": dial tcp [::1]:80: connect: connection refused
```

```
$ make testacc

=== RUN   TestAccKubernetesDataSourcePod_basic
    provider_test.go:157: File config (KUBE_CTX_AUTH_INFO and KUBE_CTX_CLUSTER) or static configuration(KUBE_HOST, KUBE_USER, KUBE_PASSWORD, KUBE_CLUSTER_CA_CERT_DATA) or (KUBE_HOST, KUBE_CLIENT_CERT_DATA, KUBE_CLIENT_KEY_DATA, KUBE_CLUSTER_CA_CERT_DATA) must be set for acceptance tests
--- FAIL: TestAccKubernetesDataSourcePod_basic (0.00s)
```
It looks like we might need an additional check for the configuration during acceptance tests:

```
$ KUBE_CTX_CLUSTER=minikube KUBE_CTX_AUTH_INFO=minikube TESTARGS="-run 'Pod_basic'" make testacc


=== RUN   TestAccKubernetesDataSourcePod_basic
    data_source_kubernetes_pod_test.go:15: Step 1/2 error: Error running apply:
        Error: Post "http://localhost/api/v1/namespaces/default/pods": dial tcp [::1]:80: connect: connection refused


--- FAIL: TestAccKubernetesDataSourcePod_basic (1.22s)
=== RUN   TestAccKubernetesPod_basic
    resource_kubernetes_pod_test.go:28: Step 1/3 error: Error running apply:
        Warning: Interpolation-only expressions are deprecated

          on config052965964/terraform_plugin_test.tf line 64, in resource "kubernetes_pod" "test":
          64:             name     = "${kubernetes_secret.test.metadata.0.name}"

        Terraform 0.11 and earlier required all non-constant expressions to be
        provided via interpolation syntax, but this pattern is now deprecated. To
        silence this warning, remove the "${ sequence from the start and the }"
        sequence from the end of this expression, leaving just the inner expression.

        Template interpolation syntax is still used to construct strings from
        expressions when the template includes multiple interpolation sequences or a
        mixture of literal strings and interpolations. This deprecation applies only
        to templates that consist entirely of a single interpolation sequence.

        (and 4 more similar warnings elsewhere)


        Error: Post "http://localhost/api/v1/namespaces/default/configmaps": dial tcp [::1]:80: connect: connection refused



        Error: Post "http://localhost/api/v1/namespaces/default/secrets": dial tcp [::1]:80: connect: connection refused



        Error: Post "http://localhost/api/v1/namespaces/default/configmaps": dial tcp [::1]:80: connect: connection refused



        Error: Post "http://localhost/api/v1/namespaces/default/secrets": dial tcp [::1]:80: connect: connection refused


--- FAIL: TestAccKubernetesPod_basic (1.29s)
FAIL
```

website/docs/index.html.markdown
```diff
@@ -131,12 +122,12 @@ The following arguments are supported:
 * `client_certificate` - (Optional) PEM-encoded client certificate for TLS authentication. Can be sourced from `KUBE_CLIENT_CERT_DATA`.
 * `client_key` - (Optional) PEM-encoded client certificate key for TLS authentication. Can be sourced from `KUBE_CLIENT_KEY_DATA`.
 * `cluster_ca_certificate` - (Optional) PEM-encoded root certificates bundle for TLS authentication. Can be sourced from `KUBE_CLUSTER_CA_CERT_DATA`.
-* `config_path` - (Optional) Path to the kube config file. Can be sourced from `KUBE_CONFIG` or `KUBECONFIG`. Defaults to `~/.kube/config`.
+* `config_path` - (Optional) A path to a kube config files. Can be sourced from `KUBE_CONFIG_PATH`.
```
dak1n1, line 125:
```suggestion
* `config_path` - (Optional) A path to a kube config file. Can be sourced from `KUBE_CONFIG_PATH`.
```

kubernetes/provider.go
```diff
@@ -67,16 +70,17 @@ func Provider() *schema.Provider {
...
-					[]string{
-						"KUBE_CONFIG",
-						"KUBECONFIG",
-					},
-					"~/.kube/config"),
-				Description: "Path to the kube config file, defaults to ~/.kube/config",
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("KUBE_CONFIG_PATH", ""),
+				Description: "Path to the kube config file. Can be set with KUBE_CONFIG_PATH.",
```
dak1n1, line 83: This is great! Could we also add `ConflictsWith = []string{"config_paths"}`? And the equivalent for config_path, so that the SDK lets a user know if they've specified both? Otherwise we're selecting one of the two at random, and the user might find it unexpected. In my testing, it will yield this result:

```
[dakini@dax kubeconfig]$ terraform validate

Error: ConflictsWith

"config_paths": conflicts with config_path


Error: ConflictsWith

"config_path": conflicts with config_paths
```

However, if specified as environment variables, it doesn't give the user any error message:

```
$ KUBE_CONFIG_PATH=$KUBECONFIG KUBE_CONFIG_PATHS=$KUBECONFIG terraform init
$ KUBE_CONFIG_PATH=$KUBECONFIG KUBE_CONFIG_PATHS=$KUBECONFIG terraform validate
Success! The configuration is valid.

$ KUBE_CONFIG_PATH=$KUBECONFIG KUBE_CONFIG_PATHS=$KUBECONFIG terraform apply --auto-approve
kubernetes_pod.main: Creating...
kubernetes_pod.main: Creation complete after 6s [id=default/test-pod]

```

That validation difference between the environment variables and the equivalent in HCL might be an SDK bug.

jrhouston:
> Would it be possible to add an error message to let users know why the provider can no longer find their `KUBECONFIG`? When I run the provider using this branch, I get these errors, neither of which tell me I need to set `KUBE_CONFIG_PATH`:

So this looks to be because of this conditional: https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/kubernetes/provider.go#L244 
From looking into the linked issues it seems that this is fixed in 0.14 but will still be an issue for previous versions so we'll need to leave this in for now.

We could explicitly check the attributes supplied to the provider block but the tricky thing here is, to know for sure what's going on, we have to check:

1. They haven't supplied any credentials attributes, AND
2. They're not running inside a cluster, which is the default behaviour for the client if you don't supply any credentials.

If we also add `in_cluster_config` an explicit attribute too, then we could just look at the supplied attributes and then tell them that they need to specify at least one of these things. What do you think?

cc @alexsomesan

dak1n1:
I was doing some additional manual testing this morning when I found what might be a bug.

I'll explain my configuration. There are two clusters. The first cluster gets a specific pod. The second cluster gets a different specific pod. And the third pod is supposed to be created on both clusters. That's the part that fails, where I'm testing `config_paths`.

Here's the output that shows the test I'm doing. There are some debug logs at the end: 

https://gist.githubusercontent.com/dak1n1/59933b87a2cef52b189e74e6800a086e/raw/8957133174384f27eebad5f6a52ac4077dfabbbb/gistfile1.txt

I found that even when I removed the parts of the config that try to create a pod on individual clusters, the part that fails is creating the pods on both clusters (`config_paths`). It might be worth adding an acceptance test for this.

kubernetes/provider.go
```diff
@@ -234,7 +233,43 @@ func (k kubeClientsets) AggregatorClientset() (*aggregator.Clientset, error) {
 	return k.aggregatorClientset, nil
 }
 
+func inCluster() bool {
+	host, port := os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT")
+	return host != "" && port != ""
+}
+
+func checkConfigurationValid(d *schema.ResourceData) error {
```
jrhouston, line 241: @dak1n1 This should generate a more helpful error if the user hasn't supplied any config through the provider block or environment variables, or terraform isn't inside a cluster. I've tried a few incantations locally but would appreciate a second 👀 .

```
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.


------------------------------------------------------------------------

Error: provider not configured: you must configure a path to your kubeconfig
or explicitly supply credentials in the provider block
```

dak1n1: Nice! This is exactly what I was hoping for. As a user, if I got this message, the next thing I would do is open up the documentation for the provider, to figure out how to specify this. (Or google the error message because I don't want to navigate through all the docs pages).

But we might be able to save them that step by giving them either the name of the environment variable, or the name of the attribute to put in their config. Something like `KUBE_CONFIG_PATHS=$KUBECONFIG` should work.

I also noticed that when `KUBE_CONFIG_PATH` and `KUBE_CONFIG_PATHS` are both set, it uses `KUBE_CONFIG_PATH`. And if `KUBE_CONFIG_PATH` points to a bad path, it gives this error:

```
$ env |grep KUBE
KUBE_CONFIG_PATH=/bad/path
KUBECONFIG=/home/dakini/.kube/config
KUBE_CONFIG_PATHS=/home/dakini/.kube/config

$ terraform apply --auto-approve
kubernetes_pod.test: Creating...

Error: Post "http://localhost/api/v1/namespaces/default/pods": dial tcp [::1]:80: connect: connection refused
```

It might be worth checking that the file exists, so the error message tells the user what went wrong.

jrhouston: We could put a hint, but we can't assume which method they actually need - and writing instructions for path, basic auth, TLS creds, env vars is going to be quite a lot of information to put in the error message. Perhaps we could just link them to the documentation page for the provider block?

jrhouston: Also yes, confirming those files even exist early on would be a good idea – I'll add that to where it does the path expansion.

dak1n1:
The error message for this is super nice! I did some testing with it today and it was a smooth user experience. So this is looking really good so far.

I tested a new scenario that might need some work: the ability to detect when one cluster has gone away. Here's the scenario:

I have two clusters. By default, the provider will apply all my kubernetes configurations to the first cluster listed in `config_paths`. But what happens when that cluster goes away? I imagine the expected behavior is that the provider will use the next cluster in `config_paths`. However, instead, it fails with this message:

```
kubernetes_pod.test: Creating...

Error: Post "http://localhost/api/v1/namespaces/default/pods": dial tcp [::1]:80: connect: connection refused
```

Here's my provider block:

```
provider "kubernetes" {
  config_paths = [
    "/home/dakini/.kube/second",
    "/home/dakini/.kube/config",
  ]
}
```

When I deleted the minikube cluster named `second`, the provider probably should try next config in line. However, this PR is really good as it is, so we could add this functionality once this one merges. I'll leave that decision to you.

dak1n1 (approved):

kubernetes/provider.go
```diff
@@ -234,7 +233,47 @@ func (k kubeClientsets) AggregatorClientset() (*aggregator.Clientset, error) {
 	return k.aggregatorClientset, nil
 }
 
+func inCluster() bool {
```
alexsomesan, line 236: For this to only yield `true` when indeed run with valid in-cluster configuration, we'd also have to check for the presence of `/var/run/secrets/kubernetes.io/serviceaccount`.

The `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` are only the endpoint information. Valid credentials are also required for a successful client connection. In the case of in-cluster config, they are injected via that volume mount of a service account token. That might not be the case when no service account is specified and `automountServiceAccountToken` is false.

Some details over here https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod

kubernetes/provider.go
```diff
@@ -234,7 +233,47 @@ func (k kubeClientsets) AggregatorClientset() (*aggregator.Clientset, error) {
...
+	host, port := os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT")
+	return host != "" && port != ""
+}
+
+var authDocumentationURL = "https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#authentication"
+
+func checkConfigurationValid(d *schema.ResourceData) error {
+	if inCluster() {
+		log.Printf("[DEBUG] Terraform appears to be running inside the Kubernetes cluster")
```
alexsomesan, line 245: In order for InCluster to count as valid configuration, we should also check for the service account token as mentioned above.

kubernetes/provider.go
```diff
@@ -234,7 +233,47 @@ func (k kubeClientsets) AggregatorClientset() (*aggregator.Clientset, error) {
...
+
+	atLeastOneOf := []string{
+		"host",
+		"config_path",
+		"config_paths",
+		"client_certificate",
+		"token",
+		"exec",
+	}
+	for _, a := range atLeastOneOf {
```
alexsomesan, line 261: I'm not sure I understand this correctly, but it looks like having just "token" or "client_certificate" set would count as valid configuration. Not sure if I'm missing something else.

This could be confusing if not covering all combinations that make a valid configuration.

Also (not 100% sure) but it might be falsely triggered by progressive apply situations. Can we rule those out?

kubernetes/provider.go
```diff
@@ -270,40 +309,64 @@ func initializeConfiguration(d *schema.ResourceData) (*restclient.Config, error)
...
-					overrides.Context.AuthInfo = authInfo.(string)
-					ctxSuffix += fmt.Sprintf("; auth_info: %s", overrides.Context.AuthInfo)
-				}
-				if clusterOk {
-					overrides.Context.Cluster = cluster.(string)
-					ctxSuffix += fmt.Sprintf("; cluster: %s", overrides.Context.Cluster)
-				}
-				log.Printf("[DEBUG] Using overidden context: %#v", overrides.Context)
+			if _, err := os.Stat(path); err != nil {
+				return nil, fmt.Errorf("could not open kubeconfig %q: %v", p, err)
```
alexsomesan, line 334: Would this fail to configure the provider if a single file is missing?
IMO, we should consider all entries in the list in a best effort to find valid credentials.

Also, in case of multiple files present it could be argued that a valid CurrentContext should be present to eliminate ambiguities about which cluster would be chosen to connect to. Otherwise we might end up applying changes to an undesirable destination.

alexsomesan (changes requested):
Finally we get rid of `load_config_file`... Nicely done!

I've had a look around and I found a spot around the InCluster stuff that I think might cause some trouble. Left some comments in there.

kubernetes/provider.go
jrhouston: The behaviour of kubectl when using a `KUBECONFIG` with multiple paths is not to error if one of the paths does not exist. That being said I think if we're erring on the side of explicitness by not supporting `KUBECONFIG` anymore (and allowing this to be set as a list in the provider block) then we should also be more explicit here and tell the user that one of the paths they specified does not exist rather than silently skipping it.

kubernetes/provider.go
jrhouston: The intent of this code is not to cover all valid possible configurations, it is to rule out that:

1. The user has left the provider block is empty 
2. Terraform is running inside a cluster. 
 
The purpose of this is so we can tell the user to explicitly configure instead of generating the confusing "could not connect" error at plan time when the client actually gets initialized. Once we fix [this](https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/kubernetes/provider.go#L245) we should be able to generate more meaningful errors that come from client go, as it will warn you about these problems.

We can use `ConflictsWith` and `RequiredWith` to validate that the right combination of attributes is being used together, but I would propose doing that in another pull request.

jrhouston:
> When I deleted the minikube cluster named second, the provider probably should try next config in line. However, this PR is really good as it is, so we could add this functionality once this one merges. I'll leave that decision to you.

@dak1n1 Did you confirm this is what's supposed to happen when you use `kubectl`? We just pass along the list of paths to client-go so I'm not sure how we would infer this. In any case I agree we should solve for that in another PR.

kubernetes/provider.go
alexsomesan: Sounds good. I didn't catch those nuances while reading through it the first time. Let try this approach out and see how it improves things. Looks like https://github.com/hashicorp/terraform/issues/24055 is confirmed fixed so we're clear to move in that regard.
I agree with doing things incrementally. Let's re-evaluate after this is released.

kubernetes/provider.go
alexsomesan: Sounds good then. Let's go with this!
We might need to keep an eye out for corner cases where this list would be set but not all entries might be actually present in the file system (yet). It's a stretch and quite unlikely, but we've seen funny things before. 

What do you think about requiring a CurrentContext to be set when this list has more than one entry? Just as a thought experiment for now. Doesn't need to be part of this PR.

alexsomesan (approved):
All good on my side. This looks like it's going to improve quality of life for everyone.
Excited to try it out!

PR #1058, opened 2020-11-05
Fix HPA causing deployment update on every apply

innovate-invent:
An associated horizontal pod autoscaler modifies the replica count. Currently the replicas setting is optional but defaults to 1 rather than accepting the computed value. This causes the deployment to update on every apply back to a size of 1.

This is such a trivial change I don't think it warrants an acceptance test of its own.


jrhouston:
This should be fixed in the latest version of the provider.

PR #1068, opened 2020-11-16
Add configurable provider level ignored metadata keys

### LINKED ISSUE #746: Feature: Provider level annotation ignore-list
As per this PR https://github.com/terraform-providers/terraform-provider-kubernetes/pull/741 there are a number of possible automatically applied annotations that might trigger an unwanted diff.

Rather than call out 'kubernetes.io' for special mention in the code (e.g. here https://github.com/terraform-providers/terraform-provider-kubernetes/blob/master/kubernetes/structures.go#L161) would it be possible to pass a list of hostnames at the provider level?

Thank you

ctrox:
This introduces a new provider configuration parameter
`ignored_metadata_keys` that allows the user to specify expressions to
ignore matching metadata annotations and labels on all kubernetes
resources.

For example:

```terraform
provider "kubernetes" {
  ignored_metadata_keys = ["cattle.io/status", "lifecycle.cattle.io/create.namespace-auth"]
}
```

This tries to be backwards compatible by hardcoding a list of internal
keys in `internalKeys`, which were previously ignored already.

I had to change a few things in the provider config to pass the `ignoredKeys`
to every resource. Let me know if there's a better approach on how to solve that.
It was a bit tedious to have to pass this down everywhere but I cannot think
of another way without introducing global state.

### Acceptance tests
~~- [ ] Have you added an acceptance test for the functionality being added?~~

So far there are no acceptance tests for the provider config itself. I have modified the existing relevant unit test.

### Release Note

```release-note
Add configurable provider level ignored metadata keys
```

### References

Fixes #746

### Community Note

### DIFF
--- a/kubernetes/data_source_kubernetes_namespace.go
+++ b/kubernetes/data_source_kubernetes_namespace.go
@@ -2,11 +2,12 @@ package kubernetes
 
 import (
 	"context"
+	"log"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	v1 "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"log"
 )
 
 func dataSourceKubernetesNamespace() *schema.Resource {
@@ -51,7 +52,7 @@ func dataSourceKubernetesNamespaceRead(ctx context.Context, d *schema.ResourceDa
 		return diag.FromErr(err)
 	}
 	log.Printf("[INFO] Received namespace: %#v", namespace)
-	err = d.Set("metadata", flattenMetadata(namespace.ObjectMeta, d))
+	err = d.Set("metadata", flattenMetadata(namespace.ObjectMeta, d, meta.(ProviderConfig).IgnoredKeys()))
 	if err != nil {
 		return diag.FromErr(err)
 	}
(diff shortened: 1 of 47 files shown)


wpjunior:
Hi @ctrox, in the newest version of terraform you could use `lifecycle.ignore_changes` as well

see my example:
```tf
resource "kubernetes_namespace" "tsuru-system" {
  metadata {
    name = "tsuru-system"
    labels = {
      "tsuru.io/cluster-accessible" = "true"
    }
  }

  lifecycle {
    prevent_destroy = true
    ignore_changes = [
      metadata[0].annotations["cattle.io/status"],
      metadata[0].annotations["lifecycle.cattle.io/create.namespace-auth"]
    ]
  }
}```

Documentation: https://www.terraform.io/docs/language/meta-arguments/lifecycle.html#ignore_changes

ctrox:
> Hi @ctrox, in the newest version of terraform you could use `lifecycle.ignore_changes` as well

I'm aware of that but it is just not a great solution. It only works if you also add placeholder annotations like described [here](https://github.com/rancher/rancher/issues/12821#issuecomment-516421250). As Rancher adds these annotations to every namespace and every ingress resource in a cluster. We would be adding annotations to a ton of resources. Plus in the future, everytime someone creates a namespace with terraform, we need to think about adding that ugly hack again just so it won't show up in every plan.

amithkk:
This is still very much a problem. Related to https://github.com/rancher/terraform-provider-rancher2/issues/27 (which was erroneously closed)

jrhouston:
This was added in https://github.com/hashicorp/terraform-provider-kubernetes/commit/a0768145ba442faa3c939cc764a2d47b3aac1737

PR #1082, opened 2020-12-04
Add StateUpgraders for resources field change

jrhouston:
### Description

This PR adds StateUpgraders to handle migrating resources with `resources` fields from TypeList to TypeMap, without this Terraform will error out when trying to parse the old version of the state.

### Acceptance tests

```

...
```

### Release Note

```release-note
NONE
```

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_cron_job_migrate.go
+++ b/kubernetes/resource_kubernetes_cron_job_migrate.go
@@ -1,0 +1,17 @@
+package kubernetes
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceKubernetesCronJobV0() *schema.Resource {
+	schemaV1 := resourceKubernetesCronJobSchemaV1()
+	schemaV0 := patchJobTemplatePodSpecWithResourcesFieldV0(schemaV1)
+	return &schema.Resource{Schema: schemaV0}
+}
+
+func resourceKubernetesCronJobUpgradeV0(ctx context.Context, rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+	return upgradeJobTemplatePodSpecWithResourcesFieldV0(ctx, rawState, meta)
+}
--- a/kubernetes/schema_resources_migrate.go
+++ b/kubernetes/schema_resources_migrate.go
@@ -47,7 +47,7 @@ func upgradeJobTemplatePodSpecWithResourcesFieldV0(ctx context.Context, rawState
 
 	spec := s[0].(map[string]interface{})
 	jt, ok := spec["job_template"].([]interface{})
-	if !ok || len(jt) > 0 {
+	if !ok || len(jt) == 0 {
 		return rawState, nil
 	}
 
--- a/kubernetes/schema_resources_migrate.go
+++ b/kubernetes/schema_resources_migrate.go
@@ -1,0 +1,90 @@
+package kubernetes
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func patchTemplatePodSpecWithResourcesFieldV0(m map[string]*schema.Schema, templateFieldName string) map[string]*schema.Schema {
+	spec := m["spec"].Elem.(*schema.Resource)
+	template := spec.Schema[templateFieldName].Elem.(*schema.Resource)
+	podSpec := template.Schema["spec"].Elem.(*schema.Resource)
+
+	initContainer := podSpec.Schema["init_container"].Elem.(*schema.Resource)
+	initContainer.Schema["resources"].Elem = &schema.Resource{
+		Schema: resourcesFieldV0(),
+	}
+
+	container := podSpec.Schema["container"].Elem.(*schema.Resource)
+	container.Schema["resources"].Elem = &schema.Resource{
+		Schema: resourcesFieldV0(),
+	}
+	return m
+}
+
+func patchPodSpecWithResourcesFieldV0(m map[string]*schema.Schema, templateFieldName string) map[string]*schema.Schema {
+	spec := m["spec"].Elem.(*schema.Resource)
+
+	initContainer := spec.Schema["init_container"].Elem.(*schema.Resource)
+	initContainer.Schema["resources"].Elem = &schema.Resource{
+		Schema: resourcesFieldV0(),
+	}
+
+	container := spec.Schema["container"].Elem.(*schema.Resource)
+	container.Schema["resources"].Elem = &schema.Resource{
+		Schema: resourcesFieldV0(),
+	}
+	return m
+}
+
+func upgradeTemplatePodSpecWithResourcesFieldV0(ctx context.Context, rawState map[string]interface{}, meta interface{}, templateFieldName string) (map[string]interface{}, error) {
+	if s, ok := rawState["spec"].([]interface{}); ok && len(s) > 0 {
+		spec := s[0].(map[string]interface{})
+		if t, ok := spec[templateFieldName].([]interface{}); ok && len(t) > 0 {
+			template := t[0].(map[string]interface{})
+			if ps, ok := template["spec"].([]interface{}); ok && len(ps) > 0 {
+				podSpec := ps[0].(map[string]interface{})
+				upgradedPodSpec := upgradePodSpecWithResourcesFieldV0(podSpec)
+				template["spec"] = []interface{}{upgradedPodSpec}
+			}
+		}
+	}
+	return rawState, nil
+}
+
+func upgradePodSpecWithResourcesFieldV0(rawState map[string]interface{}) map[string]interface{} {
+	if initContainers, ok := rawState["init_container"].([]interface{}); ok && len(initContainers) > 0 {
+		for _, c := range initContainers {
+			initContainer := c.(map[string]interface{})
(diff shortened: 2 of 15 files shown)


dak1n1:
Oops, I just noticed this PR. Sorry for the wait. @jrhouston FYI, I'll be able to review these faster if you assign me as a reviewer (or assign TF-Providers). Either one of those will send me an email, which puts it in my queue.

I started a TC run for this branch, but I think the tests aren't going to show anything particularly useful until we merge this change in with my PodSpec changes (https://github.com/hashicorp/terraform-provider-kubernetes/pull/1074) . But I'll do some manual testing tomorrow and get back to you with some feedback.

jrhouston:
My bad @dak1n1 I thought I had assigned this!

dak1n1:
Here's some raw debug output from my manual testing. It includes a main.tf that you can use to test manually too.

https://gist.githubusercontent.com/dak1n1/3f75ad5478b3e7d916ccca351489a02f/raw/f9a4892e42b8c2c91e4157b979473a305be8dd14/gistfile1.txt

Interestingly, this shows the same error as all of our upgrade tests (`_regression` tests), which matches the [TC test run](https://ci-oss.hashicorp.engineering/viewLog.html?buildId=163190&tab=buildResultsDiv&buildTypeId=Kubernetes_ProviderKubernetesGke118) for this branch.

At first glance, the StateUpgraders do not appear to be working. But I'll test more after lunch.

dak1n1:
I just realized that the `images` lines in my main.tf might be confusing. Those images use localhost because of the way I have minikube set up to host my images, so that I don't pull from dockerhub as much. ([Here is a related script for setting that up](https://gist.githubusercontent.com/dak1n1/7d2a6f9458efb028378d514712866fe2/raw/dde59aa834b7b6e98371fbaff91bd1c7aebf1f65/gistfile1.txt), if you're curious, but it's not needed for testing this PR).

kubernetes/schema_resources_migrate.go
```diff
@@ -0,0 +1,149 @@
...
+		Schema: resourcesFieldV0(),
+	}
+	return m
+}
+
+func upgradeJobTemplatePodSpecWithResourcesFieldV0(ctx context.Context, rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+	s, ok := rawState["spec"].([]interface{})
+	if !ok || len(s) == 0 {
+		return rawState, nil
+	}
```
dak1n1, line 46: This is just a guess on my part, but I think if a user has not specified some of these fields in their config, then they might not be StateUpgraded like they're supposed to.

For example, I had one resource where I didn't specify any resource.requests.limits. I didn't have a `resources{}` block at all in my config. But when I looked at my Terraform state, I saw that `resources{}` (an empty block) had been placed in my state file, and therefore also needed the state upgrade. I think. Again, this is just a hunch.

kubernetes/schema_resources_migrate_test.go
```diff
@@ -0,0 +1,137 @@
...
+								},
+							}},
+						},
+					},
+				}},
+			}},
+		}},
+	}
+
+	actual, _ := upgradeTemplatePodSpecWithResourcesFieldV0(context.Background(), v0, nil)
```
dak1n1, line 132: This is just a best-practice nit on my part, but if a context doesn't have a deadline or a cancel function, it's best to use [context.TODO()](https://golang.org/pkg/context/#TODO). That makes it easier to find/replace if we do end up using a real context some day.

```suggestion
	actual, _ := upgradeTemplatePodSpecWithResourcesFieldV0(context.TODO(), v0, nil)
```

kubernetes/schema_resources_migrate.go
jrhouston: This function looks at the state though, not their config – if it's not in `rawState` then there's nothing there to upgrade.

kubernetes/resource_kubernetes_cron_job_migrate.go
```diff
@@ -0,0 +1,17 @@
...
+
+func resourceKubernetesCronJobV0() *schema.Resource {
+	schemaV1 := resourceKubernetesCronJobSchemaV1()
+	schemaV0 := patchJobTemplatePodSpecWithResourcesFieldV0(schemaV1)
+	return &schema.Resource{Schema: schemaV0}
+}
+
+func resourceKubernetesCronJobUpgradeV0(ctx context.Context, rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+	return upgradeJobTemplatePodSpecWithResourcesFieldV0(ctx, rawState, meta)
+}
```
dak1n1, line 17: How would you feel about adding some debug logging to all the files named `_migrate.go`? Like this one https://github.com/hashicorp/terraform-provider-aws/blob/master/aws/resource_aws_subnet_migrate.go

That would tell us when it finds an empty state for this resource and maybe give us a before/after view of it.

jrhouston: Good idea, I'll add in some logging.

dak1n1:
I tested this again today. I think it might actually be working. I just need to make a change in the _regression tests to ensure the state isn't read until the StateUpgraders have run. I'm testing out that theory now.

dak1n1:
hmm.  No luck. It's still failing in my manual testing. But you can use these files for manual testing, if you want.

[main.tf for kubernetes provider 1.13](https://gist.githubusercontent.com/dak1n1/a40f6aafe5a0e482026807ecee158b9a/raw/5cb5c877f65c45bd0de801cb38c0b29a26ba952d/gistfile1.txt)
[main.tf for kubernetes provider 2.0](https://gist.githubusercontent.com/dak1n1/b70ad48f3bd4a96e12c53b64674305e9/raw/8ca0630982703d10296e7f77bbbc673a196ecc07/gistfile1.txt)

Using those configs, I'm getting these errors:

```
[dakini@dax stateupgraders]$ terraform plan
kubernetes_daemonset.test: Refreshing state... [id=default/test]
kubernetes_stateful_set.test: Refreshing state... [id=default/test]
kubernetes_pod.test: Refreshing state... [id=default/test]
kubernetes_job.test: Refreshing state... [id=default/test]
kubernetes_deployment.test: Refreshing state... [id=default/test]

Error: missing expected {


[dakini@dax stateupgraders]$ tfapply
kubernetes_job.test: Refreshing state... [id=default/test]
kubernetes_pod.test: Refreshing state... [id=default/test]
kubernetes_deployment.test: Refreshing state... [id=default/test]
kubernetes_daemonset.test: Refreshing state... [id=default/test]
kubernetes_stateful_set.test: Refreshing state... [id=default/test]

Error: missing expected {
```

jrhouston:
OK I think I figured out the problem here – it was with `kubernetes_cron_job`. I had the opposite logic from what I wanted [here](https://github.com/hashicorp/terraform-provider-kubernetes/pull/1082/commits/edbc4d212f1b1b8d9966720e5388346381c8cc4a) and it was causing the upgrader function to short circuit and return the non-upgraded state.

Very frustrating to debug. That error message that suggests a syntax problem but doesn't give you any trace or position 🙀

dak1n1:
It works! Here is output from my manual testing. https://gist.githubusercontent.com/dak1n1/059681546305c1daf82f18d8fad85735/raw/b382f2b1cb216d39f7f0e5fc5a4d2eb456ecd87f/gistfile1.txt

We'll have to mention in our upgrade guide that users need to `terraform apply` even if `terraform plan` shows that no changes are needed. I had to do that in order for the StateUpgraders to run. I'll give the code a read next.

kubernetes/schema_resources_migrate.go
```diff
@@ -0,0 +1,149 @@
...
+				}
+				if lim, ok := resources["limits"].([]interface{}); ok && len(lim) > 0 {
+					limits := lim[0].(map[string]interface{})
+					resources["limits"] = limits
+				}
+			}
+		}
+	}
+	return rawState
+}
```
dak1n1, line 149: I can tell by my testing that this function is doing what it's supposed to do, but I can't seem to figure out how it works. Maybe next year we can refactor it to be easier to follow. Specifically, I'm looking at how this function returns `rawState`... but `rawState` is never modified (as far as I can tell). A variable named `resources` is created and modified to contain the right information, but I have no idea how that makes its way into `rawState`. :thinking: 

To give you a clearer picture of what I mean, here's how I made my equivalent of this function for a StateUpgrader: https://github.com/hashicorp/terraform-provider-kubernetes/blob/74244d7f1902e3c586657d50e81ceb896a01e30c/kubernetes/resource_kubernetes_service_migrate.go#L179-L202

It does a a couple things:
1. Create a new object to hold the migrated state.
2. Descend into the old object, checking that certain fields are set, and copying them into the new object if they do.

I see how your function descends into the old object and checks fields too. And copies those fields into a new object... I just don't understand what happens after that, since the old object is returned.

dak1n1: I'm finding I have this same confusion with every function that returns rawState instead of the new objects. :exploding_head:  I trust that it works. It's just not something I'm familiar with.

dak1n1 (approved):
This all seems to work. Thanks for leading the way through this uncharted territory! It's some pretty challenging stuff.

kubernetes/schema_resources_migrate.go
jrhouston: Yeah it is confusing, we could rewrite it to return an entirely new object like you have done in your other PR. It works because maps are mutable, so even if you assign a map within a map to a new var you are still modifying the underlying map, not creating a new one. 🤯

Issue #1085, opened 2020-12-07
Pod with volume is recreated on each apply


dak1n1:
### Terraform Version, Provider Version and Kubernetes Version

It's a problem in the latest release, and in the current version of `master` branch.

```
$ terraform version
Terraform v0.14.0
+ provider registry.terraform.io/hashicorp/kubernetes v1.13.3
```

### Affected Resource(s)

kubernetes_pod

### Terraform Configuration Files
```hcl
resource "kubernetes_persistent_volume_claim" "test" {
  wait_until_bound = false
  metadata {
    name = "test"
  }
  spec {
    access_modes = ["ReadWriteOnce"]
    resources {
      requests = {
        storage = "1Gi"
      }
    }
  }
}

resource "kubernetes_pod" "main" {
  metadata {
    name = "test"
  }
  spec {
    automount_service_account_token = true
    container {
      name = "default"
      image = "localhost:5000/alpine:3.12.1"
      command = ["sleep", "3600s"]
      volume_mount {
        mount_path = "/etc/test"
        name = "pvc"
      }
    }
    volume {
      name = "pvc"
      persistent_volume_claim {
        claim_name = kubernetes_persistent_volume_claim.test.metadata[0].name
      }
    }
  }
}
```

### Debug Output

https://gist.githubusercontent.com/dak1n1/3a481d1f3d7821dae229a9ceaec8351c/raw/b553996ef15702cc4ff120f670ad0659d0b0f30b/gistfile1.txt

The important part seems to be the volume and volume mount.

```
              ~ volume_mount { # forces replacement
                    name              = "pvc"
                    # (3 unchanged attributes hidden)
                }
              - volume_mount { # forces replacement
                  - mount_path = "/var/run/secrets/kubernetes.io/serviceaccount" -> null
                  - name       = "default-token-pf8cs" -> null
                  - read_only  = true -> null
                }
            }

          + image_pull_secrets {
              + name = (known after apply)
            }

          - volume { # forces replacement
              - name = "default-token-pf8cs" -> null

              - secret {
                  - default_mode = "0644" -> null
                  - optional     = false -> null
                  - secret_name  = "default-token-pf8cs" -> null
                }
            }
```

In the Pod resource, these are updated each time `terraform apply` is run. But in the Deployment resource, these are not updated. Somehow Deployment is working properly, but Pod is not, even though they share a lot of the same code.

### Panic Output

N/A

### Steps to Reproduce

[...]

PR #1096, opened 2020-12-15
Remove default token volume and volume_mount from Pod state

dak1n1:
Fixes https://github.com/hashicorp/terraform-provider-kubernetes/issues/1085

### Acceptance tests

```
```

### Release Note

```release-note
Ensure default service account token is not saved in Pod state.
```

### Community Note

### DIFF
--- a/kubernetes/structures_container.go
+++ b/kubernetes/structures_container.go
@@ -414,6 +414,7 @@ func flattenContainers(in []v1.Container, serviceAccountRegex string) ([]interfa
 				}
 				if nameMatchesDefaultToken {
 					v.VolumeMounts = removeVolumeMountFromContainer(num, v.VolumeMounts)
+					break
 				}
 			}
 
--- a/kubernetes/structures_container.go
+++ b/kubernetes/structures_container.go
@@ -262,14 +262,6 @@ func flattenContainerVolumeMounts(in []v1.VolumeMount) ([]interface{}, error) {
 	att := make([]interface{}, len(in))
 
 	for i, v := range in {
-		// To avoid perpetual diff, skip adding default service account volumeMount to state.
-		skip, err := volumeMountIsDefaultServiceAccount(&v)
-		if err != nil {
-			return att, err
-		}
-		if skip {
-			continue
-		}
 		m := map[string]interface{}{}
 		m["read_only"] = v.ReadOnly
 
@@ -290,17 +282,6 @@ func flattenContainerVolumeMounts(in []v1.VolumeMount) ([]interface{}, error) {
 		att[i] = m
 	}
 	return att, nil
-}
-
-func volumeMountIsDefaultServiceAccount(v *v1.VolumeMount) (bool, error) {
-	nameMatchesDefaultToken, err := regexp.MatchString("default-token-([a-z0-9]{5})", v.Name)
-	if err != nil {
-		return false, err
-	}
-	if v.MountPath == "/var/run/secrets/kubernetes.io/serviceaccount" && nameMatchesDefaultToken {
-		return true, nil
-	}
-	return false, nil
 }
 
 func flattenContainerEnvs(in []v1.EnvVar) []interface{} {
--- a/kubernetes/structures_container.go
+++ b/kubernetes/structures_container.go
@@ -4,6 +4,7 @@ import (
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"regexp"
 )
 
 func flattenCapability(in []v1.Capability) []string {
@@ -259,7 +260,16 @@ func flattenValueFrom(in *v1.EnvVarSource) []interface{} {
 
 func flattenContainerVolumeMounts(in []v1.VolumeMount) ([]interface{}, error) {
 	att := make([]interface{}, len(in))
+
 	for i, v := range in {
+		// To avoid perpetual diff, skip adding default service account volumeMount to state.
+		skip, err := volumeMountIsDefaultServiceAccount(&v)
+		if err != nil {
+			return att, err
+		}
+		if skip {
+			continue
+		}
 		m := map[string]interface{}{}
 		m["read_only"] = v.ReadOnly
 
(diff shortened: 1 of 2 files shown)


kubernetes/structures_pod.go
```diff
@@ -107,6 +108,23 @@ func flattenPodSpec(in v1.PodSpec) ([]interface{}, error) {
...
+		if in.AutomountServiceAccountToken != nil && *in.AutomountServiceAccountToken == true {
+			for i, volume := range in.Volumes {
+				nameMatchesDefaultToken, err := regexp.MatchString("default-token-([a-z0-9]{5})", volume.Name)
+				if err != nil {
+					return []interface{}{att}, err
+				}
+				if nameMatchesDefaultToken && volume.Secret != nil {
+					for _, secretVolume := range volume.Secret.Items {
+						if secretVolume.Path == "/var/run/secrets/kubernetes.io/serviceaccount" {
+							removeVolume(i, in.Volumes)
```
jrhouston, line 122: The reason this isn't working is because your `removeVolume` func returns a new slice, so you need to assign that new slide to `in.Volumes`.

kubernetes/structures_pod.go
```diff
@@ -107,6 +108,23 @@ func flattenPodSpec(in v1.PodSpec) ([]interface{}, error) {
 	}
 
 	if len(in.Volumes) > 0 {
+		// If the serviceAccountToken is being added to the list of volumes,
+		// we need to remove it prior to recording the state.
+		if in.AutomountServiceAccountToken != nil && *in.AutomountServiceAccountToken == true {
+			for i, volume := range in.Volumes {
+				nameMatchesDefaultToken, err := regexp.MatchString("default-token-([a-z0-9]{5})", volume.Name)
```
jrhouston, line 115: What I would do is put this logic [here](https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/kubernetes/structures_pod.go#L256) in the loop that expands the volumes and use a `continue` statement to skip it, rather than removing it here.

dak1n1:
Hmm, well this does look neater. And it successfully skips adding both the volume and volumeMount to state... well, almost. It adds them as empty blocks if I put the logic into `flattenVolumes()` and `flattenContainerVolumeMounts()`, rather than in `flattenPodSpec()` . ("Empty blocks" meaning: `volume{}` and `volume_mount{}`). Here is the state for context:

https://gist.githubusercontent.com/dak1n1/3ec5fc36cb9c4ab16d7d5ed8c159009c/raw/bc670e3486d17d5eb7eafe8016ebb1e63d09d06f/gistfile1.txt

However, if I remove these at the PodSpec level, it omits the blocks entirely, which prevents the diff. I'm going to explore that route next.

dak1n1:
Got it working! Pods are no longer destroyed on each apply.

```
[dakini@dax pod_with_volume]$ terraform show 
[...]
            volume_mount {
                mount_path        = "/etc/test"
                mount_propagation = "None"
                name              = "pvc"
                read_only         = false
            }
        }

        volume {
            name = "pvc"

            persistent_volume_claim {
                claim_name = "test"
                read_only  = false
            }
        }
    }
}
[dakini@dax pod_with_volume]$ terraform plan
kubernetes_persistent_volume_claim.test: Refreshing state... [id=default/test]
kubernetes_pod.main: Refreshing state... [id=default/test]

No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, no
actions need to be performed.
```

kubernetes/structures_pod.go
```diff
@@ -107,6 +108,17 @@ func flattenPodSpec(in v1.PodSpec) ([]interface{}, error) {
 	}
 
 	if len(in.Volumes) > 0 {
+		for i, volume := range in.Volumes {
+			// To avoid perpetual diff, remove the default service account token volume from PodSpec.
+			nameMatchesDefaultToken, err := regexp.MatchString("default-token-([a-z0-9]{5})", volume.Name)
```
jrhouston, line 113: Ah, I just realized we are going to have to be a bit more clever than this. 

When the token get's auto-mounted it takes the name from the service account, so only if they don't supply a service account will it be called `default`, otherwise it will be the name of the service account. You can see how it does this in the admission controller code for it: [kubernetes/blob/master/plugin/pkg/admission/serviceaccount/admission.go#L46](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/serviceaccount/admission.go#L468)

So we'll have to actually pass down the serviceAccount name and make the regex match that. 😢 

We could just match on `-token-` but that creates a surprising edge case for someone to discover.

dak1n1: oh, nice! That is really helpful. I hadn't thought to test with a service account specified. I'll put in a test to catch these cases and make it clever enough to handle them. Thanks!

dak1n1:
It works now. I'll put the test into PR #1074, since that contains the test for bug #1085

kubernetes/structures_pod.go
```diff
@@ -107,6 +116,17 @@ func flattenPodSpec(in v1.PodSpec) ([]interface{}, error) {
...
 
 	if len(in.Volumes) > 0 {
+		for i, volume := range in.Volumes {
+			// To avoid perpetual diff, remove the service account token volume from PodSpec.
+			nameMatchesDefaultToken, err := regexp.MatchString(serviceAccountRegex, volume.Name)
+			if err != nil {
+				return []interface{}{att}, err
+			}
+			if nameMatchesDefaultToken {
+				in.Volumes = removeVolumeFromPodSpec(i, in.Volumes)
```
jrhouston, line 126: At this point you can `break` the loop because you've found the index of the volume and so don't need to keep looping.

kubernetes/structures_container.go
```diff
@@ -404,6 +406,17 @@ func flattenContainers(in []v1.Container) ([]interface{}, error) {
...
 
 		if len(v.VolumeMounts) > 0 {
+			for num, m := range v.VolumeMounts {
+				// To avoid perpetual diff, remove the default service account token volume from the container's list of volumeMounts.
+				nameMatchesDefaultToken, err := regexp.MatchString(serviceAccountRegex, m.Name)
+				if err != nil {
+					return att, err
+				}
+				if nameMatchesDefaultToken {
+					v.VolumeMounts = removeVolumeMountFromContainer(num, v.VolumeMounts)
```
jrhouston, line 416: Can also just `break` here.

jrhouston (changes requested):
Looks good just a couple of nits.

jrhouston (approved):
🚀

PR #1112, opened 2021-01-10
feat: add option for provider specific proxy

hrvolapeter:
### Description
Adds `proxy_url` option to provider settings which allows to use providers specific proxy.

[Example](https://github.com/jenkins-x/terraform-google-jx/pull/170/commits/647decad45fb192d70295bdcc5bd3289938f0aae#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR66) usage, note how this can be combined with external provider for form dependency on bastion host

Use case:
Setting up private GKE and provisioning kubernetes resources after cluster is initialized

Alternative (current) solutions:
- use `HTTPS_PROXY` which is then used by all providers and if bastion host is created by google provider it cannot be used to bootstrap the project and would require two runs (once without HTTPS proxy which will fail kubernetes resources but will create bastion, second with bastion as HTTP_PROXY)
- separate  terraform to multiple independent stages, first provision GKE infrastructure and subsequently run kubernetes specific resources. Downside is that this requires manual step and complicated usage of provided modules. E.g using this approach with https://github.com/jenkins-x/terraform-google-jx would force users to manually apply multi stage approach even when not using private cluster 

Implementation:
Expose setting of proxy which is currently only set through env variables.

### Acceptance tests

```

...
```

### Release Note

```release-note
Add option `proxy_url` for provider specific proxy. Supports http(s), socks5 proxy schemes
```

### References

Example usage https://github.com/jenkins-x/terraform-google-jx/pull/170/commits/647decad45fb192d70295bdcc5bd3289938f0aae#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR66

hashicorp/terraform-provider-helm#654
jenkins-x/terraform-google-jx#170

### Community Note

### DIFF
--- a/kubernetes/test-infra/eks/kubernetes-config/main.tf
+++ b/kubernetes/test-infra/eks/kubernetes-config/main.tf
@@ -1,21 +1,13 @@
-resource "kubernetes_config_map" "name" {
-  metadata {
-    name      = "aws-auth"
-    namespace = "kube-system"
-  }
-
-  data = {
-    mapRoles = join(
-      "\n",
-      formatlist(local.mapped_role_format, var.k8s_node_role_arn),
-    )
-  }
-}
-
-# Optional: this kubeconfig file is only used for manual CLI access to the cluster.
-resource "null_resource" "generate-kubeconfig" {
-  provisioner "local-exec" {
-    command = "aws eks update-kubeconfig --name ${var.cluster_name} --kubeconfig ${path.root}/kubeconfig"
+terraform {
+  required_providers {
+    kubernetes = {
+      source = "localhost/test/kubernetes"
+      version = "9.9.9"
+    }
+    helm = {
+      source  = "localhost/test/helm"
+      version = "9.9.9"
+    }
   }
 }
 
--- a/kubernetes/test-infra/eks/main.tf
+++ b/kubernetes/test-infra/eks/main.tf
@@ -1,8 +1,18 @@
 terraform {
   required_providers {
-    kubernetes = {
+    # This is the locally compiled version of the provider, based on the current branch.
+    kubernetes-local = {
       source = "localhost/test/kubernetes"
       version = "9.9.9"
+    }
+    # The following block configures the latest released version of the provider, which is needed for the EKS cluster module.
+    # This configuration is a work-around, because required_providers blocks are not inherited by sub-modules.
+    # A "required_providers" block needs to be added to all sub-modules in order to use a custom "source" and "version".
+    # Otherwise, the sub-module will use defaults, which in our case means an empty provider config.
+    # https://github.com/hashicorp/terraform/issues/27361
+    kubernetes-released = {
+      source = "hashicorp/kubernetes"
+      version = ">= 2.0.2"
     }
     helm = {
       source  = "localhost/test/helm"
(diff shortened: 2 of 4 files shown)


jrhouston:
I think it makes sense to expose this option explicitly via the provider block since there is a `ProxyURL` attribute in the client config. I wouldn't call it bastion_host though, I think `proxy_url` is probably a more accurate name.

hrvolapeter:
@jrhouston  I've changed the naming per your suggestion

aareet:
Similar PR for Helm - https://github.com/hashicorp/terraform-provider-helm/pull/654

aareet:
@hrvolapeter could you please also add documentation for this change?

hrvolapeter:
@aareet I've added description to index for option also for the other PR

aareet:
Thanks @hrvolapeter - we're cutting 2.0 of the provider soon so master is frozen, but will take a look at this after that release is cut

hrvolapeter:
@aareet Looks lie v2.0 is out, are we ready for next steps?

aareet:
@hrvolapeter yes, it's now in our queue for review.

I believe https://github.com/hashicorp/terraform-provider-helm/pull/574 may be related to this as well.

davidmitchell2019:
Release the beast!!!!

redeux:
Hi everyone. I want to provide a quick status update, because I know there are some folks that are anxious to get this merged.  We're currently investigating whether this is best implemented at the provider or sdk level. We'll circle back with an update on this PR when that evaluation is done. Thank you for your continued patience while we evaluate the best long term solution to this problem.

hrvolapeter:
Hi @redeux, could you describe the alternative approach that you're considering?

WillieWookiee:
@redeux I think the problem with the SDK level if you are referring to adding the `https_proxy` env var is that this breaks the gcloud command and any other SDK level commands that need to reach out to the Internet. There is a use case where some people are building a proxy that only allows access to the K8s cluster. This means if we define it at the SDK level, it breaks everything else. 

I might be off with my assumption for what you are proposing, but wanted to put that out there.

This is how Google recommends setting up the proxy for just getting to a private cluster using privoxy.
[GKE proxy with Privoxy](https://cloud.google.com/solutions/creating-kubernetes-engine-private-clusters-with-net-proxies)

redeux:
@hrvolapeter @WillieWookiee The alternative we're exploring is proposed here: hashicorp/terraform-plugin-go-contrib/issues/4

mutru:
I'd really love to see this merged. My use case is the same as the original author's, wanting to setup a private GKE cluster and using a proxy to connect to Kubernetes master. I don't want to use that proxy for anything else, so I don't want to setup a system-wide HTTPS proxy.

I'm not an expert on Terraform internals, but this use case requires configuring the proxy at the provider level in any case, and the implementation is a very simple pass-through of the value. Thus, I would think that getting this PR in would only take us to the right direction.

ajohnstone:
proxy-url is supported natively in newer go-clients and pretty well hidden, albeit you have to patch it into your ~/.kube/config.
presently it seems k8s.io/client-go v0.21.0 is used, so seems to be supported already..

https://github.com/kubernetes/client-go/blob/d412730e5f0160f6dc0a83459c14b05df8ea56fb/pkg/apis/clientauthentication/v1beta1/types.go#L95-L98

With that said much better to have natively supported in terraform.


```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: ...
    server: ...
    proxy-url: http://abc.com:443/
  name: cluster-a
```

jrhouston:
Accidentally pushed the wrong branch to this PR which closed it, but now I can't reopen the PR from the branch. I'm just going to reopen this from [a branch](https://github.com/hashicorp/terraform-provider-kubernetes/tree/proxy-url) inside the main repo – I will set you as the author @hrvolapeter.

hrvolapeter:
The suggested changes have been merged now as part of the above PR and in the next release there should be a support for proxy_url option

PR #1185, opened 2021-03-03
Add kubernetes_config data resource

kirecek:
### Description

Hi, this PR adds a new data resource that allows to pull KUBECONFIG data .. similarly as [google_client_config](https://github.com/hashicorp/terraform-provider-google/blob/master/google/data_source_google_client_config.go) for example. The use-case is to be able to provide auth data to kubectl command wrappers or other terraform modules that are sometimes used in cases where the provider is not sufficient.

I agree that the use-cases for this PR might not be the best practices tho but .... you know. I wonder if this makes sense for other people

### Acceptance tests

**WIP** (not sure yet how to exactly pull test values).

```

...
```

### Release Note

```release-note
...
```

### DIFF
--- a/kubernetes/data_source_kubernetes_context.go
+++ b/kubernetes/data_source_kubernetes_context.go
@@ -1,0 +1,86 @@
+package kubernetes
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceKubernetesConfig() *schema.Resource {
+
+	return &schema.Resource{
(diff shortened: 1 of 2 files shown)


dak1n1:
Would this module cover your use case instead? https://registry.terraform.io/modules/redeux/kubeconfig/kubernetes/latest

If not, could you post a configuration sample that demonstrates the use case? It might be possible to pass the same authentication values into the other providers, such as the kubeconfig or token.

kirecek:
yup, it's solvable with other providers. The one you mentioned would also work, but we ended up using [kubectl](https://registry.terraform.io/providers/gavinbunney/kubectl/latest) provider. 

Thanks, closing the MR (sorry for not updating it).

PR #1210, opened 2021-03-22
Fix grammar issue in Network Policy documentation

michielkempen:
### Description

Fix grammar issue in Network Policy documentation


theqwertyosc (approved):
Is better. Also looks like build failed because of a quota, can we give it a retry?

dak1n1 (approved):
Thanks for this fix!

PR #1227, opened 2021-04-12
Fix perpetual diff in claim_ref

dak1n1:
The `claim_ref` attribute would perpetually show a diff if the PV was created without a `claim_ref` specified.
This is because all PVs gain a claimRef once a PVC binds to it. (This creates the bi-directional link between PV and PVC,
and it's done on the Kubernetes API side).

To support this, `claim_ref` is now a Computed attribute, and it will update to show the claimRef once a PVC has bound to the PV,
(only now it will do so without creating a diff in terraform).

Also added extra tests to catch this case and moved claim_ref's flatteners/expanders/tests into the correct files.

### Acceptance tests

```
go test -run TestFlattenObjectRef '-timeout=30s' '-parallel=4' /home/dakini/go/src/github.com/hashicorp/terraform-provider-kubernetes/kubernetes 
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   0.036s

=== RUN   TestAccKubernetesPersistentVolume_minimal
--- PASS: TestAccKubernetesPersistentVolume_minimal (2.53s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   2.578s


=== RUN   TestAccKubernetesPersistentVolume_hostpath_claimRef
--- PASS: TestAccKubernetesPersistentVolume_hostpath_claimRef (13.56s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   13.604s
```

### Release Note

```release-note
Fix perpetual diff in claim_ref
```

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_persistent_volume.go
+++ b/kubernetes/resource_kubernetes_persistent_volume.go
@@ -118,14 +118,16 @@ func resourceKubernetesPersistentVolume() *schema.Resource {
 							Type:        schema.TypeList,
 							Description: "A reference to the persistent volume claim details for statically managed PVs. More Info: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#binding",
 							Optional:    true,
+							Computed:    true,
 							MaxItems:    1,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"namespace": {
 										Type:        schema.TypeString,
-										Description: "The namespace of the PersistentVolumeClaim",
+										Description: "The namespace of the PersistentVolumeClaim. Uses 'default' namespace if none is specified.",
 										Elem:        schema.TypeString,
-										Required:    true,
+										Optional:    true,
+										Default:     "default",
 									},
 									"name": {
 										Type:        schema.TypeString,
--- a/kubernetes/resource_kubernetes_persistent_volume_claim_test.go
+++ b/kubernetes/resource_kubernetes_persistent_volume_claim_test.go
@@ -716,11 +716,6 @@ resource "kubernetes_persistent_volume" "test" {
       }
     }
   }
-  lifecycle {
-    ignore_changes = [
-      spec[0].claim_ref,
-    ]
-  }
 }
 resource "kubernetes_persistent_volume_claim" "test" {
   wait_until_bound = true
--- a/kubernetes/resource_kubernetes_persistent_volume_test.go
+++ b/kubernetes/resource_kubernetes_persistent_volume_test.go
@@ -16,6 +16,29 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+func TestAccKubernetesPersistentVolume_minimal(t *testing.T) {
+	var conf api.PersistentVolume
+	randString := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+	name := fmt.Sprintf("tf-acc-test-%s", randString)
+
+	const resourceName = "kubernetes_persistent_volume.test"
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		IDRefreshName:     resourceName,
+		ProviderFactories: testAccProviderFactories,
+		CheckDestroy:      testAccCheckKubernetesPersistentVolumeDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccKubernetesPersistentVolumeConfig_hostPath_basic(name),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckKubernetesPersistentVolumeExists(resourceName, &conf),
+					resource.TestCheckResourceAttr(resourceName, "spec.0.claim_ref.#", "0"),
+				),
+			},
+		},
+	})
+}
+
 func TestAccKubernetesPersistentVolume_azure_basic(t *testing.T) {
 	var conf api.PersistentVolume
 	randString := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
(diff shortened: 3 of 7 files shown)


kubernetes/resource_kubernetes_persistent_volume.go
```diff
@@ -118,14 +118,16 @@ func resourceKubernetesPersistentVolume() *schema.Resource {
...
 								Schema: map[string]*schema.Schema{
 									"namespace": {
 										Type:        schema.TypeString,
-										Description: "The namespace of the PersistentVolumeClaim",
+										Description: "The namespace of the PersistentVolumeClaim. Uses 'default' namespace if none is specified.",
 										Elem:        schema.TypeString,
-										Required:    true,
+										Optional:    true,
+										Default:     "default",
 									},
```
dak1n1, line 131: Since all PVCs are namespaced, I figured we should put some guardrails in here to catch cases where users accidentally leave it blank. In my testing, the Kubernetes API allows the ClaimRef to be blank, but it will never bind until the correct namespace is specified.

kubernetes/resource_kubernetes_persistent_volume_claim_test.go
```diff
@@ -716,11 +716,6 @@ resource "kubernetes_persistent_volume" "test" {
       }
     }
   }
-  lifecycle {
-    ignore_changes = [
-      spec[0].claim_ref,
-    ]
-  }
 }
```
dak1n1, line 719: This section was causing the test to pass, when it shouldn't have. It masked the perpetual diff.

kubernetes/structure_persistent_volume_spec.go
```diff
@@ -1025,11 +1036,11 @@ func expandPersistentVolumeSpec(l []interface{}) (*v1.PersistentVolumeSpec, erro
...
-func expandClaimRef(v []interface{}) *v1.ObjectReference {
-	if len(v) == 0 || v[0] == nil {
-		return nil
+func expandClaimRef(l []interface{}) *v1.ObjectReference {
+	if len(l) == 0 || l[0] == nil {
+		return &v1.ObjectReference{}
 	}
-	o := v[0].(map[string]interface{})
+	o := l[0].(map[string]interface{})
 	return &v1.ObjectReference{
```
dak1n1, line 1044: I just changed the variable name here to be consistent with the rest of the file.

kubernetes/structures_container.go
```diff
@@ -1028,14 +1028,3 @@ func expandContainerResourceRequirements(l []interface{}) (*v1.ResourceRequireme
...
-func flattenObjectRef(in *v1.ObjectReference) []interface{} {
-	att := make(map[string]interface{})
-	if in.Name != "" {
-		att["name"] = in.Name
-	}
-	if in.Namespace != "" {
-		att["namespace"] = in.Namespace
-	}
-	return []interface{}{att}
-}
```
dak1n1, line 1041: This was moved because it's only used in the volume schema.

kubernetes/resource_kubernetes_persistent_volume_test.go
```diff
@@ -16,6 +16,29 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+func TestAccKubernetesPersistentVolume_minimal(t *testing.T) {
+	var conf api.PersistentVolume
```
dak1n1, line 20: `_minimal` tests are a type I started adding a few months back. The goal with these is just to provision the bare minimum amount of infrastructure, with as few options set as possible, which should work in local testing and on cloud environments. These kinds of tests will tell you if the basic resource provisioning is working properly, and if the default values for the resource are working properly.

kubernetes/resource_kubernetes_persistent_volume.go
alexsomesan: I think that's a good reasoning.

I wasn't sure if namespace is actually required in the Kubernetes schema or not, so I checked in the schema reference docs. It's indeed not required. However, I noticed that `claimRef` is actually defined as type `ObjectReference`. That type has a bunch of other attributes that are missing in our schema (were also missing in the original PR).
I wonder if we should take this opportunity to backfill those as well so that `claimRef` actually follows the official API schema.
I also missed them when reviewing the original PR.

Complete set of attributes of claimRef described here: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#objectreference-v1-core

kubernetes/structures_container.go
alexsomesan: Technically `v1.ObjectReference` is used in a bunch of other resources (listed here: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#objectreference-v1-core).

We might want to keep this and think about consolidating all those use-cases to re-use the same schema snippet and flatten / expand code (not in this PR, of course).

alexsomesan (commented):
Thanks for the nice resolution here, @dak1n1 !

I left some thoughts here and there, but most importantly I realised that the current schema for `claimRef` seems to be incomplete.

I think this might be a good opportunity to fix that too, before users start pointing it out.

kubernetes/structures_container.go
dak1n1: Yeah, I did see that in the API docs, but in our codebase it's only used for the persistent volume schema, so I thought it made sense to keep it in that file until it's used more broadly. We may never use it outside of the persistent volume schema. But even if we do, the containers schema isn't quite the right place for it either, since it wouldn't be used in containers.

kubernetes/resource_kubernetes_persistent_volume.go
dak1n1: Yeah, I noticed all those extra attributes coming up inside the PV during testing, and in the API docs. However, we only need name and namespace for this feature. If there is a need for the other attributes later, like if users need to use an  ObjectReference outside of a PV, I think it makes sense to add these things as people ask for it. Then we can evaluate the usefulness of each feature that users are asking for, and prioritize accordingly. So far it seems we've implemented all the useful types without creating an ObjectReference (we have Secret, CronJob, ServiceAccount). I imagine it would involve some refactoring to go back and change how those work.

alexsomesan: Sounds good then. I'll document the missing ObjectReference fields in a separate issue and we can get back to them when there is demand.

PR #1257, opened 2021-05-03
Proposal: kubernetes_patch resource

jrhouston:
### Description

This is a proposal to solicit feedback on adding a `kubernetes_patch` resource to the provider. 

### References

https://github.com/hashicorp/terraform-provider-kubernetes/issues/723

### Community Note

### DIFF
--- a/proposals/001_resource_kubernetes_patch.md
+++ b/proposals/001_resource_kubernetes_patch.md
@@ -1,0 +1,165 @@
+# Adding a `kubernetes_patch` resource 
+
+This proposal discusses adding a `kubernetes_patch` resource to the provider. 
+
+
+## Some Background
+
+[kubectl patch](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/) allows the user to mutate an existing API resource by specifying its name and kind, and a JSON or YAML partial with the desired change, e.g:
+
+```
+kubectl patch node k8s-node-1 -p '{"spec":{"unschedulable":true}}
+```
+
+There are 3 types of patch that can be applied:
+
+1. **strategic** – This is the default and does a Strategic Merge Patch on the resource. This means that lists in the resource will be merged with lists in the patch, combining the items. This lets you add things to lists without knowing the current contents. This type is not supported on custom resources.
+
+
+2. **merge** – This patch type causes your patch to replace lists entirely, which means if you want to add items to a list, the list in your patch needs to contain the current contents. This is the more blunt “clobber everything” strategy.
+
+
+3. **json** – The patch type allows you to do a [JSON patch](http://jsonpatch.com/) using a list of positional operations. This lets you specify a list of operations, i.e add/replace/remove, the JSON path a specific field, and the new value, e.g:
+
+```json
+[
+   {
+       "op": "replace",
+       "path": "/spec/containers/0/image",
+       "value":"newimage"
+   }
+]
+```
+
+Technically, there is a 4th kind of patch supported by the API – `apply-patch+yaml` which is used to do server side apply although this is not exposed via the `--type` option of the `kubectl patch` command. 
+
+
+# Use Cases
+
+The [feature request](https://github.com/hashicorp/terraform-provider-kubernetes/issues/723) for this has 150+ 👍 now. A lot of people want this for a variety of reasons:
+
+- Updating resources that come baked into AWS EKS so tolerations can be used
+- Removing annotations from resources so they can be used with AWS Fargate
+- Patching resources to be used with service meshes
+- Patching ConfigMaps to be used in projects like ArgoCD
+- Modifying the default namespace/service account on a cluster without importing / recreating
+
+Previously we viewed this functionality as more of a "carry out this operation with `kubectl` before running terraform" thing, but given the mounting list of use cases and user appetite I think this fits our workflows based approach and we should try and accommodate.
+
+
+# Proposed Configuration
+
+A generic `kubernetes_patch` resource should allow you specify:
+
+- The kind of resource being patched
+- The resource's metadata (i.e name and namespace) 
+- The type of patch operation 
+- The patch itself 
+
+In the kubernetes provider, it would look like this. 
+
+```hcl
+resource "kubernetes_patch" {
+  kind = "namespace"
+
+  metadata {
+    name = "default"
+  }
+ 
+  patch = jsonencode({
+    metadata = {
+      labels = {
+        test = "test"
+      }
+    }
+  })
+ 
+  type = "merge"
+}
+```
+
+In the kubernetes-alpha provider:
+
+```hcl
+resource "kubernetes_patch" {
+  provider = kubernetes-alpha
+
+  kind = "namespace"
+
+  metadata {
+    name = "default"
+  }
+ 
+  patch = {
+    metadata = {
+      labels = {
+        test = "test"
+      }
+    }
+  }
+ 
+  type = "merge"
+}
+```
+
+## Possible Alternatives
+
+A possible alternative to the generic patch resource above is to implement use-case specific resources for managing things like annotations, tolerations, and so on. 
+
+For example:
+
+```hcl
+resource "kubernetes_annotations" {
+  kind = "pod"
+
+  metadata {
+    name = "pod-name"
+  }  
+
+  annotations = {
+    "test-annotation" = "value"
+  }
+}
+```
+
+The big downside of this is increased maintenance burden – we will have to implement a new resource for every use case that merits a resource being patched.  
+
+
+## Discussion
+
+
+1. **Should we implement this in the kubernetes provider, or in the kubernetes-alpha provider?**
+
+Arguments _for_ kubernetes provider:
+- Stability – users of the main provider won’t need to add kubernetes-alpha which still in flux. 
+- Simple patch doesn’t have as many exotic challenges as custom resources.
+- Easier implementation than in the alpha provider because it uses the main Terraform SDK. 
+
+Arguments _against_ kubernetes provider:
+- The patch has to be stored as a text string and used with `jsondecode`/`jsonencode`.
+- We will have to do JSON munging to do any validation on the patch.
+
+
+
+Arguments _for_ kubernetes-alpha provider:
+- Cleaner syntax, no need for `jsonencode()`.
+- Adding a highly wanted feature that incentivizes people to adopt the kubernetes-alpha provider.
+- Code already exists in kubernetes-alpha to do things like validate against the OpenAPI schema at plan time.
+
+Arguments _against_ kubernetes-alpha provider:
+- Scope of the alpha provider becomes less tightly circumscribed around custom resources
+- More conservative users will have to wait for the alpha provider to hit GA 
+
+
+2. **Should we support the JSON patch type of just strategic and merge?**
+
+The k8s documentation actually doesn’t illustrate the JSON patch type – it’s only there in the `--help` information.
+
+You can do everything with “merge” and “strategic” that json patch can do (I think). None of the commenters in the issue seem to have asked for this and the examples were strategic patches.
+
+
+3. **Should the resource be a “logical resource” that is applied once and then considered created forever, or should we read the resource to see if the patch needs to be applied again?**
+
+"Patching" is something that doesn’t quite fit with the CRUD model of Terraform as it is a stateless one time operation. We could make this a resource that is simply created and has no read function, like we did with CertificateSigningRequest.
+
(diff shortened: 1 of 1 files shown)


proposals/001_resource_kubernetes_patch.md
```diff
@@ -0,0 +1,165 @@
...
+The k8s documentation actually doesn’t illustrate the JSON patch type – it’s only there in the `--help` information.
+
+You can do everything with “merge” and “strategic” that json patch can do (I think). None of the commenters in the issue seem to have asked for this and the examples were strategic patches.
+
+
+3. **Should the resource be a “logical resource” that is applied once and then considered created forever, or should we read the resource to see if the patch needs to be applied again?**
+
+"Patching" is something that doesn’t quite fit with the CRUD model of Terraform as it is a stateless one time operation. We could make this a resource that is simply created and has no read function, like we did with CertificateSigningRequest.
+
+We could also use a custom diff to do a dry-run to check if the patch is actually going to do anything and then do a force-replace on it if there is.
```
RichiCoder1, line 165: Either diff or maybe something like `null_resource`'s `triggers` if that's too complicated.

petewilcock:
- +1 for generic `kubernetes_patch` command that lets you specify type. I dislike `jsonencoding` myself and wonder if it's too ambitious to have patch statements take a structured format as per [IAM policy documents](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document)? Feels like validation logic could be reused there?
- I'd favour kubernetes provider for stability.
- JSON positional patch is an interest feature that I feel would have a viable use for many and avoid verbosity - maybe few people (including me) realised this was a thing. 
- I suspect that re-reading the underlying resource would be necessary but I have no idea what metric would be used to determine idempotence. Perhaps this would also be optionally one-off, and if not would need to be dependent on another resource created in the provider to be re-applied reliably?

proposals/001_resource_kubernetes_patch.md
```diff
@@ -0,0 +1,165 @@
...
+
+[kubectl patch](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/) allows the user to mutate an existing API resource by specifying its name and kind, and a JSON or YAML partial with the desired change, e.g:
+
+```
+kubectl patch node k8s-node-1 -p '{"spec":{"unschedulable":true}}
+```
+
+There are 3 types of patch that can be applied:
+
+1. **strategic** – This is the default and does a Strategic Merge Patch on the resource. This means that lists in the resource will be merged with lists in the patch, combining the items. This lets you add things to lists without knowing the current contents. This type is not supported on custom resources.
```
dak1n1, line 16: It might be more accurate to represent the Strategic patch as a more dynamic patch type that looks at the specific object to determine how to modify it. The behavior of the strategic patch is determined per-resource in the Kubernetes API. It will sometimes behave as 1. describes here, but sometimes it will differ.

Their docs describe it in an example [here](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch) 

```
the list in the patch was merged with the existing list. This is not always what happens when you use a strategic merge patch on a list. In some cases, the list is replaced, not merged.

With a strategic merge patch, a list is either replaced or merged depending on its patch strategy
```

The API docs also contain a reference to this:

```
When patching complex types, arrays and maps, how the patch is applied is defined on a per-field basis and may either replace the field's current value, or merge the contents into the current value.
```

Here are some instances of the patchStrategy being used in the Kubernetes code:
https://github.com/kubernetes/kubernetes/search?q=patchStrategy&type=code

proposals/001_resource_kubernetes_patch.md
```diff
@@ -0,0 +1,165 @@
...
+Arguments _against_ kubernetes-alpha provider:
+- Scope of the alpha provider becomes less tightly circumscribed around custom resources
+- More conservative users will have to wait for the alpha provider to hit GA 
+
+
+2. **Should we support the JSON patch type of just strategic and merge?**
+
+The k8s documentation actually doesn’t illustrate the JSON patch type – it’s only there in the `--help` information.
+
+You can do everything with “merge” and “strategic” that json patch can do (I think). None of the commenters in the issue seem to have asked for this and the examples were strategic patches.
```
dak1n1, line 158: I think you're right. The two options could be described as "replace" or "add to lists when possible". If the resource contains an immutable field, it wouldn't matter if we were to use a JSON patch to tell it to "add"... because the field is immutable. However a JSON patch would give us the ability to do a `remove`. I'm not sure if that's possible using strategic or merge. You could "replace" a list with an empty list... but historically we've had trouble representing zero-value objects in the provider. (Such as specifying "" or {}).

proposals/001_resource_kubernetes_patch.md
```diff
@@ -0,0 +1,165 @@
...
+
+
+3. **Should the resource be a “logical resource” that is applied once and then considered created forever, or should we read the resource to see if the patch needs to be applied again?**
```
dak1n1, line 161: I think reading the resource would be a better user experience than a one-time create. A logical resource would only make sense when you're creating an object that doesn't need to exist on the API server (such as the ephemeral CertificateSigningRequest or generating a TLS key locally and storing the result in state).

But for a patch resource, users would expect that their changes remain in place (on the server) even after the initial terraform apply, because terraform tracks the changes of all the other things they're applying.

This would help us to maintain the CRUD model a bit better. We would just be applying CRUD to a part of the resource, but at least that part of the resource would fit into the Terraform model of declarative configuration.

proposals/001_resource_kubernetes_patch.md
dak1n1, line 165: Could we do the dry run during the Read operation? That would also run after the initial Create, which would immediately tell the user if there is a problem with idempotency. I think it would either return an "inconsistent final plan" or maybe a perpetual diff, in the case of an invalid patch that updates on each run but doesn't actually succeed in modifying the object.

proposals/001_resource_kubernetes_patch.md
```diff
@@ -0,0 +1,165 @@
...
+[
+   {
+       "op": "replace",
+       "path": "/spec/containers/0/image",
+       "value":"newimage"
+   }
+]
+```
+
+Technically, there is a 4th kind of patch supported by the API – `apply-patch+yaml` which is used to do server side apply although this is not exposed via the `--type` option of the `kubectl patch` command.
```
alexsomesan, line 34: This one is accessible in kubectl via `kubectl apply --server-side`

proposals/001_resource_kubernetes_patch.md
```diff
@@ -0,0 +1,165 @@
...
+
+
+
+Arguments _for_ kubernetes-alpha provider:
+- Cleaner syntax, no need for `jsonencode()`.
+- Adding a highly wanted feature that incentivizes people to adopt the kubernetes-alpha provider.
+- Code already exists in kubernetes-alpha to do things like validate against the OpenAPI schema at plan time.
+
+Arguments _against_ kubernetes-alpha provider:
+- Scope of the alpha provider becomes less tightly circumscribed around custom resources
```
alexsomesan, line 150: The `kubernetes_manifest` resource, being backed by server-side apply, already has the ability to "patch" existing resources, albeit as a side-effect since this wasn't a design goal. Applying a manifest when its resource already exists in the cluster results in modifying the resource on the API side, but the resulting state is unreliable.
One could make the case that implementation in the alpha provider should just add the missing pieces to make this work reliably.

OTH, since the alpha provider doesn't use the high-level SDK with its per-resource abstractions, adding a second resource would result in significant code re-shuffling. We need to keep this in mind given the tight milestones to beta and GA.

proposals/001_resource_kubernetes_patch.md
```diff
@@ -0,0 +1,165 @@
...
+- More conservative users will have to wait for the alpha provider to hit GA 
+
+
+2. **Should we support the JSON patch type of just strategic and merge?**
```
alexsomesan, line 154: JSON patch seems more deterministic. Also, seems to be the only can describe differential operations, allowing to distinguish between intentions like: "add / remove THIS element to THAT list" and "replace THIS list with THAT list" (same for maps).

proposals/001_resource_kubernetes_patch.md
alexsomesan, line 165: What kind of state would the patch resource persist?

jrhouston:
> * +1 for generic `kubernetes_patch` command that lets you specify type. I dislike `jsonencoding` myself and wonder if it's too ambitious to have patch statements take a structured format as per [IAM policy documents](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document)? Feels like validation logic could be reused there?
> * I'd favour kubernetes provider for stability.
> * JSON positional patch is an interest feature that I feel would have a viable use for many and avoid verbosity - maybe few people (including me) realised this was a thing.
> * I suspect that re-reading the underlying resource would be necessary but I have no idea what metric would be used to determine idempotence. Perhaps this would also be optionally one-off, and if not would need to be dependent on another resource created in the provider to be re-applied reliably?

Thanks for adding your thoughts @petewilcock. Do you have any specific use cases you can share that you would implement with this feature?

I agree that having to use `jsonencode()` is not ideal. The problem with having a structured format like AWS IAM Policy is that we need to know the schema ahead of time as we can't currently have an untyped object in the kubernetes provider due to a limitation with the Terraform SDK. While it would be possible – albeit tedious and very time consuming – to support the standard k8s API surface area here, it would be impossible to know the schemas for custom resources. So, if we want to use a plain object we would have to put the resource in the kubernetes-alpha project.

proposals/001_resource_kubernetes_patch.md
jrhouston: Thanks for your input @RichiCoder1 – do you have a specific use case you can share for this feature? I'm collecting acceptance test criteria.

petewilcock:
> Do you have any specific use cases you can share that you would implement with this feature?

Being completely honest I feel like many use-cases here are to cover deficits in EKS. Most notably patching add-ons that get automatically bundled but need 'fixing' due to a lack of a configurability on creation. Tolerations and upgrades that are currently very manual being at the top of the list. 

Not to detract from the usefulness of this feature, as I'm sure it would be for a variety of transient cases, but I suspect you'd have seen fewer :+1: on the source issue if the workarounds weren't needed in the first place.

proposals/001_resource_kubernetes_patch.md
RichiCoder1: Mostly patching or updating EKS builtins. This is something that will eventually be fixed by EKS Addons, but if you walk through the EKS upgrade guide you'll see about 3-5 manual patch steps.

RichiCoder1:
There are some cases (like patching default service accounts with annotations) that aren't EKS, but agree that it's 95% because of EKS limitations.

alexmnyc:
https://github.com/hashicorp/terraform-provider-kubernetes/issues/723#issuecomment-832022277

alexmnyc:
https://docs.aws.amazon.com/eks/latest/userguide/cluster-autoscaler.html

Patch the deployment to add the cluster-autoscaler.kubernetes.io/safe-to-evict annotation to the Cluster Autoscaler pods with the following command.

```
kubectl patch deployment cluster-autoscaler \
  -n kube-system \
  -p '{"spec":{"template":{"metadata":{"annotations":{"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"}}}}}'
```

elebertus:
This would be a very helpful change for our team. Right now for patching resources we typically:
- read them out into a `data` type
- update the desired values
- remove the resource
- create the resource
Or:
- Store a template file for the resource
- Render the template
- Wrap a shell script with `local-exec` that runs `kubectl patch` or `kubectl apply`

This is problematic for various reasons, one of them being shelling out with `local-exec` (yes I am aware there are various methods of accomplishing this). Which makes it at best hard to test.

One of our real world examples (mentioned in the proposal) is patching aws-node for CNI configs.

rwkarg:
Adding a GKE use case where using DataPlaneV2 for networking will create a config map in the cluster to configure network logging with logging disabled. To get logging, the already created config map needs to be updated.

This isn't just limited to EKS.

jrhouston:
Having collected feedback here we are going to close this proposal for the moment. We are going to follow up with proposals for resources that are more specifically targeted at the use-cases that have been discussed here.

rabidscorpio:
@jrhouston Can you at least give an explanation as to why this proposal is being thrown out? Is there a reason this can't be implemented? Or why there's a need to provide separate resources for individual use cases?

It seems like providing a single option that everyone can use including for use cases that haven't been brought up yet would be a lot simpler than trying to identify individual use cases.

Issue #872, opened 2020-06-03
Provider behavior differs from kubectl: unable to specify empty resource attribute


dak1n1:
### Terraform Version and Provider Version

Terraform v0.12.26
terraform-provider-kubernetes v1.11.3

### Affected Resource(s)

`kubernetes_persistent_volume_claim` is the most widely reported, but the issue is at the SDK level and therefore affects all resources.

### Terraform Configuration Files

This is one example of trying to specify an empty value for the storageclass.

```hcl
resource "kubernetes_persistent_volume" "example" {
  metadata {
    name = "nfs-volume"
  }
  spec {
    capacity = {
      storage = "1T"
    }
    storage_class_name = ""
    access_modes = ["ReadWriteMany"]
    persistent_volume_source {
      nfs {
        server = google_filestore_instance.test.networks[0].ip_addresses[0]
        path = "/${google_filestore_instance.test.file_shares[0].name}"
      }
    }
  }
}

resource "kubernetes_persistent_volume_claim" "example" {
  metadata {
    name = "mariadb-data"
    namespace = "test"
  }
  spec {
    access_modes = ["ReadWriteMany"]
    storage_class_name = ""
    volume_name = kubernetes_persistent_volume.example.metadata[0].name
    resources {
      requests = {
        storage = "1T"
      }
    }
  }
}
```

### Expected Behavior

Users should be able to specify an empty storageclass (`storage_class = ""`), which is behavior supported by kubectl.

### Actual Behavior

The provider interprets `""` to be the same as omitting the value altogether. [There is no way to differentiate between null and empty values](https://discuss.hashicorp.com/t/differentiate-between-null-and-empty-value/8134) in the current version of the SDK.

### Steps to Reproduce

1. Create a PVC with an empty storage class.
2. `terraform apply`

When `terraform apply` has finished, the storageclass is now set to `default` instead of empty (`""`). This behavior is unexpected because it differs from kubectl, subsequently breaking users' workflows.

### References

The recommended work-around is to create a storageclass and use it in PVs and PVCs, instead of using `""` for the storageclass name. An example of that work-around can be found [here](https://github.com/terraform-providers/terraform-provider-kubernetes/tree/master/_examples/google-gke-nfs-filestore).

Other issues and PRs relating to this will be closed, so we can track them all in this single issue.
* https://github.com/terraform-providers/terraform-provider-kubernetes/pull/590
[...]

PR #1266, opened 2021-05-11
Allow storage_class_name to be set to empty string

derek-burdick:
spec

### Acceptance tests

```

...
```

### Release Note

```release-note
Corrected an issue where storage_class_name could not be set to empty string. Setting storage_class_name to null will map it to the default storage class in the cluster. Setting storage_class_name to empty string will not choose a default storage class.
```

### Community Note

### DIFF
--- a/kubernetes/structure_persistent_volume_claim.go
+++ b/kubernetes/structure_persistent_volume_claim.go
@@ -4,7 +4,8 @@ import (
 	"errors"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 // Flatteners
@@ -38,8 +39,8 @@ func flattenResourceRequirements(in v1.ResourceRequirements) []interface{} {
 
 // Expanders
 
-func expandPersistentVolumeClaim(p map[string]interface{}) (*v1.PersistentVolumeClaim, error) {
-	pvc := &v1.PersistentVolumeClaim{}
+func expandPersistentVolumeClaim(p map[string]interface{}) (*corev1.PersistentVolumeClaim, error) {
+	pvc := &corev1.PersistentVolumeClaim{}
 	if len(p) == 0 {
 		return pvc, nil
 	}
--- a/kubernetes/structure_persistent_volume_claim.go
+++ b/kubernetes/structure_persistent_volume_claim.go
@@ -4,8 +4,7 @@ import (
 	"errors"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"k8s.io/api/core/v1"
-	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


kubernetes/structure_persistent_volume_claim.go
```diff
@@ -4,8 +4,7 @@ import (
 	"errors"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"k8s.io/api/core/v1"
-	corev1 "k8s.io/api/core/v1"
```
alexsomesan, line 8: What's the reasoning for renaming this import alias? 

The new value of `v1` is superfluous because it's also implied by the last element in the import path. That aside, we tend to rename import paths that end with `v1` to avoid conflicts, often with `k8s.io/apimachinery/pkg/apis/meta/v1`. That isn't yet the case here, but should that need to be imported, we're going to have to rename this yet again, so I'd rather keep the `corev1` alias as is.

alexsomesan (commented):
Thanks for this! 
Change looks good overall, but we would prefer if the import path was left as is.

kubernetes/structure_persistent_volume_claim.go
derek-burdick: This was done by a golang auto formatter. We can easily revert it.

derek-burdick: Reverted

dak1n1:
@alexsomesan  There is actually a long history to this particular issue, which we should discuss during the next triage meeting. This has been attempted many times in the past, but this particular solution won't cover the cases we need. And actually it will break certain workflows. So we'll have to solve the larger problem of being able to differentiate between a user-specified value of `""` and omitting the attribute altogether.

Last year, I thought was impossible due to an SDK limitation, but I recently learned that we might actually be able to accomplish this. [Here is the full history of the issue](https://github.com/hashicorp/terraform-provider-kubernetes/issues/872), along with some PRs that are nearly identical to this one.

@aareet would you be able to unlock issue #872 for me? I re-opened it but I can't unlock it. Thanks!

dak1n1:
One work-around you can use is to [specify the name of the cluster's default storage class in your PV configuration](https://gist.githubusercontent.com/dak1n1/6a95719c4e68cfc9e736c8a7624f53a1/raw/939f038b810d314a689bb687ddfbe674aacc69e6/gistfile1.txt). 

I've had to work around this a lot in the past, either by creating a new storage class and using that storage class name, or by using the name of the default storage class. I'm hoping we'll be able to support a blank storage class in the provider in the future, rather than having to work around the issue, now that we know a bit more about how to do this in the SDK.

Here is another example that could help work around the issue in the mean time.  
[Creating a storageclass and using it with google NFS filestore
](https://github.com/hashicorp/terraform-provider-kubernetes/blob/main/_examples/google-gke-nfs-filestore/main.tf#L130-L172)

dak1n1:
We're going to take a different approach to solve this particular issue, which will cover some additional use cases. It will be tracked in https://github.com/hashicorp/terraform-provider-kubernetes/issues/872. Thanks!

PR #1283, opened 2021-05-24
Migrate travis tests to actions

mdeggies:
### Description

Travis-ci.org is being deprecated at the end of this week, so we're moving things to GitHub actions. Tests run using go 1.16 on the ubuntu executor.

The only change is that a new `ci-website-lint` make target has been defined. This is the same as `website-lint`, except it doesn't call the markdown-link-checker bash script. Instead, it calls [native markdown-link-check github action](https://github.com/gaurav-nelson/github-action-markdown-link-check) that was written by the same author.  The ignore and replace patterns have been added to `markdown.links.config.json` as the author recommends.


.github/workflows/test.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@37335c7bb261b353407cff977110895fa0b4f7d8
+      with:
+        go-version: ${{ matrix.go-version }}
+    - name: Login to Docker Hub
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.RO_DOCKERHUB_USER }}
+        password: ${{ secrets.RO_DOCKERHUB_TOKEN }}
```
dak1n1, line 37: Thanks for writing this up for us! It seems to be working as expected. The only part I was confused about is the secrets. I couldn't find them in the org level or the repo level. But I think it's because I just don't have access to them. Is this expected?

dak1n1 (approved):
Thanks for the help! This is great.

.github/workflows/test.yml
```diff
@@ -0,0 +1,50 @@
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+      - migrate-tests-to-actions
```
dak1n1, line 8: I'll just remove this part before merging. It was probably just for testing. Is that right?

```suggestion
```

.github/workflows/test.yml
mdeggies: Ah right, you may need specific permissions to see them under 'secrets', but I added them as repo-level env vars. It uses use read only dockerhub creds associated with a service account.

.github/workflows/test.yml
mdeggies: Thanks!!! Whoops :)

PR #1300, opened 2021-06-08
Github Action for tfproviderlint

dak1n1:
### Description

This sets up a Github Action for tfproviderlint, and addresses the lint failures we currently have, including one for [IDRefresh](https://github.com/bflad/tfproviderlint/tree/main/passes/AT010).

And replacing all the [randString](https://github.com/bflad/tfproviderlint/tree/main/passes/AT009) functions.

As part of testing this, I also updated the AKS example.

### Acceptance tests

tfproviderlint results:

https://github.com/hashicorp/terraform-provider-kubernetes/runs/2779033647?check_suite_focus=true

### Release Note

```release-note
...
```

### Community Note


kubernetes/data_source_kubernetes_config_map.go
```diff
@@ -15,11 +15,13 @@ func dataSourceKubernetesConfigMap() *schema.Resource {
 			"metadata": namespacedMetadataSchema("config_map", false),
 			"data": {
 				Type:        schema.TypeMap,
+				Elem:        &schema.Schema{Type: schema.TypeString},
 				Description: "A map of the config map data.",
```
dak1n1, line 19: To quote [tfproviderlint](https://github.com/bflad/tfproviderlint/tree/main/passes/S006):

```
The S006 analyzer reports cases of TypeMap schemas missing Elem, which currently passes Terraform schema validation, but breaks downstream tools and may be required in the future.
```

I tested this particular instance by creating a configmap resource using a released version of the provider and then doing a `terraform init -upgrade` to the test version of the provider (built from this branch). This schema changes doesn't seem to have any user-visible impact.

.github/workflows/issue-opened.yml
```diff
@@ -3,13 +3,30 @@ name: Issue Opened Triage
...
     steps:
-    - uses: actions/checkout@v2
-    - uses: github/issue-labeler@v2
+    - name: Checkout code
+      uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+    - name: Label new issue
+      uses: github/issue-labeler@829a8bf1b184f74cc575b5435093a92c7b846983
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
         configuration-path: .github/labeler-issue-triage.yml
```
dak1n1, line 30: I'm not sure if the issue labeler will work, since it requires a github token. From what I've observed so far, we don't allow access to any secrets unless you're running from the main branch. The behavior on issues and PRs is unknown. (I don't have access to view or configure those settings, but it seems fine to just see what happens).

.github/workflows/codeql-analysis.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
```
dak1n1, line 33: Choosing a specific commit instead of a tag is part of security hardening for Github Actions. https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions

.github/workflows/issue-opened.yml
```diff
@@ -3,13 +3,30 @@ name: Issue Opened Triage
...
+  actions: none
+  checks: none
+  contents: none
+  deployments: none
+  issues: write
+  packages: none
+  pull-requests: none
+  repository-projects: none
+  security-events: none
+  statuses: none
```
dak1n1, line 18: This is also part of security hardening. 

```
 An action can access the GITHUB_TOKEN through the github.token context even if the 
workflow does not explicitly pass the GITHUB_TOKEN to the action. As a good security practice,
 you should always make sure that actions only have the minimum access they require by 
limiting the permissions granted to the GITHUB_TOKEN. For more information, see 
"Permissions for the GITHUB_TOKEN."
 ```
 
 https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

**Comment by dak1n1** [comment]
I ran some acceptance tests and found that I have a little more work to do.

```
=== RUN   TestAccKubernetesStatefulSet_Update
    resource_kubernetes_stateful_set_test.go:133: Step 5/10 error: Error running pre-apply refresh: exit status 1
        
        Error: invalid value for spec.0.replicas (Must be a positive number)
```

**Comment by dak1n1** [comment]
It's working now:

```
=== RUN   TestAccKubernetesStatefulSet_Update
--- PASS: TestAccKubernetesStatefulSet_Update (257.95s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   258.005s
```

File: kubernetes/schema_pod_spec.go [hunk 5]
```diff
@@ -524,7 +525,7 @@ func volumeSchema(isUpdatable bool) *schema.Resource {
 					Type:         schema.TypeString,
 					Description:  "Target directory name. Must not contain or start with '..'. If '.' is supplied, the volume directory will be the git repository. Otherwise, if specified, the volume will contain the git repository in the subdirectory with the given name.",
 					Optional:     true,
-					ValidateFunc: validateAttributeValueDoesNotContain(".."),
+					ValidateFunc: validation.StringNotInSlice([]string{".."}, true),
```
jrhouston, line 528: This is not quite the same as `validateAttributeValueDoesNotContain`. 

`StringNotInSlice` will only validate that the whole value is not in the given slice, whereas `validateAttributeValueDoesNotContain` is checking that the value does not _contain_ the given string. 

This will fail for `..` but pass on the value `../test` for example.

jrhouston (changes requested):
Couple of things, mostly validation related. 

`validation.IntAtLeast(1)` is used in a bunch of places – we need to check if 0 is also valid in each case. 

`StringNotInSlice` is also used in a bunch of places instead of `validateAttributeValueDoesNotContain` which is not quite right.

_examples/eks/eks-cluster/cluster.tf
```diff
@@ -22,8 +22,8 @@ resource "aws_eks_node_group" "k8s-acc" {
   subnet_ids      = aws_subnet.k8s-acc.*.id
 
   scaling_config {
-    desired_size = 1
-    max_size     = 1
+    desired_size = 3
+    max_size     = 3
```
dak1n1, line 26: I increased this because the acceptance tests couldn't run with so few nodes. It was unable to schedule my pods.

kubernetes/resource_kubernetes_job_test.go
```diff
@@ -189,30 +181,25 @@ func TestAccKubernetesJob_update(t *testing.T) {
...
-//					testAccCheckKubernetesJobExists("kubernetes_job.test", &conf),
-//					resource.TestCheckResourceAttr("kubernetes_job.test", "spec.0.ttl_seconds_after_finished", "60"),
-//				),
-//			},
-//		},
-//	})
-//}
+func TestAccKubernetesJob_ttl_seconds_after_finished(t *testing.T) {
+	var conf api.Job
+	name := fmt.Sprintf("tf-acc-test-%s", acctest.RandString(8))
```
dak1n1, line 186: This is enabled by default on kubernetes version 1.21, so I uncommented it.

kubernetes/schema_container.go
```diff
@@ -650,22 +650,22 @@ func probeSchema() *schema.Resource {
 		Type:         schema.TypeInt,
 		Optional:     true,
 		Default:      10,
-		ValidateFunc: validatePositiveInteger,
+		ValidateFunc: validation.IntAtLeast(1),
```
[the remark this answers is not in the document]
dak1n1: Thanks for catching this! I had lost track of which one was `> 0` vs `>= 0`. I'll get it sorted out.

kubernetes/schema_pod_spec.go
dak1n1: Thanks, I'll check this case and see if can get it fixed.

dak1n1: Done, and I added a test case for it.

kubernetes/resource_kubernetes_pod_disruption_budget.go
```diff
@@ -49,14 +51,14 @@ func resourceKubernetesPodDisruptionBudget() *schema.Resource {
...
-							ValidateFunc: validateTypeStringNullableIntOrPercent,
+							ValidateFunc: validation.StringMatch(regexp.MustCompile(`^([0-9]+|[0-9]+%|)$`), "Must be an absolute number (ex: 5) or a percentage of desired pods (ex: 10%)."),
 						},
 						"min_available": {
 							Type:         schema.TypeString,
 							Description:  podDisruptionBudgetSpecMinAvailableDoc,
 							Optional:     true,
 							ForceNew:     true,
-							ValidateFunc: validateTypeStringNullableIntOrPercent,
+							ValidateFunc: validation.StringMatch(regexp.MustCompile(`^([0-9]+|[0-9]+%|)$`), "Must be an absolute number (ex: 5) or a percentage of desired pods (ex: 10%)."),
```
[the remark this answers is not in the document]
dak1n1: Done. I ended up naming the new function something different. Let me know if I should change it back. I found it more descriptive though.

kubernetes/resource_kubernetes_certificate_signing_request.go
```diff
@@ -128,7 +128,7 @@ func resourceKubernetesCertificateSigningRequestCreate(ctx context.Context, d *s
 			return updateErr
 		})
 		if retryErr != nil {
-			panic(diag.Errorf("CSR auto-approve update failed: %v", retryErr))
+			diag.Errorf("CSR auto-approve update failed: %v", retryErr)
```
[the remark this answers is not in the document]
dak1n1: You're right! I was creating a new diag without returning it. It's fixed now.

kubernetes/schema_container.go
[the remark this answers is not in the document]
dak1n1: This looks correct, actually because validatePositiveInteger just checked to see that the number was at least 1.

Here's the function from the git log:
```
-func validatePositiveInteger(value interface{}, key string) (ws []string, es []error) {
-       v := value.(int)
-       if v <= 0 {
-               es = append(es, fmt.Errorf("%s must be greater than 0", key))
-       }
-       return
-}
```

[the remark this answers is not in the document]
dak1n1: I also double-checked the API docs for this particular field. It says:

```
periodSeconds
integer	How often (in seconds) to perform the probe. Default to 10 seconds. Minimum value is 1.
```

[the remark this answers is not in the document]
dak1n1: I see why you thought it was InitialDelaySeconds though. The diff made it look like I had modified that field, but I didn't. There's no validation for InitialDelaySeconds, since that one is just a plain int.

dak1n1:
I think I've addressed all the feedback.

dak1n1:
I found one more case I need to fix:

```
Error: expected value of spec.0.volume.1.config_map.0.items.0.path to not contain any of "..", got one.txt
```

dak1n1:
This PR should be ready to go. I added a test in here that uncovered a bug, but the bug is happening in the released version of the provider too. https://github.com/hashicorp/terraform-provider-kubernetes/issues/1330

PR #1344, opened 2021-08-03
Import kubernetes_manifest

alexsomesan:
### Description

This change introduces the ability to import `kubernetes_manifest` resources.

### Acceptance tests

```
[import-manifest][~/work/terraform-provider-kubernetes/manifest]$ make testacc TESTARGS="-test.run 'TestKubernetesManifest_Import'"

go test -count=1 -tags acceptance "./test/acceptance" -v -test.run 'TestKubernetesManifest_Import' -timeout 120m
2021/08/03 14:04:49 Testing against Kubernetes API version: v1.21.2
=== RUN   TestKubernetesManifest_Import
--- PASS: TestKubernetesManifest_Import (1.65s)
PASS
ok  	github.com/hashicorp/terraform-provider-kubernetes/manifest/test/acceptance	1.766s
...
```

### Release Note

```release-note
FEATURES:
* Support for importing kubernetes_manifest resources.
```

### Community Note

### DIFF
--- a/manifest/provider/import.go
+++ b/manifest/provider/import.go
@@ -21,6 +21,13 @@ func (s *RawProviderServer) ImportResourceState(ctx context.Context, req *tfprot
 	// Without the user supplying the GRV there is no way to fully identify the resource when making the Get API call to K8s.
 	// Presumably the Kubernetes API machinery already has a standard for expressing such a group. We should look there first.
 	resp := &tfprotov5.ImportResourceStateResponse{}
+
+	execDiag := s.canExecute()
+	if len(execDiag) > 0 {
+		resp.Diagnostics = append(resp.Diagnostics, execDiag...)
+		return resp, nil
+	}
+
 	gvk, name, namespace, err := parseImportID(req.ID)
 	if err != nil {
 		resp.Diagnostics = append(resp.Diagnostics, &tfprotov5.Diagnostic{
--- a/manifest/provider/import.go
+++ b/manifest/provider/import.go
@@ -158,6 +158,11 @@ func (s *RawProviderServer) ImportResourceState(ctx context.Context, req *tfprot
 		TypeName: req.TypeName,
 		State:    &impState,
 	})
+	resp.Diagnostics = append(resp.Diagnostics, &tfprotov5.Diagnostic{
+		Severity: tfprotov5.DiagnosticSeverityWarning,
+		Summary:  "Apply needed after 'import'",
+		Detail:   "Please run apply after a successful import to realign the resource state to the configuration in Terraform.",
+	})
 	return resp, nil
 }
 
@@ -167,26 +172,40 @@ func (s *RawProviderServer) ImportResourceState(ctx context.Context, req *tfprot
 //
 // The expected format for the import resource ID is:
 //
-// "<apiGroup/><apiVersion>#<Kind>#<namespace>#<name>"
+// "apiVersion=<value>,kind=<value>,name=<value>[,namespace=<value>"]
 //
 // where 'namespace' is only required for resources that expect a namespace.
 //
-// Note the '#' separator between the elements of the ID string.
-//
-// Example: "v1#Secret#default#default-token-qgm6s"
+// Example: "apiVersion=v1,kind=Secret,namespace=default,name=default-token-qgm6s"
 //
 func parseImportID(id string) (gvk schema.GroupVersionKind, name string, namespace string, err error) {
-	parts := strings.Split(id, "#")
+	tokens := map[string]string{"apiVersion": "", "kind": "", "name": "", "namespace": "default"}
+	var invalidFormat bool = false
+
+	parts := strings.Split(id, ",")
 	if len(parts) < 3 || len(parts) > 4 {
-		err = fmt.Errorf("invalid format for import ID [%s]", id)
+		invalidFormat = true
+	}
+	for _, p := range parts {
+		t := strings.Split(p, "=")
+		if len(t) != 2 {
+			invalidFormat = true
+			continue
+		}
+		_, ok := tokens[t[0]]
+		if !ok {
+			invalidFormat = true
+			continue
+		}
+		tokens[t[0]] = t[1]
+	}
+	if invalidFormat {
+		err = fmt.Errorf("invalid format for import ID [%s]\nExpected format is: apiVersion=<value>,kind=<value>,name=<value>[,namespace=<value>]", id)
 		return
 	}
-	gvk = schema.FromAPIVersionAndKind(parts[0], parts[1])
-	if len(parts) == 4 {
-		namespace = parts[2]
-		name = parts[3]
-	} else {
-		name = parts[2]
-	}
+	gvk = schema.FromAPIVersionAndKind(tokens["apiVersion"], tokens["kind"])
+	namespace = tokens["namespace"]
+	name = tokens["name"]
+
 	return
 }
--- a/manifest/provider/import.go
+++ b/manifest/provider/import.go
@@ -140,9 +140,15 @@ func (s *RawProviderServer) ImportResourceState(ctx context.Context, req *tfprot
 
 	newState := make(map[string]tftypes.Value)
 	wftype := rt.(tftypes.Object).AttributeTypes["wait_for"]
+	timeoutsType := rt.(tftypes.Object).AttributeTypes["timeouts"]
+	fmType := rt.(tftypes.Object).AttributeTypes["field_manager"]
+
 	newState["manifest"] = tftypes.NewValue(tftypes.Object{AttributeTypes: map[string]tftypes.Type{}}, nil)
 	newState["object"] = morph.UnknownToNull(nobj)
 	newState["wait_for"] = tftypes.NewValue(wftype, nil)
+	newState["timeouts"] = tftypes.NewValue(timeoutsType, nil)
+	newState["field_manager"] = tftypes.NewValue(fmType, nil)
+
 	nsVal := tftypes.NewValue(rt, newState)
 
 	impState, err := tfprotov5.NewDynamicValue(nsVal.Type(), nsVal)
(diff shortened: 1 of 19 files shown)


manifest/provider/import.go
```diff
@@ -0,0 +1,192 @@
...
+	return resp, nil
+}
+
+// parseImportID processes the resource ID string passed by the user to the "terraform import" command
+// and extracts the values for GVK, name and (optionally) namespace of the target resource as required
+// during the import process.
+//
+// The expected format for the import resource ID is:
+//
+// "<apiGroup/><apiVersion>#<Kind>#<namespace>#<name>"
```
jrhouston, line 170: Not sure about the `#` delimiter. We could just use the API path verbatim, then you don't really need to worry about slashes in the api version. 

`v1/namespaces/my-namespace/pods/my-pod`



As an alternative we could use key value pairs to be explicit like:

`apiVersion=v1,kind=Namespace,name=kube-system`

manifest/provider/import.go
```diff
@@ -0,0 +1,192 @@
...
+// where 'namespace' is only required for resources that expect a namespace.
+//
+// Note the '#' separator between the elements of the ID string.
+//
+// Example: "v1#Secret#default#default-token-qgm6s"
+//
+func parseImportID(id string) (gvk schema.GroupVersionKind, name string, namespace string, err error) {
+	parts := strings.Split(id, "#")
+	if len(parts) < 3 || len(parts) > 4 {
+		err = fmt.Errorf("invalid format for import ID [%s]", id)
```
jrhouston, line 181: I would tell the user the expected format in the error message.

jrhouston (commented):
Hmm, so I tried this and the quirk of not being able to access the config at import time definitely makes for a confusing user experience. I wonder if we should generate a warning to let people that this is going to happen, and that it should be ok to run apply again to normalize the state:

```
terraform import kubernetes_manifest.example1 'v1#ConfigMap#default#example1'
kubernetes_manifest.example1: Importing from ID "v1#ConfigMap#default#example1"...
kubernetes_manifest.example1: Import prepared!
  Prepared kubernetes_manifest for import
kubernetes_manifest.example1: Refreshing state...

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.

~/d/h/t/scratch ❯❯❯ terraform plan                                                                  import-manifest ◼
kubernetes_manifest.example1: Refreshing state...

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with
the following symbols:
  ~ update in-place

  # kubernetes_manifest.example1 will be updated in-place
  ~ resource "kubernetes_manifest" "example1" {
      + manifest = {
          + apiVersion = "v1"
          + data       = {
              + TEST = "hello"
            }
          + kind       = "ConfigMap"
          + metadata   = {
              + name      = "example1"
              + namespace = "default"
            }
        }
    }

```

manifest/provider/import.go
jrhouston: the `#` also forces the whole thing to be quoted on the shell since it's the comment delimiter – you get a confusing error if you dont quote it.

alexsomesan:
> > Hmm, so I tried this and the quirk of not being able to access the config at import time definitely makes for a confusing user experience. I wonder if we should generate a warning to let people that this is going to happen, and that it should be ok to run apply again to normalize the state:
> 
> That's a very nice idea! I'll see if I can return an informational diagnostic message telling the user to re-run apply.

Added the followin message after import:
```
kubernetes_manifest.test: Importing from ID "apiVersion=v1,kind=Service,name=somesan,namespace=default"...
kubernetes_manifest.test: Import prepared!
  Prepared kubernetes_manifest for import
kubernetes_manifest.test: Refreshing state...

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.

╷
│ Warning: Apply needed after 'import'
│
│ Please run apply after a successful import to realign the resource state to the configuration in Terraform.
╵

```

manifest/provider/import.go
```diff
@@ -0,0 +1,211 @@
...
+// and extracts the values for GVK, name and (optionally) namespace of the target resource as required
+// during the import process.
+//
+// The expected format for the import resource ID is:
+//
+// "apiVersion=<value>,kind=<value>,name=<value>[,namespace=<value>"]
+//
+// where 'namespace' is only required for resources that expect a namespace.
+//
+// Example: "apiVersion=v1,kind=Secret,namespace=default,name=default-token-qgm6s"
```
jrhouston, line 179: It occurred to me that if the ordering of the keys is fixed we could just write a regex with named groups instead of having to parse the tokens and figure out if it's valid or not. Like this:

```go
func parseImportID(id string) (schema.GroupVersionKind, string, string, error) {
	re := regexp.MustCompile(
		`apiVersion=(?P<apiVersion>[-\w]+),` +
			`kind=(?P<kind>[-\w]+),` +
			`name=(?P<name>[-\w]+)` +
			`(,namespace=(?P<namespace>[-\w]+))?`)
	matches := re.FindStringSubmatch(id)
	if len(matches) == 0 {
		return schema.GroupVersionKind{}, "", "", fmt.Errorf("invalid import ID: expected format: apiVersion=<value>,kind=<value>,name=<value>[,namespace=<value>]")
	}

	m := map[string]string{"namespace": "default"}
	for _, k := range []string{"apiVersion", "kind", "name", "namespace"} {
		if v := matches[re.SubexpIndex(k)]; v != "" {
			m[k] = v
		}
	}

	gvk := schema.FromAPIVersionAndKind(m["apiVersion"], m["kind"])
	name := m["name"]
	namespace := m["namespace"]
	return gvk, name, namespace, nil
}
```

manifest/provider/import_test.go
```diff
@@ -0,0 +1,65 @@
...
+			Namespace: "foo",
+			Err:       nil,
+		},
+		{
+			ID:  "foobar",
+			Err: fmt.Errorf("invalid format for import ID [%s]\nExpected format is: apiVersion=<value>,kind=<value>,name=<value>[,namespace=<value>]", "foobar"),
+		},
+	}
+	for _, s := range samples {
+		gotGvk, gotName, gotNamespace, gotErr := parseImportID(s.ID)
```
jrhouston, line 45: We should rename `s` -> `expected` and then change the `got` prefix in each of these variables to `actual` – expected and actual are the usual names in our tests I think.

manifest/test/helper/kubernetes/kubernetes_helper.go
```diff
@@ -74,25 +74,56 @@ func (k *Helper) CreateNamespace(t *testing.T, name string) {
...
+	gvr := NewGroupVersionResource("v1", "namespaces")
 	_, err := k.client.Resource(gvr).Create(context.TODO(), namespace, metav1.CreateOptions{})
 	if err != nil {
 		t.Fatalf("Failed to create namespace %q: %v", name, err)
 	}
 }
 
-// DeleteNamespace deletes a namespace
-func (k *Helper) DeleteNamespace(t *testing.T, name string) {
+// CreateNamespace creates a new namespace
```
jrhouston, line 84: This should be CreateConfigMap

website/docs/r/manifest.html.markdown
```diff
@@ -98,6 +98,25 @@ resource "kubernetes_manifest" "test-crd" {
...
+### Extract the resource from Kubernetes and transform it into Terraform configuration
+
+```
+kubectl get secrets sample -o yaml | tfk8s --strip -o sample.tf
+```
+
+### Import the resource state from the cluster
+
+```
+terraform import kubernetes_manifest.secret_sample "v1#Secret#default#sample"
```
jrhouston, line 114: Need to update the syntax here and in the paragraph below.

jrhouston (changes requested):
Left some individual comments

manifest/provider/import.go
jrhouston: Just playing regex golf now, but something like this works if we want to allow any ordering of the key value pairs:

```go
func parseImportID(id string) (schema.GroupVersionKind, string, string, error) {
	re := regexp.MustCompile(
		`((apiVersion=(?P<apiVersion>[-\w]+))|` +
			`(kind=(?P<kind>[-\w]+))|` +
			`(name=(?P<name>[-\w]+))|` +
			`(namespace=(?P<namespace>[-\w]+)))`)
	matches := re.FindAllStringSubmatch(id, -1)
	if len(matches) < 3 {
		return schema.GroupVersionKind{}, "", "", fmt.Errorf("invalid import ID: expected format: apiVersion=<value>,kind=<value>,name=<value>[,namespace=<value>]")
	}

	m := map[string]string{"namespace": "default"}
	for _, mm := range matches {
		for _, k := range []string{"apiVersion", "kind", "name", "namespace"} {
			if v := mm[re.SubexpIndex(k)]; v != "" {
				m[k] = v
			}
		}
	}

	gvk := schema.FromAPIVersionAndKind(m["apiVersion"], m["kind"])
	name := m["name"]
	namespace := m["namespace"]
	return gvk, name, namespace, nil
}
```

alexsomesan: I can switch to the regexp version, I have no preference TBH.
But I'm pretty sure the multiple branch regexp in the second example is a bit slower than the simple iteration when tokenizing. Then again, we're not building real-time trading systems here, so I have no preference one way or the other.

Would you say the regexp version improves the code readability? I can't really tell for sure. I'm a fan of regexp myself, but I know a lot of folks aren't.

jrhouston: Nah – it's probably easier for someone new to the code to reason about a simple parser it they aren't a ninja with regular expressions and have awareness of named and capture groups and so on. We can keep as is, I'm just a fan of regex for little things like this.

alexsomesan: TBH, regexp used to be my first choice for things like these too. Until I noticed a lot of push-back from the less-than-ninja collaborators and I was surprised by how prevalent that is.

jrhouston (approved):

PR #1353, opened 2021-08-05
Add timeouts block to kubernetes_manifest

jrhouston:
I decided just to redo this feature, based on my previous PRs in the alpha repo, to be an implementation of the `timeouts` block as described here: https://www.terraform.io/docs/extend/resources/retries-and-customizable-timeouts.html

The main thing that's different than the old PRs is that I've used a context with a deadline rather than having the waiter support a timeout parameter. 

### Acceptance tests

```

...
```

### Release Note

```release-note
Add support for timeouts block to `kubernetes_manifest` resource
```

### References

https://github.com/hashicorp/terraform-provider-kubernetes-alpha/pull/238
https://github.com/hashicorp/terraform-provider-kubernetes-alpha/pull/237

### Community Note

### DIFF
--- a/manifest/provider/waiter.go
+++ b/manifest/provider/waiter.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math/big"
 	"regexp"
+	"time"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
@@ -109,6 +110,12 @@ type FieldWaiter struct {
 func (w *FieldWaiter) Wait(ctx context.Context) error {
 	w.logger.Info("[ApplyResourceChange][Wait] Waiting until ready...\n")
 	for {
+		if deadline, ok := ctx.Deadline(); ok {
+			if time.Now().After(deadline) {
+				return context.DeadlineExceeded
+			}
+		}
+
 		// NOTE The typed API resource is actually returned in the
 		// event object but I haven't yet figured out how to convert it
 		// to a cty.Value.
@@ -174,6 +181,8 @@ func (w *FieldWaiter) Wait(ctx context.Context) error {
 		if done {
 			return err
 		}
+
+		time.Sleep(1 * time.Second) // lintignore:R018
 	}
 }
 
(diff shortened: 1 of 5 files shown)


manifest/provider/waiter.go
```diff
@@ -174,6 +181,8 @@ func (w *FieldWaiter) Wait(ctx context.Context) error {
 		if done {
 			return err
 		}
+
+		time.Sleep(1 * time.Second) // lintignore:R018
```
jrhouston, line 185: We were hammering the API here so I added this sleep.

mmingdai:
This PR is great! I run into some issue recently due to the lack of waiting on deletion. It would be great if we can merge this and publish a version ASAP. Thanks!

mmingdai:
@alexsomesan ^

mmingdai:
👍

EcaterinaGr:
I have opened this issue too - they are both linked https://github.com/hashicorp/terraform-provider-kubernetes/issues/1357

mmingdai:
@alexsomesan

manifest/provider/waiter.go
alexsomesan: Nice! This should work great!

At some point we could make it extra spiffy by leveraging some of the exponential back-off helpers in API machinery and client-go:
https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait
https://pkg.go.dev/k8s.io/client-go/util/retry

alexsomesan: Let's just add a TODO so that this doesn't get forgotten (I will for sure).
```suggestion
              // TODO: implement with exponential back-off.
		time.Sleep(1 * time.Second) // lintignore:R018
```

alexsomesan (approved):
Looks good!

This seems not so trivial to test. 
I'm wondering if we could delay the API responses through an admission webhook that just sleeps a certain amount based on the presence of an annotation or label on the resource.
We could extend the webhook I added for the computed attributes test to do this as well. 

Let's re-think this once both PRs are merged.

jrhouston:
We could test it with a Deployment or Job that will never succeed so it times out.

PR #1426, opened 2021-09-27
Fix failing unit test for parseImportID

jrhouston:
### Description

This unit test was failing because parseImportID will set namespace to default if it's not present. We would have to check if the kind was a namespaced resource to figure out if we should omit it or not. AFAIK namespace should just get ignored for non-namespaced resources so this should be fine for now. 

### Acceptance tests

```

...
```

### Release Note

```release-note
NONE
```

### Community Note

### DIFF
--- a/manifest/provider/import_test.go
+++ b/manifest/provider/import_test.go
@@ -26,7 +26,7 @@ func TestParseImportID(t *testing.T) {
 			ID:        "apiVersion=rbac.authorization.k8s.io/v1,kind=ClusterRole,name=test",
 			GVK:       schema.GroupVersionKind{Group: "rbac.authorization.k8s.io", Version: "v1", Kind: "ClusterRole"},
 			Name:      "test",
-			Namespace: "",
+			Namespace: "default",
 			Err:       nil,
 		},
 		{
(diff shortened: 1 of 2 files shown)


manifest/provider/import_test.go
```diff
@@ -50,15 +50,15 @@ func TestParseImportID(t *testing.T) {
 			t.Fatal(actualErr.Error())
 		}
 		if expected.GVK != actualGvk {
-			t.Log("GVK (actual / wanted):", actualGvk, expected.GVK)
+			t.Logf("GVK did not match. actual: %q  expected: %q.", actualGvk, expected.GVK)
```
jrhouston, line 53: Took me a minute to figure out which case was actually failing so I made this a bit more readable.

keithmattix:
I actually took the opposite approach in [my PR](https://github.com/hashicorp/terraform-provider-kubernetes/pull/1423/files#diff-14c58efa27e2354bdcb761d2e5c092a6edf4a1a8bdff5b284a2b8901084c3f29R188) because I would think cluster scoped resources (like ClusterRole, or MutatingWebhookConfiguration) wouldn't need to carry around a namespace in their TF ID. If I'm wrong about that, I'm definitely open to modifying my PR

alexsomesan (approved):
LGTM! Nice improvement!

jrhouston:
> I actually took the opposite approach in my PR because I would think cluster scoped resources (like ClusterRole, or MutatingWebhookConfiguration) wouldn't need to carry around a namespace in their TF ID. If I'm wrong about that, I'm definitely open to modifying my PR

Thanks for commenting @keithmattix – I think it has to be this way around, otherwise people will always have to specify a `namespace=` in the import ID even if it is `default`. We want the function to set the namespace to `default` if it's not present. Ideally the function would check if the kind is namespaced or not. I've added a FIXME to tidy this up later, for the moment I just want to get the tests passing in main again.

PR #1439, opened 2021-10-05
Updating cron module and tests for cronjob in kuberenetes version 1.22

jdziat:
### Description

Fixes https://github.com/hashicorp/terraform-provider-kubernetes/issues/1459

### Acceptance tests

```
$ make testacc TESTARGS='-run=TestCronJobFormatting'

...
```

### Release Note

```release-note
Adding ability to use CRON_TZ in cronjobs for the versions of K8s that support it
```

### Community Note


kubernetes/schema_cron_job_spec_test.go
```diff
@@ -0,0 +1,23 @@
+package kubernetes
```
jrhouston, line 1: Let's put this unit test into validators_test.go rather than having its own file.

jrhouston: Also need to run go fmt on this file.

jrhouston (changes requested):
Thanks for updating this @jdziat – just a couple of changes.

kubernetes/schema_cron_job_spec_test.go
jdziat: @jrhouston thanks should be fixed now

kubernetes/validators_test.go
```diff
@@ -107,6 +107,22 @@ func TestValidateBase64EncodedMap(t *testing.T) {
 	}
 }
 
+func TestValidateCronJobFormatting(t *testing.T) {
+	validator := validateCronExpression()
+	_, es := validator("CRON_TZ=UTC 30 04 * * *", "Should parse out UTC")
```
jrhouston, line 112: Realized there's a couple more problems here:

- The second argument of the validation func is actually supposed to be the path to the attribute being validated. In this test case this is actually irrelevant and can be anything, but it confused me when I was looking at the test.
- The error statements below contain `#{err}` – I suspect you wanted to put the text of the error in the message, but you've accidentally used Ruby style string interpolation. Here's the docs for how to use formatting directives with the `Errorf` function in Go: https://pkg.go.dev/fmt#hdr-Printing

I also thought it would be better to make each test case a subtest, rather than 3 separate if statements. Here is how I would write the test:

```go
func TestValidateCronJobFormatting(t *testing.T) {
	testCases := []string{
		"30 04 * * *",
		"TZ=UTC 30 04 * * *",
		"CRON_TZ=UTC 30 04 * * *",
	}
	validator := validateCronExpression()
	for _, tc := range testCases {
		t.Run(tc, func(t *testing.T) {
			_, err := validator(tc, "spec.0.schedule")
			if len(err) != 0 {
				t.Errorf("Failed to parse cron expression %q: %v", tc, err)
			}
		})
	}
}
```

jrhouston (changes requested):

kubernetes/validators_test.go
jdziat: I don't use Goland much and it shows it collapsed like that. Good call outs i'll get them updated.

nitaigao:
Any update? Could really use this at the end of the month!

PR #1473, opened 2021-10-27
Add support for certificates/v1 API

jrhouston:
### Description

This PR adds support for the certificates/v1 API by adding a new `kubernetes_certificate_signing_request_v1` resource. 

### Acceptance tests

```
=== RUN   TestAccKubernetesCertificateSigningRequestV1_basic
--- PASS: TestAccKubernetesCertificateSigningRequestV1_basic (3.34s)
=== RUN   TestAccKubernetesCertificateSigningRequestV1_generateName
--- PASS: TestAccKubernetesCertificateSigningRequestV1_generateName (3.31s)
PASS
ok  	github.com/hashicorp/terraform-provider-kubernetes/kubernetes	7.706s
```

### Release Note

```release-note
Add support for certificates/v1 API 
```

### Community Note

### DIFF
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -216,7 +216,8 @@ func Provider() *schema.Provider {
 			"kubernetes_horizontal_pod_autoscaler": resourceKubernetesHorizontalPodAutoscaler(),
 
 			// certificates
-			"kubernetes_certificate_signing_request": resourceKubernetesCertificateSigningRequest(),
+			"kubernetes_certificate_signing_request":    resourceKubernetesCertificateSigningRequest(),
+			"kubernetes_certificate_signing_request_v1": resourceKubernetesCertificateSigningRequestV1(),
 
 			// rbac
 			"kubernetes_role":                 resourceKubernetesRole(),
--- a/kubernetes/resource_kubernetes_certificate_signing_request_test.go
+++ b/kubernetes/resource_kubernetes_certificate_signing_request_test.go
@@ -17,7 +17,10 @@ func TestAccKubernetesCertificateSigningRequest_basic(t *testing.T) {
 	usages := []string{"client auth"}
 	signerName := "kubernetes.io/legacy-unknown"
 	resource.Test(t, resource.TestCase{
-		PreCheck:          func() { testAccPreCheck(t) },
+		PreCheck: func() {
+			testAccPreCheck(t)
+			skipIfClusterVersionGreaterThanOrEqual(t, "1.22.0")
+		},
 		IDRefreshName:     "kubernetes_certificate_signing_request.test",
 		ProviderFactories: testAccProviderFactories,
 		CheckDestroy:      testAccCheckKubernetesCertificateSigningRequestDestroy,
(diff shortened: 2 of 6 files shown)


alexsomesan (approved):
Looks good, works in my testing.

sftim:
The trouble with putting the API version into the Terraform code: if in future you want to replace a `kubernetes_certificate_signing_request_v1` with a `kubernetes_certificate_signing_request_v2` (because you want to, eg, turn off support for v1 CSRs in your cluster), you'll have to do some state file surgery to keep your existing resources in the Kubernetes API.

With cloud providers that provide multiple APIs for the same backing resource, Terraform typically _doesn't_ expose details of which API Terraform is using. I'd expect to see a preference for any particular API configured at the provider level, not per-resource.

jrhouston:
Hi @sftim thanks for sharing your thoughts. Fundamentally, I agree with what you're saying – I would love it were possible to specify the API group/version (or use the discovery client to check) and then have the provider use the appropriate schema for the resource when the provider plugin is loaded, either by specifying it in the provider block or even the resource itself in the same way that Kubernetes does.  

Unfortunately down to the way the Terraform's protocol and type system works this isn't possible right now, and that kind of change is likely a major version away for Terraform if it's even feasible. The schema for the built-in resources (ones that aren't kubernetes_manifest) has to be known at build time, and there isn't a mechanism for resources to have multiple schema versions that can be built into the provider and toggled based on configuration. Certainly as Terraform evolves we will come back and revisit this. 

So at present it's kind of a trade-off between:
- Cobbling together frankenstein resources that combine the fields from multiple schemas to support different API versions, and have the provider figure out which API version needs to be used **(this is what we do now)**
- Support only the one version of an API and have user rely on version pinning to get the API versions they need, and then backport fixes and changes to previous versions of the provider. This gets confusing for things like HPA where there is v1 but also v2 that's in beta, and potentially means including multiple kubernetes providers pinned at different versions.
- Support distinct resources so we can have a cleanly defined schema for each API version, but rely on the user having to do some "state surgery" (i.e state rm, then import) if they need to move their resource to a different version.  

I have written more about our motivation for doing this on this (unreleased) [documentation page](https://github.com/hashicorp/terraform-provider-kubernetes/blob/main/website/docs/guides/versioned-resources.markdown) but basically the tl;dr is this is going to make it easier for us to use code generation to generate each resource in the provider from the Kubernetes API definition, rather than the currently laborious manual process it is today. The third option above (I feel) is the lesser of those 3 evils to get us to where the lag time between changes the Kubernetes API landing as built-in resources in the provider is much shorter.  

Hopefully that makes sense.

jrhouston:
> Terraform typically doesn't expose details of which API Terraform is using

Also, to your point on this, there is some prior art on this in the AWS provider where to support multiple versions of the underlying cloud API the version number actually is exposed in the resource name for a few services like CloudHSM and API Gateway. 

Another option that occurred to me would be to actually break the provider down into multiple provider plugins for each API group (e.g having `kubernetes-core`, `kubernetes-autoscaling` etc.) and let the user select API versions that way. That's something to be explored after we tackle code generation though.

sftim:
@jrhouston if there's a GitHub issue or discussion about this (in the `terraform-provider-kubernetes`, maybe I can point that out to some folks from the Kubernetes architecture folks and see if there's some options to recommend or explore?

I can start one but also very happy to let you take a lead there.

PR #1487, opened 2021-11-10
Fix code example in kubernetes_manifest

jkimbo:
### Description

This PR updates the provider code example to show to how you can enable the manifest experiment.

### Release Note

```release-note
NONE
```

### DIFF
--- a/website/docs/r/manifest.html.markdown
+++ b/website/docs/r/manifest.html.markdown
@@ -17,7 +17,11 @@ How to enable the experiment:
 
 ```hcl
 provider "kubernetes" {
-  config_path = "~/.kube/config"
+  experiments {
+    manifest_resource = true
+  }
+
+  // provider config...
 }
 ```
 


website/docs/r/manifest.html.markdown
```diff
@@ -17,7 +17,11 @@ How to enable the experiment:
 
 ```hcl
 provider "kubernetes" {
-  config_path = "~/.kube/config"
+  experiments {
```
jrhouston, line 20: This isn't required now, as we defaulted this to `true` in version v2.6.0.

alexsomesan:
@jkimbo As @jrhouston mentioned, setting the experiments for manifest flag is no longer necessary since it was defaulted to `true` in provider version v2.6.0. We're keeping the flag available in the provider schema for now to avoid unnecessarily breaking people's existing configurations. New configurations should not set the flag. We'll announce its deprecation in due time.

I'm going to close this PR, but not before thanking you for taking the time to look into this. We look forward to your future contributions.

PR #1504, opened 2021-11-22
Add documentation page explaining version suffixes

jrhouston:
### Description

This PR adds documentation explaining the version suffixes in resource names. 

```

...
```

### Release Note

```release-note
NONE
```

### Community Note

### DIFF
--- a/website/docs/guides/versioned-resources.markdown
+++ b/website/docs/guides/versioned-resources.markdown
@@ -28,7 +28,7 @@ Ultimately, we plan to completely automate the generation of Terraform resources
 
 ## What will happen to the resources without versions in the name?
 
-These resources will continue to be supported and maintained as is until the next major version of the provider, at which point they will be deprecated and then subsequently removed in the following major version. 
+These resources will continue to be supported and maintained as is through to v3.0.0 of the provider, at which point they will be marked as deprecated and then subsequently removed in v4.0.0.
 
 
 ## `v1` resources
@@ -48,7 +48,7 @@ We will continue to bring support for API resources which reach `beta` however i
 We will continue our policy of not building support for `alpha` versioned resources into the provider. Please use the [kubernetes_manifest](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest) resource to manage those resources. 
 
 
-## How can I move an resource without a version to its versioned resource name?
+## How can I move a resource without a version to its versioned resource name?
 
 The simplest, non-destructive way to do this is to modify the name of the resource to include the version suffix. Then remove the old resource from state and import the resource under the versioned resource like so:
 
@@ -57,4 +57,6 @@ terraform state rm kubernetes_config_map.example
 terraform import kubernetes_config_map_v1.example default/example
 ```
 
-You can also skip this and just allow Terraform to destroy and recreate the resource, but this is not recommend for resources like `kubernetes_service` and `kubernetes_deployment`. 
+Then run `terraform plan` to confirm that the import was successful. **NOTE: Do not run the plan after renaming the resource in the configuration until after the above steps have been carried out.** 
+
+You can also skip this and just allow Terraform to destroy and recreate the resource, but this is not recommended for resources like `kubernetes_service` and `kubernetes_deployment`. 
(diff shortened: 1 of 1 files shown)


website/docs/guides/versioned-resources.markdown
```diff
@@ -0,0 +1,60 @@
...
+
+Ultimately, we plan to completely automate the generation of Terraform resources to cover the core Kubernetes API. Having a set of versioned schemas that more closely matches the Kubernetes API definition is going to make this easier to achieve and will enable us to add built-in support for new API versions much faster. 
+
+
+## What will happen to the resources without versions in the name?
+
+These resources will continue to be supported and maintained as is until the next major version of the provider, at which point they will be deprecated and then subsequently removed in the following major version.
```
redeux, line 31: It might be worth calling out the version numbers, since we know them.  deprecated in v3, removed in v4.

website/docs/guides/versioned-resources.markdown
```diff
@@ -0,0 +1,62 @@
...
+
+We will continue our policy of not building support for `alpha` versioned resources into the provider. Please use the [kubernetes_manifest](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest) resource to manage those resources. 
+
+
+## How can I move a resource without a version to its versioned resource name?
+
+The simplest, non-destructive way to do this is to modify the name of the resource to include the version suffix. Then remove the old resource from state and import the resource under the versioned resource like so:
+
+```
+terraform state rm kubernetes_config_map.example
```
alexsomesan, line 56: Would `terraform state mv` not be an easier option in this case?

website/docs/guides/versioned-resources.markdown
```diff
@@ -0,0 +1,62 @@
...
+
+Ultimately, we plan to completely automate the generation of Terraform resources to cover the core Kubernetes API. Having a set of versioned schemas that more closely matches the Kubernetes API definition is going to make this easier to achieve and will enable us to add built-in support for new API versions much faster. 
+
+
+## What will happen to the resources without versions in the name?
+
+These resources will continue to be supported and maintained as is through to v3.0.0 of the provider, at which point they will be marked as deprecated and then subsequently removed in v4.0.0.
+
+
+## `v1` resources
```
alexsomesan, line 34: This rule applies to all "mature" apiVersions.
Should we call this something like `>= v1` or `v1+` ?

website/docs/guides/versioned-resources.markdown
```diff
@@ -0,0 +1,62 @@
...
+---
+
+# Versioned resource names 
+
+This guide explains the naming conventions for resources and data sources in the Kubernetes provider. 
+
+
+## Version suffixes
+
+From provider version v2.7.0 onwards Terraform resources and data sources that cover the [core set of Kubernetes APIs](https://kubernetes.io/docs/reference/kubernetes-api/) will be suffixed with their corresponding Kubernetes API version (e.g `v1`, `v2`, `v2beta1`). The existing resources in the provider will continue to be maintained as is.
```
alexsomesan, line 15: Kubernetes also has an API group called "core".

In this case, I would use "standard" or some other term to avoid confusion with the actual API group and not imply that it only applies to those resources.

alexsomesan (approved):
Very nicely and thoroughly written. Like it a lot!

I highlighted a couple of small things that stood out, but nothing critical.

website/docs/guides/versioned-resources.markdown
jrhouston: Yep, I'll change it to `v1` and above. It's just that there are no `v2` resources yet.

website/docs/guides/versioned-resources.markdown
jrhouston: I considered this but `terraform state mv` does not work because, even though the resources actually share the same schema, terraform considers them different types and so produces an error message if you try to use mv:

```
╷
│ Error: Invalid state move request
│
│ Cannot move kubernetes_config_map.example to kubernetes_config_map_v1.example: resource types don't match.
╵
```

This only works for the same resource type, e.g for changing the name or moving it into a module.

PR #1548, opened 2021-12-13
Add `kubernetes_resource` data source

jrhouston:
### Description

This PR adds a generic `kubernetes_resource` data source as an analogue to the `kubernetes_manifest` resource. 

### Acceptance tests

```

...
```

### Release Note

```release-note
Add kubernetes_resource data source
```

### Community Note

### DIFF
--- a/manifest/provider/datasource.go
+++ b/manifest/provider/datasource.go
@@ -1,0 +1,207 @@
+package provider
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/hashicorp/terraform-provider-kubernetes/manifest/morph"
+	"github.com/hashicorp/terraform-provider-kubernetes/manifest/payload"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// ReadDataSource function
+func (s *RawProviderServer) ReadDataSource(ctx context.Context, req *tfprotov5.ReadDataSourceRequest) (*tfprotov5.ReadDataSourceResponse, error) {
+	s.logger.Trace("[ReadDataSource][Request]\n%s\n", dump(*req))
+
+	resp := &tfprotov5.ReadDataSourceResponse{}
+
+	execDiag := s.canExecute()
+	if len(execDiag) > 0 {
+		resp.Diagnostics = append(resp.Diagnostics, execDiag...)
+		return resp, nil
+	}
+
+	rt, err := GetDataSourceType(req.TypeName)
+	if err != nil {
+		resp.Diagnostics = append(resp.Diagnostics, &tfprotov5.Diagnostic{
+			Severity: tfprotov5.DiagnosticSeverityError,
+			Summary:  "Failed to determine data source type",
+			Detail:   err.Error(),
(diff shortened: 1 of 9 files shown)


website/docs/d/resource.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+---
+
+# kubernetes_resource
+
+This data source is a generic way to retrieve resources from the Kubernetes API. 
+
+### Example: Get data from a ConfigMap
+
+```hcl
+data "kubernetes_resource" "example" {
```
jrhouston, line 15: @alexsomesan open to thoughts on the naming here, I didn't use `manifest` because that specifically refers to the YAML file you apply to the cluster whereas here we're retrieving an API resource so I thought `resource` made sense. `kubernetes_object` would also work too.

website/docs/d/resource.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+  apiVersion = "v1"
```
jrhouston, line 16: Worth noting that we're going to break the `snake_case` convention that is used everywhere else (and all other providers too) for resource attributes. `api_version` would be fine here, just that the camelCase convention is probably more familiar to users of this provider.

manifest/provider/datasource.go
```diff
@@ -0,0 +1,164 @@
...
+		d := tfprotov5.Diagnostic{
+			Severity: tfprotov5.DiagnosticSeverityError,
+			Summary:  fmt.Sprintf("Failed to get data source"),
+			Detail:   err.Error(),
+		}
+		resp.Diagnostics = append(resp.Diagnostics, &d)
+		return resp, nil
+	}
+
+	objectType, err := s.TFTypeFromOpenAPI(ctx, gvk, false)
```
alexsomesan, line 123: A point could be made that this should best happen before the actual API Get call. Reason being that it might serve as validation for the correctness of the input attributes independently of potential permission errors on the resource Get call.
What I'm not sure about is whether client-go distinguishes (with different error codes) between a RBAC permission issue on a valid path and an invalid path (likely also reported by 'not found').

manifest/provider/datasource.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return resp, fmt.Errorf("failed to determine data source type: %s", err)
```
alexsomesan, line 125: Diagnostic vs raw error here too

...and a few more spot down below.

website/docs/d/resource.html.markdown
alexsomesan: I honestly am ok with either option. There are valid points, which you raised, for both of them.
I'm with you there that keeping with the spirit of this resources trying to track as close as possible to native K8s, this looks like the slightly better option.

website/docs/d/resource.html.markdown
alexsomesan: I'm with you. This is a good choice of name. Using manifest would likely be more confusing. Also, this datasource could be used together with the native resources (non kubernetes_manifest) and then it makes even more sense to call it like this.

alexsomesan (commented):
I like it! I played around with the datasource and it works nicely as expected.

I dropped some observations about UX down below, but structurally this is great work! 
Thanks! a lot!

website/docs/d/resource.html.markdown
jrhouston: I've changed this to `api_version` after my conversation with DevEx – they recommended not breaking the convention.

jrhouston:
As discussed, have changed `metadata` to a block instead of an object – this should be good to go.

alexsomesan:
Looks good! How do you feel about the Diagnostics suggestions I made earlier?

jrhouston:
Yeah sounds good, I've wrapped all the errors in diagnostics.

alexsomesan (approved):
LGTM

jrhouston:
Looks like the tests are failing because I missed something when rebasing  – will fix.

PR #1586, opened 2022-01-26
Fix kubernetes default service account

husunal:
### Description

Fixes `kubernetes_default_service_account` to set `automount_service_account_token` to false. 
Issue https://github.com/hashicorp/terraform-provider-kubernetes/issues/1247

### Acceptance tests

```
=== RUN   TestAccKubernetesDefaultServiceAccount_basic
--- PASS: TestAccKubernetesDefaultServiceAccount_basic (16.35s)
=== RUN   TestAccKubernetesDefaultServiceAccount_secrets
--- PASS: TestAccKubernetesDefaultServiceAccount_secrets (10.32s)
=== RUN   TestAccKubernetesDefaultServiceAccount_automountServiceAccountToken
--- PASS: TestAccKubernetesDefaultServiceAccount_automountServiceAccountToken (10.77s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   38.141s
```

### Release Note

```release-note
Fix `kubernetes_default_service_account` doesn't set the `automount_service_account_token` to false.
```

### References
https://github.com/hashicorp/terraform-provider-kubernetes/issues/1247

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_default_service_account.go
+++ b/kubernetes/resource_kubernetes_default_service_account.go
@@ -60,20 +60,27 @@ func resourceKubernetesDefaultServiceAccountCreate(ctx context.Context, d *schem
 		return diag.FromErr(err)
 	}
 
+	secret, err := getServiceAccountDefaultSecret(ctx, "default", svcAcc, d.Timeout(schema.TimeoutCreate), conn)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.Set("default_secret_name", secret.Name)
+
 	ops := patchMetadata("metadata.0.", "/metadata/", d)
-	imagePullSecrets := d.Get("image_pull_secret").(*schema.Set).List()
-	ops = append(ops, &ReplaceOperation{
-		Path:  "/imagePullSecrets",
-		Value: expandLocalObjectReferenceArray(imagePullSecrets),
-	})
-
-	secrets := d.Get("secret").(*schema.Set).List()
-	defaultSecretName := d.Get("default_secret_name").(string)
-
-	ops = append(ops, &ReplaceOperation{
-		Path:  "/secrets",
-		Value: expandServiceAccountSecrets(secrets, defaultSecretName),
-	})
+	if d.HasChange("image_pull_secret") {
+		v := d.Get("image_pull_secret").(*schema.Set).List()
+		ops = append(ops, &ReplaceOperation{
+			Path:  "/imagePullSecrets",
+			Value: expandLocalObjectReferenceArray(v),
+		})
+	}
+	if d.HasChange("secret") {
+		v := d.Get("secret").(*schema.Set).List()
+		ops = append(ops, &ReplaceOperation{
+			Path:  "/secrets",
+			Value: expandServiceAccountSecrets(v, secret.Name),
+		})
+	}
 
 	automountServiceAccountToken := d.Get("automount_service_account_token").(bool)
 	ops = append(ops, &ReplaceOperation{
@@ -93,12 +100,6 @@ func resourceKubernetesDefaultServiceAccountCreate(ctx context.Context, d *schem
 	log.Printf("[INFO] Submitted updated default service account: %#v", out)
 
 	d.SetId(buildId(metadata))
-
-	secret, err := getServiceAccountDefaultSecret(ctx, "default", svcAcc, d.Timeout(schema.TimeoutCreate), conn)
-	if err != nil {
-		return diag.FromErr(err)
-	}
-	d.Set("default_secret_name", secret.Name)
 
 	return resourceKubernetesServiceAccountRead(ctx, d, meta)
 }
--- a/kubernetes/resource_kubernetes_default_service_account.go
+++ b/kubernetes/resource_kubernetes_default_service_account.go
@@ -61,22 +61,19 @@ func resourceKubernetesDefaultServiceAccountCreate(ctx context.Context, d *schem
 	}
 
 	ops := patchMetadata("metadata.0.", "/metadata/", d)
-	if d.HasChange("image_pull_secret") {
-		v := d.Get("image_pull_secret").(*schema.Set).List()
-		ops = append(ops, &ReplaceOperation{
-			Path:  "/imagePullSecrets",
-			Value: expandLocalObjectReferenceArray(v),
-		})
-	}
-	if d.HasChange("secret") {
-		v := d.Get("secret").(*schema.Set).List()
-		defaultSecretName := d.Get("default_secret_name").(string)
+	imagePullSecrets := d.Get("image_pull_secret").(*schema.Set).List()
+	ops = append(ops, &ReplaceOperation{
+		Path:  "/imagePullSecrets",
+		Value: expandLocalObjectReferenceArray(imagePullSecrets),
+	})
 
-		ops = append(ops, &ReplaceOperation{
-			Path:  "/secrets",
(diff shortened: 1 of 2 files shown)


kubernetes/resource_kubernetes_default_service_account.go
```diff
@@ -59,6 +60,25 @@ func resourceKubernetesDefaultServiceAccountCreate(ctx context.Context, d *schem
 		return diag.FromErr(err)
 	}
 
+	if v, ok := d.GetOkExists("automount_service_account_token"); ok {
```
arybolovlev, line 63: This condition is verifying whether the argument `automount_service_account_token` is in the terraform code and if so, it will be taken to apply. Let's remove the condition and let Terraform pick up a default value of the argument if it is not set. In this particular case, it will be `true` and that is matches with the default value on the Kubernetes end.

kubernetes/resource_kubernetes_default_service_account.go
```diff
@@ -59,6 +60,25 @@ func resourceKubernetesDefaultServiceAccountCreate(ctx context.Context, d *schem
...
+		ops := patchMetadata("metadata.0.", "/metadata/", d)
+		ops = append(ops, &ReplaceOperation{
+			Path:  "/automountServiceAccountToken",
+			Value: v.(bool),
+		})
```
arybolovlev, line 68: Could you please also add the replacement operation for the `image_pull_secret` and `secret` arguments?

kubernetes/resource_kubernetes_default_service_account.go
```diff
@@ -59,6 +60,25 @@ func resourceKubernetesDefaultServiceAccountCreate(ctx context.Context, d *schem
...
+			return diag.Errorf("Failed to marshal update operations: %s", err)
+		}
+		log.Printf("[INFO] Updating default service account %q: %v", metadata.Name, string(data))
+		out, err := conn.CoreV1().ServiceAccounts(metadata.Namespace).Patch(ctx, metadata.Name, pkgApi.JSONPatchType, data, metav1.PatchOptions{})
+		if err != nil {
+			return diag.Errorf("Failed to update default service account: %s", err)
+		}
+		log.Printf("[INFO] Submitted updated default service account: %#v", out)
+	}
+
```
arybolovlev, line 81: With the above changes in place, we don't need to call the update function at the end of the create function. Instead, we can go ahead and call the read function right away to get the current state of the resource from Kubernetes.

Could you please replace `resourceKubernetesServiceAccountUpdate` with `resourceKubernetesServiceAccountRead`?

arybolovlev:
Hi @husunal,

Thank you for your contribution to this provider and the great job you did. I have left a few comments on your changes. If you have any questions, please let me know.

Thank you!

kubernetes/resource_kubernetes_default_service_account.go
husunal: I thought update function is still needed for the other arguments like `image_pull_secret` or `secret` but as you suggested I'll move them to this resource and call `resourceKubernetesServiceAccountRead`.

kubernetes/resource_kubernetes_default_service_account.go
husunal: Thanks for the review @arybolovlev 
Since the default value is `true`, `GetOkExists` will also get that default value but seems like we don't have a use case to use `GetOkExists` here.

husunal:
Hi @arybolovlev

I made the suggested changes, ready for review.

Thanks

arybolovlev:
Hi @husunal,

I am sorry for the delay here. I just realized that I didn't send a comment. There is one small thing I would like to ask you to change and we are good to go. Great job!

Thank you.

kubernetes/resource_kubernetes_default_service_account.go
```diff
@@ -59,6 +60,38 @@ func resourceKubernetesDefaultServiceAccountCreate(ctx context.Context, d *schem
...
+	ops := patchMetadata("metadata.0.", "/metadata/", d)
+	imagePullSecrets := d.Get("image_pull_secret").(*schema.Set).List()
+	ops = append(ops, &ReplaceOperation{
+		Path:  "/imagePullSecrets",
+		Value: expandLocalObjectReferenceArray(imagePullSecrets),
+	})
+
+	secrets := d.Get("secret").(*schema.Set).List()
+	defaultSecretName := d.Get("default_secret_name").(string)
+
```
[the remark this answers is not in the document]
husunal: Thanks @arybolovlev

Since the default service account is already created by K8s `default_secret_name` shouldn't return an empty string, same as https://github.com/hashicorp/terraform-provider-kubernetes/blob/main/kubernetes/resource_kubernetes_service_account.go#L319

Please correct me if I'm wrong.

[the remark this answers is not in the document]
husunal: @arybolovlev it would be great if you can re-review.

husunal:
@arybolovlev do you mind reviewing this MR again? Thanks!

kubernetes/resource_kubernetes_default_service_account.go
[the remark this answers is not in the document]
arybolovlev: Hi @husunal,

`default_secret_name` is something that doesn't exist in the Kubernetes and is purely computed by this provider. It means that you cannot configure this attribute in your Terraform code. If you try, you will get the following error:
```
Can't configure a value for "default_secret_name": its value will be decided automatically based on the result of applying this configuration.
```
That happens on the schema validation step before any CRUD operations.

The create operation that you have added is called only when Terraform detects that the given code has a resource that doesn't exist in the state file. Then it gets resource attributes and does whatever is necessary with them to call Kubernetes API. At this point, since we cannot use attribute `default_secret_name` in our Terraform code, the `Get` method will return the empty string. The function, that you call later, `expandServiceAccountSecrets` can handle case when one of the arguments `defaultSecretName` is empty string. If we know that the string in our case in empty all the time, let just pass emtpy string instead of getting it.

Actually, I found another interesting behaviour here.
I have the following Terraform code:
```hcl
resource "kubernetes_default_service_account" "this" {
  metadata {
    namespace = "this"
  }

}
```
As you can see `image_pull_secrets` and `secrets` attributes are ommited. Below is a JSON that will be passed to `Patch` method:
```
[
    {"path":"/imagePullSecrets","value\":[],"op":"replace"},
    {"path":"/secrets","value":[],"op":"replace"},
    {"path":"/automountServiceAccountToken","value":false,"op":"replace"}
]
```
It means that `imagePullSecrets` and `secrets` will be erased from the default service account on the Kubernetes end if they don't set in Terraform code. Moreover, if you pass `secrets` and none of them has type `kubernetes.io/service-account-token`, then a new token will be generated for the default service account.

I would say, that we need to check if length of `imagePullSecrets` and `secrets` more than 0 and only then append `ops`. In this case, I would also recomment to replace `defaultSecretName := d.Get("default_secret_name").(string)` with function `getServiceAccountDefaultSecret` that you use below to get default secret name. Then we don't need to pass empty string to function `expandServiceAccountSecrets`.

Thank you.

arybolovlev (commented):
I have added a more detailed answer and found another behavior that we need to fix. This is related to the fact that with this create operation we actually do updates.

husunal:
@arybolovlev Thanks for your detailed review. Ready for the review again!

arybolovlev (approved):
Great job, @husunal!! 👍🏻 🚀

Thanks a lot for your contribution and patience!

PR #1595, opened 2022-02-08
Add nested condition block to wait block

jrhouston:
### Description

This PR adds a nested `condition` block to the `wait` block to configure sets of conditions to wait to be met. 

### Acceptance tests

```

...
```

### Release Note

```release-note
Add nested condition block to wait block 
```

### Community Note

### DIFF
--- a/manifest/provider/waiter.go
+++ b/manifest/provider/waiter.go
@@ -59,6 +59,19 @@ func NewResourceWaiter(resource dynamic.ResourceInterface, resourceName string, 
 		}
 	}
 
+	if v, ok := waitForBlockVal["condition"]; ok {
+		var conditionsBlocks []tftypes.Value
+		v.As(&conditionsBlocks)
+		if len(conditionsBlocks) > 0 {
+			return &ConditionsWaiter{
+				resource,
+				resourceName,
+				conditionsBlocks,
+				hl,
+			}, nil
+		}
+	}
+
 	fields, ok := waitForBlockVal["fields"]
 	if !ok || fields.IsNull() || !fields.IsKnown() {
 		return &NoopWaiter{}, nil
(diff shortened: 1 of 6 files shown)


website/docs/r/manifest.html.markdown
```diff
@@ -200,6 +200,7 @@ The following arguments are supported:
 #### Arguments
 
 - **rollout** (Optional) When set to `true` will wait for the resource to roll out, equivalent to `kubectl rollout status`. 
+- **conditions** (Optional) A set of conditions to wait for.
```
arybolovlev, line 203: The `conditions` block has two nested attributes `status` and `type`. Could you please mention them in the documentation too?

One small comment regarding style here. Could you please make all argument names formatted as `code` instead of **bold text**?

arybolovlev (approved):
Nice addition! 🚀 Just one minor comment in the documentations section.

manifest/provider/waiter.go
```diff
@@ -300,3 +313,67 @@ func (w *RolloutWaiter) Wait(ctx context.Context) error {
...
+		status := res.Object["status"].(map[string]interface{})
+		conditions := status["conditions"].([]interface{})
+		conditionsMet := map[string]bool{}
+		for _, c := range w.conditions {
+			var condition map[string]tftypes.Value
+			c.As(&condition)
+			var conditionType, conditionStatus string
+			condition["type"].As(&conditionType)
+			condition["status"].As(&conditionStatus)
+			conditionsMet[conditionType] = false
```
alexsomesan, line 354: This could actually just be a boolean that is progressively AND-ed to the current condition status in the loop. This way, it only becomes true when no condition is false. Seems a bit simpler, since we're not reporting the individual conditions status up to the user.

manifest/provider/waiter.go
```diff
@@ -300,3 +313,67 @@ func (w *RolloutWaiter) Wait(ctx context.Context) error {
...
+			for _, cc := range conditions {
+				ccc := cc.(map[string]interface{})
+				if ccc["type"].(string) != conditionType {
+					continue
+				}
+				conditionsMet[conditionType] = ccc["status"].(string) == conditionStatus
```
alexsomesan, line 360: AND updating of boolean `conditionsMet` could happen here.

manifest/provider/waiter.go
```diff
@@ -300,3 +313,67 @@ func (w *RolloutWaiter) Wait(ctx context.Context) error {
...
+			}
+		}
+
+		done := true
+		for _, v := range conditionsMet {
+			if !v {
+				done = false
+			}
+		}
```
alexsomesan, line 369: This wouldn't be needed when using accumulator boolean. Instead of "done", the check would be on directly "conditionsMet".

alexsomesan (commented):
Looks good!
Small observation where we could simplify logic a bit, otherwise all good stuff!

PR #1615, opened 2022-02-25
Add release GitHub Action

jrhouston:
### Description

This adds a new action to replace the TeamCity release process as per the [migration guide](https://github.com/hashicorp/engineering-docs/blob/main/terraform/provider/migrating-releases-to-github-actions.md).

### DIFF
--- a/terraform-registry-manifest.json
+++ b/terraform-registry-manifest.json
@@ -1,0 +1,6 @@
+{
+    "version": 1,
+    "metadata": {
+        "protocol_versions": ["5.0"]
+    }
+}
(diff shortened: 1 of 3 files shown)


terraform-registry-manifest.json
```diff
@@ -0,0 +1,6 @@
+{
+    "version": 1,
+    "metadata": {
+        "protocol_versions": ["5.0"]
```
alexsomesan, line 4: I'm not sure how precise or exhaustive this list has to be, but the manifest provider uses protocol, `5.2`
Unsure if we need to mention that in there. I'll check with the registry folks.

alexsomesan: Seems like we only need to specify major versions.
5.0 should be fine for us.

alexsomesan (approved):
LGTM

PR #1616, opened 2022-02-25
Add `kubernetes_labels` resource

jrhouston:
### Description

This PR adds a `kubernetes_labels` resource for managing labels for Kubernetes resources which already exist. 

## Discussion 

There are some quirks to be considered here as this resource is managing a part of a Kubernetes resource that already exists:
- Should we delete labels that are already there if they are not present in the config? If not, this resource will likely often produce diffs after it is created. 
- Should we delete all the labels when the resource is destroyed, or simply "disown" them?
- Should this be a "logical" resource that simply patches the labels onto the resource, and then any subsequent change to the resource is a ForceNew?

The way this works at the moment:
- When the resource is created, the labels field will be overwritten with whatever is in the terraform config
- Updates to the resource will do a JSON patch allowing labels to be added and deleted individually
- Destroying the resource will delete all of the labels 

#### edit: Follow up to Discussion 

It occurred to me that we could use server-side apply and Field Manager to solve the problem of Terraform either blowing away pre-existing resources, or showing spurious diffs for labels not in the configuration. This PR has been updated to use field manager such that only labels present in the configuration will managed by Terraform, and anything else will be ignored. I've also added a `force` option so that labels managed by a different field manager can be forcibly overwritten if needed.  

## Example

```terraform
resource kubernetes_labels example {
  apiVersion = "v1"
  kind       = "ConfigMap"
  metadata   = {
    name = "example"
  }
  labels = {
    "owner" = "team"
    "app"   = "test"
  }
}
```

### TODO
- ~[ ] Add import support~ Note: we can't read the config to see what labels we should claim ownership of, and there wont be any managed fields metadata to read the labels from on the k8s API, because they're not managed by us. So import doesn't really make sense for this resource.
- [x] Add kind and apiVersion to the resource ID 
- [x] Use Field Manager so the resource doesn't blow away pre-existing labels
- [x] Add `force` option so labels managed by another client can be overwritten 

### Acceptance tests

```

...
```

### Release Note

```release-note
Add `kubernetes_labels` resource
```

### References

#692 

### Community Note


kubernetes/provider.go
```diff
@@ -27,6 +28,8 @@ import (
 	aggregator "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset"
 )
 
+const defaultFieldManagerName = "Terraform"
```
jrhouston, line 31: we should probably have a module at the top level for things like this that are shared between the main provider and manifest

website/docs/r/labels.html.markdown
```diff
@@ -0,0 +1,48 @@
+---
+subcategory: "helper"
```
jrhouston, line 2: Not really sure how to categorize this, maybe it doesn't need a category at all.

kubernetes/structures.go
```diff
@@ -26,6 +26,15 @@ func buildId(meta metav1.ObjectMeta) string {
...
 
+func buildIdWithVersionKind(meta metav1.ObjectMeta, kind, apiVersion string) string {
+	id := fmt.Sprintf("apiVersion=%v,kind=%v,name=%s",
+		apiVersion, kind, meta.Name)
+	if meta.Namespace != "" {
+		id += fmt.Sprintf(",namespace=%v", meta.Namespace)
+	}
+	return id
+}
+
```
jrhouston, line 37: We have a function for parsing this in the manifest provider, but I'll need to shuffle some things around to share it between both providers.

kubernetes/resource_kubernetes_labels.go
```diff
@@ -0,0 +1,196 @@
...
+	if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
+		if namespace == "" {
+			namespace = "default"
+		}
+		r = conn.Resource(mapping.Resource).Namespace(namespace)
+	} else {
+		r = conn.Resource(mapping.Resource)
+	}
+
+	old, new := d.GetChange("labels")
```
jrhouston, line 164: If we are to allow the existing labels to remain, here we could call `d.IsNewResource()` and if so fetch the current map of labels and use that to craft the patch below.

website/docs/r/labels.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+# kubernetes_labels
+
+This resource allows Terraform to manage the labels for a resource that already exists.
+
+~> **Note:** Existing labels must be included in the configuration for the resource, otherwise they will be destroyed when the resource is created. 
+
+## Example Usage
+
+```hcl
+data "kubernetes_labels" "example" {
```
arybolovlev, line 18: It looks like a small typo here and should be `resource` instead of `data`.

website/docs/r/labels.html.markdown
arybolovlev: I think we can omit the subcategory here and let it leave under `Resources` together with `kubernetes_manifest`.

kubernetes/resource_kubernetes_labels.go
```diff
@@ -0,0 +1,196 @@
...
+	metadata := expandMetadata(d.Get("metadata").([]interface{}))
+	name := metadata.GetName()
+	namespace := metadata.GetNamespace()
+
+	// figure out which resource client to use
+	dc, err := m.(KubeClientsets).DiscoveryClient()
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	agr, err := restmapper.GetAPIGroupResources(dc)
```
arybolovlev, line 91: Do we want to handle an error that potentially can be returned here? If so, then the same question for the line [141](https://github.com/hashicorp/terraform-provider-kubernetes/pull/1616/files#diff-73ee614c2215562c5d1ba3f2a74f548683458bab4735ff64730748458dd59c85R141).

simongottschlag:
@jrhouston @arybolovlev not sure if you've seen the PR I created a while ago: https://github.com/hashicorp/terraform-provider-kubernetes/pull/1562

It would be great to know if you don't want it and if that's the case I'll close it.

jrhouston:
I think this is ready for final review @arybolovlev @alexsomesan

arybolovlev (approved):
Amazing job, @jrhouston!

jrhouston:
@arybolovlev As discussed, I've updated the PR to:

- Check the resource exists before trying to PATCH
- Only call SetId() if the patch is successful at create time

arybolovlev:
Great! That works! :)

I think we need to cover one more scenario here when a target label resource doesn't exist anymore:
- Create a k8s resource manually(simulation of the case when a resource is not managed by the provider)
- Add labels via the provider
- Manually delete the resource(simulation of the case when a resource is not managed by the provider)

Since this time the TF state file remains broken. Even if I clean up my code by removing resource `kubernetes_labels`, it still tries to read k8s resource state based on the TF state file, gets 404, and exits with an error. What do you think?

jrhouston:
Good call @arybolovlev, I've updated it so that the user gets a warning if the underlying resource has been deleted outside of terraform:

```
╷
│ Warning: Resource deleted
│
│   with kubernetes_labels.example,
│   on main.tf line 5, in resource "kubernetes_labels" "example":
│    5: resource kubernetes_labels example {
│
│ The underlying resource "example-cm" has been deleted. You should recreate the underlying resource, or remove
│ it from your configuration.
╵
```

When they remove it from their config, if the underlying resource is already gone the delete operation will short-circuit to being done instead of throwing an error.

arybolovlev:
Just tested your last changes, great! I don't have any other scenarios in my head. 🚀

PR #1647, opened 2022-03-16
Add support for HPA Container Resource metrics

arybolovlev:
### Description

This PR adds support for HPA Container Resource metrics. More details in Kubernetes [documentation](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#container-resource-metrics).

### Acceptance tests

```
$ make testacc TESTARGS='-run ^TestAccKubernetesHorizontalPodAutoscalerV2Beta2*'

=== RUN   TestAccKubernetesHorizontalPodAutoscalerV2Beta2_basic
--- PASS: TestAccKubernetesHorizontalPodAutoscalerV2Beta2_basic (12.44s)
=== RUN   TestAccKubernetesHorizontalPodAutoscalerV2Beta2_containerResource
--- PASS: TestAccKubernetesHorizontalPodAutoscalerV2Beta2_containerResource (5.79s)
PASS
ok  	github.com/hashicorp/terraform-provider-kubernetes/kubernetes	18.748s
```

### Release Note

```release-note
Add a new metric type `container_resource` in resource `kubernetes_horizontal_pod_autoscaler_v2beta2`.
```

### References

https://github.com/hashicorp/terraform-provider-kubernetes/issues/1637

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_horizontal_pod_autoscaler_v2beta2.go
+++ b/kubernetes/resource_kubernetes_horizontal_pod_autoscaler_v2beta2.go
@@ -53,7 +53,6 @@ func resourceKubernetesHorizontalPodAutoscalerV2Beta2() *schema.Resource {
 						"behavior": {
 							Type:        schema.TypeList,
 							Description: "Behavior configures the scaling behavior of the target in both Up and Down directions (scale_up and scale_down fields respectively).",
-							Computed:    true,
 							Optional:    true,
 							MaxItems:    1,
 							Elem: &schema.Resource{
--- a/kubernetes/resource_kubernetes_horizontal_pod_autoscaler_v2beta2.go
+++ b/kubernetes/resource_kubernetes_horizontal_pod_autoscaler_v2beta2.go
@@ -53,6 +53,7 @@ func resourceKubernetesHorizontalPodAutoscalerV2Beta2() *schema.Resource {
 						"behavior": {
 							Type:        schema.TypeList,
 							Description: "Behavior configures the scaling behavior of the target in both Up and Down directions (scale_up and scale_down fields respectively).",
+							Computed:    true,
 							Optional:    true,
 							MaxItems:    1,
 							Elem: &schema.Resource{
--- a/kubernetes/resource_kubernetes_horizontal_pod_autoscaler_v2beta2_test.go
+++ b/kubernetes/resource_kubernetes_horizontal_pod_autoscaler_v2beta2_test.go
@@ -138,6 +138,10 @@ func TestAccKubernetesHorizontalPodAutoscalerV2Beta2_containerResource(t *testin
 		IDRefreshName:     "kubernetes_horizontal_pod_autoscaler_v2beta2.test",
 		ProviderFactories: testAccProviderFactories,
 		CheckDestroy:      testAccCheckKubernetesHorizontalPodAutoscalerDestroy,
+		ErrorCheck: func(err error) error {
+			t.Skipf("HPAContainerMetrics feature might not be enabled on the cluster and therefore this step will be skipped if an error occurs. Refer to the error for more details:\n%s", err)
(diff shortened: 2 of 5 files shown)


kubernetes/resource_kubernetes_horizontal_pod_autoscaler_v2beta2.go
```diff
@@ -53,6 +53,7 @@ func resourceKubernetesHorizontalPodAutoscalerV2Beta2() *schema.Resource {
 						"behavior": {
 							Type:        schema.TypeList,
 							Description: "Behavior configures the scaling behavior of the target in both Up and Down directions (scale_up and scale_down fields respectively).",
+							Computed:    true,
```
jrhouston, line 56: How come this needs to be computed?

arybolovlev: While I was working on this PR I noticed that terraform apply constantly produces a plan with changes related to the `behavior` attribute. Kubernetes uses default HPA Scaling rules if they are not set and the default values of these rules are part of the API response. That updates the state file with the `behavior` attribute after applying and causes the attribute deletion during the next plan.

Kubernetes documentation [link](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#horizontalpodautoscaler-v2beta2-autoscaling): _`behavior` ... If not set, the default HPAScalingRules for scale up and scale down are used._

You are right, it doesn't relate to the aim of this PR and shouldn't be here. Removed, thank you!

kubernetes/schema_horizontal_pod_autoscaler.go
```diff
@@ -194,7 +225,7 @@ func metricSpecFields() *schema.Resource {
 			"type": {
 				Type:        schema.TypeString,
 				Required:    true,
-				Description: `type is the type of metric source. It should be one of "Object", "Pods", "External" or "Resource", each mapping to a matching field in the object.`,
+				Description: `type is the type of metric source. It should be one of "ContainerResource", "External", "Object", "Pods" or "Resource", each mapping to a matching field in the object. Note: "ContainerResource" type is available on when the feature-gate HPAContainerMetrics is enabled`,
```
jrhouston, line 228: Let's add validation to this attribute to check that it's one of the values documented here. You can use [validation.StringInSlice](https://pkg.go.dev/github.com/hashicorp/terraform/helper/validation#StringInSlice)

kubernetes/resource_kubernetes_horizontal_pod_autoscaler_v2beta2_test.go
```diff
@@ -129,6 +129,48 @@ func TestAccKubernetesHorizontalPodAutoscalerV2Beta2_basic(t *testing.T) {
 	})
 }
 
+func TestAccKubernetesHorizontalPodAutoscalerV2Beta2_containerResource(t *testing.T) {
```
jrhouston, line 132: This test will fail on clusters where the `HPAContainerMetrics` feature gate is not enabled. Right now there isn't a reliable way to check if a feature gate has been enabled programmatically. You have to check the `--feature-gates` args supplied to the apiserver, which we can't even do on EKS, GKE, et al. 

For the moment I would add a `t.Skip()` to this test with a message explaining this. We can remove when the Featuer gate becomes defaulted to true.

jrhouston (changes requested):
Good work @arybolovlev, just a couple of things.

kubernetes/resource_kubernetes_horizontal_pod_autoscaler_v2beta2_test.go
arybolovlev: I have added `ErrorCheck` with `t.Skip()`. In this case, the step will be executed and skipped only if it fails.

PR #1661, opened 2022-03-25
Fix conversion of big.Float to float64 in kubernetes_manifest

alexsomesan:
### Description

This change adjusts the conversion from big.Float to float64 to not fail on precision differences.

In [terraform-plugin-go](https://github.com/hashicorp/terraform-plugin-go), a `tftypes.Number` value is represented internally as a Go standard lib `big.Float` which can be constructed in a few different ways. When constructed from a primitve float64 value, the conversion back to float64 happens with exact precision. However, when parsed from a string, using a precision higher than 53, the converstion back to float64 happens imprecisly and does not yield a `big.Exact` precision indicator. This is because primitive float64 values only go up to 53 bits precision, while big.Float can do better precision.

The logic that is handling `tftypes.Number` values in `kubernetes_manifest` was based on the assumption that the big.Float values obtained from Terraform via terraform-plugin-go would always convert exactly to float64. 

It turns out, terraform doesn't always pass a precision indicator into the msgpack encoded floats as seen here: https://github.com/hashicorp/terraform-plugin-go/blob/4f323c9cd630bcc2238434bbee9176418746d015/tftypes/value_msgpack.go#L108 In this case the conversion back to float64 will be lossy.

### Acceptance tests

```

...
```

### Release Note

```release-note
...
```

### References

Fixes https://github.com/hashicorp/terraform-provider-kubernetes/issues/1659

### Community Note

### DIFF
--- a/.github/workflows/check_examples.yaml
+++ b/.github/workflows/check_examples.yaml
@@ -20,7 +20,8 @@ jobs:
         terraform_version:
           - "0.14.10"
           - "0.15.5"
-          - "1.0.6"
+          - "1.0.11"
+          - "1.1.7"
     env:
       TF_X_KUBERNETES_MANIFEST_RESOURCE: 1
       KUBE_CONFIG_PATH: "~/.kube/config"
--- a/.github/workflows/manifest_acc.yaml
+++ b/.github/workflows/manifest_acc.yaml
@@ -23,7 +23,8 @@ jobs:
           # and there is no image available for the latest v1.19.x
           # - v1.19.11@sha256:07db187ae84b4b7de440a73886f008cf903fcf5764ba8106a9fd5243d6f32729
         terraform_version:
-          - 1.0.8
+          - 1.1.7
+          - 1.0.11
           - 0.15.5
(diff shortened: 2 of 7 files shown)


alexsomesan:
Need to figure out why is the test failing here. It passes when run locally.

jrhouston:
Looks like a Terraform version thing – it failed for me locally on Terraform 1.0 but passed when I upgraded to 1.1.

alexsomesan:
> Looks like a Terraform version thing – it failed for me locally on Terraform 1.0 but passed when I upgraded to 1.1.

Good find!
I did some version bisecting and found it's a manifestation of a TF bug that was fixed here: https://github.com/hashicorp/terraform/pull/29864

That fix was released in TF version 1.0.11, so everything before that would cause issues with this kind of floats.

joe-a-t:
Any update on this? Want to make sure it didn't get lost

alexsomesan:
This is ready for review.

jrhouston (approved):

PR #1663, opened 2022-03-26
Ignore service account volumes with kube-api-access prefix

vin01:
This is already fixed for `container` resources in https://github.com/hashicorp/terraform-provider-kubernetes/pull/1648/commits/ef15ac81a3cb316aca358cc8576a2b1432e2ffed. Eventually it might make sense to just use this prefix for filtering as it is updated in kubernetes itself: https://github.com/kubernetes/kubernetes/commit/1244ee66517bbe603d899bbd24458ebc0e185fd9


jrhouston:
> Eventually it might make sense to just use this prefix for filtering as it is updated in kubernetes itself: https://github.com/kubernetes/kubernetes/commit/1244ee66517bbe603d899bbd24458ebc0e185fd9

Agreed but for now we still have to support users on older cluster versions – so we'll need to wait a few versions before we can remove the old prefix.

jrhouston (approved):
Thanks for this fix @vin01

PR #1660, opened 2022-03-24
Add kubernetes_annotations resource

jrhouston:
### Description

This PR adds a `kubernetes_annotations` resource to go with `kubernetes_labels`. This is pretty much an exact copy of the implementation for `kubernetes_labels` but changed to annotations.

Closes #692 

### Acceptance tests
- [x] Have you added an acceptance test for the functionality being added?
- [x] Have you run the acceptance tests on this branch?

Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccXXX'

...
```

### Release Note
Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-kubernetes/blob/main/CHANGELOG.md):

```release-note
Add kubernetes_annotations resource
```

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #1671, opened 2022-03-30
Add kubernetes_config_map_v1_data resource

jrhouston:
### Description

This PR adds a resource which allows Terraform to use field manager to manage the data within a ConfigMap that already exists. The implementation of this resource is similar to kubernetes_annotations (#1660) and kubernetes_labels (#1616) but does not require the dynamic client, so does not need the user to specify the apiVersion and kind. 

### Acceptance tests

```
=== RUN   TestAccKubernetesConfigMapV1Data_basic
--- PASS: TestAccKubernetesConfigMapV1Data_basic (34.48s)
PASS
ok  	github.com/hashicorp/terraform-provider-kubernetes/kubernetes	35.445s
```

### Release Note

```release-note
Add kubernetes_config_map_v1_data resource
```

### References

#723 – one of the frequent use cases in this issue is patching configmaps  

### Community Note

### DIFF
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -299,8 +299,9 @@ func Provider() *schema.Provider {
 			"kubernetes_csi_driver_v1":    resourceKubernetesCSIDriverV1(),
 
 			// provider helper resources
-			"kubernetes_labels":      resourceKubernetesLabels(),
-			"kubernetes_annotations": resourceKubernetesAnnotations(),
+			"kubernetes_labels":             resourceKubernetesLabels(),
+			"kubernetes_annotations":        resourceKubernetesAnnotations(),
+			"kubernetes_config_map_v1_data": resourceKubernetesConfigMapV1Data(),
 		},
 	}
 
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -214,6 +214,7 @@ func Provider() *schema.Provider {
 			"kubernetes_default_service_account_v1": resourceKubernetesDefaultServiceAccount(),
 			"kubernetes_config_map":                 resourceKubernetesConfigMap(),
 			"kubernetes_config_map_v1":              resourceKubernetesConfigMap(),
+			"kubernetes_config_map_v1_data":         resourceKubernetesConfigMapV1Data(),
 			"kubernetes_secret":                     resourceKubernetesSecret(),
 			"kubernetes_secret_v1":                  resourceKubernetesSecret(),
 			"kubernetes_pod":                        resourceKubernetesPod(),
@@ -299,9 +300,8 @@ func Provider() *schema.Provider {
 			"kubernetes_csi_driver_v1":    resourceKubernetesCSIDriverV1(),
 
 			// provider helper resources
-			"kubernetes_labels":             resourceKubernetesLabels(),
-			"kubernetes_annotations":        resourceKubernetesAnnotations(),
-			"kubernetes_config_map_v1_data": resourceKubernetesConfigMapV1Data(),
+			"kubernetes_labels":      resourceKubernetesLabels(),
+			"kubernetes_annotations": resourceKubernetesAnnotations(),
 		},
 	}
 
--- a/kubernetes/resource_kubernetes_config_map_v1_data.go
+++ b/kubernetes/resource_kubernetes_config_map_v1_data.go
@@ -1,0 +1,213 @@
+package kubernetes
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func resourceKubernetesConfigMapV1Data() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceKubernetesConfigMapV1DataCreate,
+		ReadContext:   resourceKubernetesConfigMapV1DataRead,
+		UpdateContext: resourceKubernetesConfigMapV1DataUpdate,
+		DeleteContext: resourceKubernetesConfigMapV1DataDelete,
+		Schema: map[string]*schema.Schema{
+			"metadata": {
+				Type:     schema.TypeList,
+				Required: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:        schema.TypeString,
+							Description: "The name of the ConfigMap.",
+							Required:    true,
+							ForceNew:    true,
+						},
+						"namespace": {
+							Type:        schema.TypeString,
+							Description: "The namespace of the ConfigMap.",
+							Optional:    true,
+							ForceNew:    true,
+						},
+					},
+				},
+			},
+			"data": {
+				Type:        schema.TypeMap,
+				Description: "The data we want to add to the ConfigMap.",
+				Required:    true,
+			},
+			"force": {
+				Type:        schema.TypeBool,
+				Description: "Force overwriting data that is managed outside of Terraform.",
+				Optional:    true,
+			},
+		},
+	}
+}
+
+func resourceKubernetesConfigMapV1DataCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	metadata := expandMetadata(d.Get("metadata").([]interface{}))
+	d.SetId(buildId(metadata))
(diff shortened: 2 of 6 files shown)


kubernetes/resource_kubernetes_config_map_v1_data.go
```diff
@@ -0,0 +1,213 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
arybolovlev, line 13: `v1` and `metav1` seem to be a name of the same package that is imported twice and then below you use both names

kubernetes/provider.go
```diff
@@ -299,8 +299,9 @@ func Provider() *schema.Provider {
 			"kubernetes_csi_driver_v1":    resourceKubernetesCSIDriverV1(),
 
 			// provider helper resources
-			"kubernetes_labels":      resourceKubernetesLabels(),
-			"kubernetes_annotations": resourceKubernetesAnnotations(),
+			"kubernetes_labels":             resourceKubernetesLabels(),
+			"kubernetes_annotations":        resourceKubernetesAnnotations(),
+			"kubernetes_config_map_v1_data": resourceKubernetesConfigMapV1Data(),
```
arybolovlev, line 304: Most of our resources have a version at the end of their names, shall we continue this practice and call it `kubernetes_config_map_data_v1`? I am not sure though if that one won't overlap with any native Kubernetes resources in the future. I am totally fine with the current name and can see the logic here too, so I am just curious.

arybolovlev (commented):
Two minor comments.

kubernetes/resource_kubernetes_config_map_v1_data.go
jrhouston: Yep you're right, did this by accident.

aidanmelen:
🥇

aidanmelen:
Since this PR is aimed at solving #723, I ran a few test runs patching the `aws-auth` configmap that was automatically generated by an AWS EKS cluster. That test run was successful 🥳 

```hcl

  # module.eks_auth.kubernetes_config_map_v1_data.aws_auth will be created
  + resource "kubernetes_config_map_v1_data" "aws_auth" {
      + data  = {
          + "mapAccounts" = <<-EOT
                - "777777777777"
                - "888888888888"
            EOT
          + "mapRoles"    = <<-EOT
                - "groups":
                  - "system:bootstrappers"
                  - "system:nodes"
                  "rolearn": "arn:aws:iam::326504147071:role/foo-eks-node-group-20220401053147969300000001"
                  "username": "system:node:{{EC2PrivateDNSName}}"
                - "groups":
                  - "system:bootstrappers"
                  - "system:nodes"
                  "rolearn": "arn:aws:iam::326504147071:role/boo-node-group-20220401053147969900000003"
                  "username": "system:node:{{EC2PrivateDNSName}}"
                - "groups":
                  - "system:bootstrappers"
                  - "system:nodes"
                  - "system:node-proxier"
                  "rolearn": "arn:aws:iam::326504147071:role/bar-20220401053147969700000002"
                  "username": "system:node:{{SessionName}}"
                - "groups":
                  - "system:masters"
                  "rolearn": "arn:aws:iam::66666666666:role/role1"
                  "username": "role1"
            EOT
          + "mapUsers"    = <<-EOT
                - "groups":
                  - "system:masters"
                  "userarn": "arn:aws:iam::66666666666:user/user1"
                  "username": "user1"
                - "groups":
                  - "system:masters"
                  "userarn": "arn:aws:iam::66666666666:user/user2"
                  "username": "user2"
            EOT
        }
      + force = true
      + id    = (known after apply)

      + metadata {
          + name      = "aws-auth"
          + namespace = "kube-system"
        }
    }

  Enter a value: yes

module.eks_auth.kubernetes_config_map_v1_data.aws_auth: Creating...
module.eks_auth.kubernetes_config_map_v1_data.aws_auth: Creation complete after 1s [id=kube-system/aws-auth]

```

However, this resource will fail if the configmap does not exist. I realize that is by design... But please consider this issue (https://github.com/aidanmelen/terraform-aws-eks-auth/issues/14). Another scenario where the `aws-auth` configmap would not exist is when creating an AWS EKS cluster without a data plane as the `aws-auth` configmap is only created when managed eks node groups or fargate profiles are added to the cluster. For this solution to work for those scenarios, I think it would be nice to have this resource also create the configmap if it does not exist with the fields provided OR patch the fields in-place when the resource already exists.

aidanmelen:
This resource will fail if the configmap does not exist. I realize that is by design... But please consider that the `aws-auth` configmap is only created when managed eks node groups or fargate profiles are added to the cluster. As a result, there are at least two scenarios where the `aws-auth` configmap will not exist on cluster creation. For example:

1. A [cluster with self managed node groups](https://github.com/aidanmelen/terraform-aws-eks-auth/issues/14) will not automatically create an `aws-auth` configmap.

2. An AWS EKS cluster that is deployed without a data plane will not automatically create an `aws-auth` configmap.

I will admit that these are edge case scenarios, but think it would be nice to have this resource also create the configmap if it does not exist with the fields provided OR patch the fields in-place when the resource already exists. That is what I was trying to merge with this [PR](https://github.com/hashicorp/terraform-provider-kubernetes/pull/1667) at least.

I tried getting around this by using the `kubernetes_resource` data resource to check if the `aws-auth` configmap exists on creation. For example:

```hcl
data "kubernetes_resource" "aws_auth" {
  api_version = "v1"
  kind        = "ConfigMap"

  metadata {
    name      = "aws-auth"
    namespace = "kube-system"
  }
}

output "aws_auth_cm_data" {
  value = try(data.kubernetes_resource.aws_auth, null)
}
```

but then i encounter this issue with the provider

```hcl
╷
│ Error: Provider produced null object
│ 
│ Provider "provider[\"registry.terraform.io/hashicorp/kubernetes\"]" produced a null value for
│ data.kubernetes_resource.aws_auth.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

again, I understand that is by design as it is consistent with other data resources like [data.aws_ami](https://registry.terraform.io/providers/hashicorp/aws/3.68.0/docs/data-sources/ami). But I think it would be nice if the data resource would return a `null` value when a resource does not exist so that the user can handle the missing resource rather than having the provider throw an error. I think this would create a sane work around for the issue described above.

aidanmelen:
I suppose another workaround could be to indirectly check for the existence of `aws-auth` by checking the managed node groups or fargate profile roles outputs from the eks module. If any exist, then use the `kubernetes_config_map_v1_data` resource. Otherwise, use the `kubernetes_manifest` data resource to create the configmap with `field_manager.force_conflicts = true`.

jrhouston:
Thanks for testing this out @aidanmelen and for the great discussion. 

Throwing an error when the data source doesn't exist is pretty standard across terraform providers and so I would prefer to stay consistent with that pattern.

However, I do have plans (soon) to add a `kubernetes_resources` resource that would be a generic way to query for a list of resources like kubectl's `--field-selector` and `--label-selector`, which would return an empty list if nothing is found. I think that's something you could use to work around the use-case you've described above.

aidanmelen:
> Throwing an error when the data source doesn't exist is pretty standard across terraform providers and so I would prefer to stay consistent with that pattern.

It is unfortunate that Terraform took that approach for data resource. It is best to stay consistent though.

aidanmelen:
> However, I do have plans (soon) to add a `kubernetes_resources` resource that would be a generic way to query for a list of resources like kubectl's `--field-selector` and `--label-selector`, which would return an empty list if nothing is found. I think that's something you could use to work around the use-case you've described above.

I like that a lot. I vote yes. It would be really similiar to the pattern used by the [kubectl_server_version](https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs/resources/kubectl_server_version) resource in kubectl provider.

kubernetes/resource_kubernetes_config_map_v1_data.go
```diff
@@ -0,0 +1,213 @@
...
+				Description: "Force overwriting data that is managed outside of Terraform.",
+				Optional:    true,
+			},
+		},
+	}
+}
+
+func resourceKubernetesConfigMapV1DataCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	metadata := expandMetadata(d.Get("metadata").([]interface{}))
+	d.SetId(buildId(metadata))
```
arybolovlev, line 62: The function `buildId` sets ID as `metadata.namespace/metadata.name`, however, in this case, if `namespace` is omitted as optional value the ID will be set to `/metadata.name`. Can we make "default" a default value in the schema `metadata.namespace`? If that works, then we can rid of conditions that check whether the `namespace` attribute is set up in read and update functions.

arybolovlev (commented):
One minor comment. The rest looks great and works as expected! 🚀 👍🏻

kubernetes/resource_kubernetes_config_map_v1_data.go
jrhouston: Good call – I've updated the schema.

arybolovlev (approved):
Awesome! 🚀

PR #1687, opened 2022-04-12
Add new provider-level attributes `ignore_annotations` and `ignore_labels`

arybolovlev:
### Description

The main goal of this PR is to add new provider-level attributes `ignore_annotations` and `ignore_labels` to ignore annotations and labels that are managed by external systems. Both attributes support RegExp to match metadata objects more effectively.

In addition to that, unused test functions were removed. They are related to the mentioned changes.

### Acceptance tests

```
$ make testacc TESTARGS='-run=TestAccKubernetesIgnoreKubernetesMetadata_basic'

=== RUN   TestAccKubernetesIgnoreKubernetesMetadata_basic
--- PASS: TestAccKubernetesIgnoreKubernetesMetadata_basic (1.86s)
PASS
ok  	github.com/hashicorp/terraform-provider-kubernetes/kubernetes	2.386s
```

### Release Note

```release-note
Add a new provider level attribute `ignore_annotations`
Add a new provider level attribute `ignore_labels`
```

### Community Note


kubernetes/resource_kubernetes_namespace_test.go
```diff
@@ -213,52 +201,6 @@ func TestAccKubernetesNamespace_deleteTimeout(t *testing.T) {
 	})
 }
 
-func testAccCheckMetaAnnotations(om *metav1.ObjectMeta, expected map[string]string) resource.TestCheckFunc {
```
arybolovlev, line 216: This is an unused function.

kubernetes/structure_cron_job.go
```diff
@@ -5,7 +5,7 @@ import (
 	"k8s.io/api/batch/v1beta1"
 )
 
-func flattenCronJobSpec(in v1beta1.CronJobSpec, d *schema.ResourceData) ([]interface{}, error) {
+func flattenCronJobSpec(in v1beta1.CronJobSpec, d *schema.ResourceData, providerMetadata interface{}) ([]interface{}, error) {
```
arybolovlev, line 8: In most functions, we do have a `meta` argument that represents provider metadata data, but in some cases, this name is used to store Kubernetes object metadata. In such cases, I gave a new name to the provider metadata argument.
Shall I leave it that or stick with `meta` for the provider meta and rename Kubernetes object meta to something like `objectMeta` to avoid confusion?

manifest/provider/provider_config.go
```diff
@@ -164,6 +164,28 @@ func GetProviderConfigSchema() *tfprotov5.Schema {
 				DescriptionKind: 0,
 				Deprecated:      false,
 			},
+			{
+				Name:            "ignore_annotations",
```
arybolovlev, line 168: For now `ignore_annotations` is a list of annotations that should be ignored across all created resources, the usage  looks like this:
```hcl
provider "kubernetes" {
  ignore_annotations = [
    "cni.projectcalico.org/podIP",
    "cni.projectcalico.org/podIPs",
    "terraform.io/provider",
  ]
}
```

The provider goes through the list of annotations and does an exact match. Do we want to leave it like this or possibly implement some sort of wildcard mask support? For instance, `keys` -- to cover the above case scenario; `key_prefixes` -- to have the ability to capture all annotations that start with what we have listed here. These are all just possible options. The usage will be something like this(rewrote above example):

```hcl
provider "kubernetes" {
  ignore_annotations {
    keys = [
      "terraform.io/provider",
    ]

    key_prefixes = [
      "cni.projectcalico.org/",
    ]
  }
}
```

arybolovlev: After the internal discussion we had, I've updated the code to support RegExp expressions as `ignore_annotations` and `ignore_labels` arguments. With RegExp, we cover a lot of potential uses cases, including exact match, prefix and suffix match, and much more.

kubernetes/structure_cron_job.go
jrhouston: I think stick with `meta` for the moment just because it's so pervasive.

kubernetes/structures.go
```diff
@@ -118,19 +119,31 @@ func expandStringSlice(s []interface{}) []string {
...
 	}
 	configAnnotations := d.Get(prefix + "metadata.0.annotations").(map[string]interface{})
-	m["annotations"] = removeInternalKeys(meta.Annotations, configAnnotations)
+
+	var ignoreAnnotations []string
+	if v, ok := providerMetadata.(kubeClientsets).configData.Get("ignore_annotations").([]interface{}); ok {
+		ignoreAnnotations = expandStringSlice(v)
+	}
+
+	m["annotations"] = removeInternalKeys(meta.Annotations, configAnnotations, ignoreAnnotations)
```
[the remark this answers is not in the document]
arybolovlev: Thanks, @jrhouston. Done!

I reverted function names back to their origins and added a new one `removeKeys` that works in the same way as `removeInternalKeys` but accepts one extra argument -- Kubernetes metadata(annotations and labels) keys that a practitioner wants to ignore. `removeKeys` also calls a new function `isIgnoteKey` instead of `isInternalKey` that verifies whether a Kubernetes object has metadata keys that should be ignored.

`removeKeys` seems to be more generic and actually can cover use cases of `removeInternalKeys`. Maybe we should think about just one function, but call it two times. The first time, to ignore "internal" keys that we can store in a constant as a list of regular expressions. The second time, to ignore keys that the practitioner wants to ignore.

kubernetes/structures.go
```diff
@@ -192,6 +218,18 @@ func isInternalKey(annotationKey string) bool {
 	return false
 }
 
+// This function verifies if a given Kubernetes metadata(annotations and labels) key
+// is among the keys that need to be ignored
+func isIgnoteKey(kubernetesMetadataKey string, ignoreKubernetesMetadata []string) bool {
```
jrhouston, line 223: Looks like a typo – should be `ignoreKey`? I might even drop the `is` here and just call this `ignoreKey()`

jrhouston: I also wouldn't be so explicit with the naming of the parameters, in principle this function is quite generic. It takes a string, and a slice of strings to be parsed as regular expressions and returns true if any of them match. So you could call the parameters `key` and `expressions`. I would also mention in the comment above that the slice should contain regular expressions.

kubernetes/structures.go
```diff
@@ -192,6 +218,18 @@ func isInternalKey(annotationKey string) bool {
...
+	for _, im := range ignoreKubernetesMetadata {
+		if ok, _ := regexp.MatchString(im, kubernetesMetadataKey); ok {
```
jrhouston, line 225: We are suppressing the error here, but it is actually important for us here to capture if the regular expression failed to parse. If the user supplies an invalid regex then this is going to fail silently.

jrhouston (changes requested):
Feels cleaner with these changes, just a couple of things I spotted.

kubernetes/structures.go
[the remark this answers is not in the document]
jrhouston: Thanks @arybolovlev. I agree, we could take the logic from `isInternalKey` and turn it into a const of regular expressions that we can then merge with the user supplied list to pass into a single `removeKeys()` call. Let's leave that for another PR though.

kubernetes/structures.go
arybolovlev: Great catch, @jrhouston!

I have added provider level schema validation `validation.StringIsValidRegExp` for both new attributes. In that case, the practitioners will get an error as early as possible if they use invalid regular expressions. Since `validation.StringIsValidRegExp` and `regexp.MatchString` use the same RegExp validation mechanism(they simply compile the RegExp and throw an error if it is incorrect), we are safe to ignore errors at this stage.

I have also addressed other comments. 👍 

Thank you! 😊

kubernetes/structures.go
arybolovlev: I like this suggestion, done!

jrhouston (approved):
Thanks for adding this @arybolovlev – works good 😄

PR #1697, opened 2022-04-21
Fix cluster role aggregation rules selectors

### LINKED ISSUE #1360: Multiple cluster_role_selectors are ignored
```
Terraform version: 1.0.0
Kubernetes provider version: 2.3.2
Kubernetes version: 1.18 (EKS)
```

### Affected Resource(s)
```
resource "kubernetes_cluster_role" "example" {

  metadata {
    name   = "example"
  }

  aggregation_rule {
    cluster_role_selectors {
      match_labels = {
        "rbac.authorization.k8s.io/aggregate-to-view" = "true"
      }
    }
    cluster_role_selectors {
      match_labels = {
        "example.io/aggregregate-to-app" = "true"
      }
    }
  }
}
````

### Expected Behavior
A role with the following aggregation rule is created:

```
aggregationRule:
  clusterRoleSelectors:
    - matchLabels:
        rbac.authorization.k8s.io/aggregate-to-view: 'true'
    - matchLabels:
        example.io/aggregregate-to-app: 'true'
```
### Actual Behavior
The second cluster_role_selectors is ignored and only the first match_label rule is created (the same happens using `match_expressions` )
```
aggregationRule:
  clusterRoleSelectors:
    - matchLabels:
        rbac.authorization.k8s.io/aggregate-to-view: 'true'
```

### References

- https://github.com/hashicorp/terraform-provider-kubernetes/issues/1303
- https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/cluster_role#aggregation_rule

### PR DESCRIPTION
**Pull request by stevehipwell**
### Description

This PR fixes the behaviour when adding multiple `cluster_role_selectors` to the `kubernetes_cluster_role.aggregation_rule` so that all are processed rather than only the first one.

### Acceptance tests

I'm on Windows and the make configuration doesn't look to be multi platform, I have WSL but my Go toolchain isn't up to date (VPN issues) so I ran the relevant tests manually and they passed.

### Release Note

```release-note
Fix not handling multiple `cluster_role_selectors` on a `kubernetes_cluster_role`
```

### References

Closes #1360.

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_cluster_role_test.go
+++ b/kubernetes/resource_kubernetes_cluster_role_test.go
@@ -182,6 +182,46 @@ func TestAccKubernetesClusterRole_aggregationRuleBasic(t *testing.T) {
 	})
 }
 
+func TestAccKubernetesClusterRole_aggregationRuleMultiple(t *testing.T) {
+	var conf api.ClusterRole
+	resourceName := "kubernetes_cluster_role.test"
+	name := acctest.RandomWithPrefix("tf-acc-test")
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		IDRefreshName:     resourceName,
+		ProviderFactories: testAccProviderFactories,
+		CheckDestroy:      testAccCheckKubernetesClusterRoleDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccKubernetesClusterRoleConfig_aggRuleMultiple(name),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckKubernetesClusterRoleExists(resourceName, &conf),
+					resource.TestCheckResourceAttr(resourceName, "aggregation_rule.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "aggregation_rule.0.cluster_role_selectors.#", "2"),
+					resource.TestCheckResourceAttr(resourceName, "aggregation_rule.0.cluster_role_selectors.0.match_labels.foo", "bar"),
+					resource.TestCheckResourceAttr(resourceName, "aggregation_rule.0.cluster_role_selectors.1.match_labels.bar", "foo"),
+				),
+			},
+			{
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"metadata.0.resource_version"},
+			},
+			{
+				Config: testAccKubernetesClusterRoleConfig_aggRuleMultipleModified(name),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckKubernetesClusterRoleExists(resourceName, &conf),
+					resource.TestCheckResourceAttr(resourceName, "aggregation_rule.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "aggregation_rule.0.cluster_role_selectors.#", "2"),
+					resource.TestCheckResourceAttr(resourceName, "aggregation_rule.0.cluster_role_selectors.0.match_labels.bar", "foo"),
+					resource.TestCheckResourceAttr(resourceName, "aggregation_rule.0.cluster_role_selectors.1.match_labels.foo", "bar"),
+				),
+			},
+		},
+	})
+}
+
 func TestAccKubernetesClusterRole_aggregationRuleRuleAggregation(t *testing.T) {
 	var conf api.ClusterRole
 	resourceName := "kubernetes_cluster_role.test"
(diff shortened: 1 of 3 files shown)


kubernetes/structures_rbac.go
```diff
@@ -89,7 +89,9 @@ func expandClusterRoleAggregationRule(in []interface{}) *api.AggregationRule {
 
 	if v, ok := m["cluster_role_selectors"].([]interface{}); ok && len(v) > 0 {
 		crs := make([]metav1.LabelSelector, 0)
```
arybolovlev, line 91: Small possible optimization here. The length of slice `crs` will be equal to the length of slice `v`. What would you say if we pass `len(v)` as length and capacity when making `crs` and then within the for loop use indexes to assign values instead of `append`?

arybolovlev (commented):
Hi @stevehipwell,

Thank you for your contribution! Just one minor comment and request. Could you please also update the acceptance test `TestAccKubernetesClusterRole_aggregationRuleBasic` to test your changes? Just add one more `match_labels` `bar = "foo"` and that should be enougth.

kubernetes/structures_rbac.go
stevehipwell: That was my initial intention, but I looked at the rest of the code and copied the pattern most used to keep it consistent. I'll make this change.

stevehipwell: RE acceptance test - Is it multi pass or do we need tests for each case? Off the top of my head there are a minimum of 5 test scenarios to prove that the aggregation is working correctly.

stevehipwell:
@arybolovlev I've made the slice length change and added a new acceptance test `TestAccKubernetesClusterRole_aggregationRuleMultiple` to make sure that supporting multiple rule selectors doesn't break supporting only a single rule selector.

stevehipwell:
@arybolovlev could we get this progressed?

kubernetes/resource_kubernetes_cluster_role_test.go
```diff
@@ -182,6 +182,52 @@ func TestAccKubernetesClusterRole_aggregationRuleBasic(t *testing.T) {
...
+	resourceName := "kubernetes_cluster_role.test"
+	name := acctest.RandomWithPrefix("tf-acc-test")
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		IDRefreshName:     resourceName,
+		ProviderFactories: testAccProviderFactories,
+		CheckDestroy:      testAccCheckKubernetesClusterRoleDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccKubernetesClusterRoleConfig_aggRule(name),
```
arybolovlev, line 196: Thank you for adding the acceptance test, however, this one is failing:
```
=== RUN   TestAccKubernetesClusterRole_aggregationRuleMultiple
    resource_kubernetes_cluster_role_test.go:189: Step 1/2 error: Check failed: 2 errors occurred:
                * Check 3/5 error: kubernetes_cluster_role.test: Attribute 'aggregation_rule.0.cluster_role_selectors.#' expected "2", got "1"
                * Check 5/5 error: kubernetes_cluster_role.test: Attribute 'aggregation_rule.0.cluster_role_selectors.1.match_labels.bar' not found
```

As a configuration here you use `testAccKubernetesClusterRoleConfig_aggRule` and it has only one `cluster_role_selectors`. Since this config is used in some other tests, let's create a new one `testAccKubernetesClusterRoleConfig_aggRuleMultiple` with a simplified config in there and use it instead. Something like this:

```
func testAccKubernetesClusterRoleConfig_aggRuleMultiple(name string) string {
	return fmt.Sprintf(`resource "kubernetes_cluster_role" "test" {
  metadata {
    name = "%s"
  }

  aggregation_rule {
    cluster_role_selectors {
      match_labels = {
        foo = "bar"
      }
    }
    cluster_role_selectors {
      match_labels = {
        bar = "foo"
      }
    }  
  }
}
`, name)
}
```

In that case, we can also take out the third step from the test and has fewer checks in the first one. Something like this:
```
resource.TestCheckResourceAttr(resourceName, "aggregation_rule.#", "1"),
resource.TestCheckResourceAttr(resourceName, "aggregation_rule.0.cluster_role_selectors.#", "2"),
resource.TestCheckResourceAttr(resourceName, "aggregation_rule.0.cluster_role_selectors.0.match_labels.foo", "bar"),
resource.TestCheckResourceAttr(resourceName, "aggregation_rule.0.cluster_role_selectors.1.match_labels.bar", "foo"),
```

And thank you for taking care of this issue!

arybolovlev (commented):
@stevehipwell, thank you for taking care of this issue. One small comment regarding the acceptance test and we are good to go!

kubernetes/resource_kubernetes_cluster_role_test.go
stevehipwell: Thanks for the feedback @arybolovlev, I'm not sure why it ever passed locally; I've updated it now with two new configurations, one for create and another for modify.

arybolovlev (approved):
Looks good! Thank you for your contribution @stevehipwell! 👍🏻 🚀

This fix will be available in the upcoming release.

stevehipwell:
@arybolovlev did this make it into `v2.11.0`?

arybolovlev:
> @arybolovlev did this make it into `v2.11.0`?

Your fix is a part of `v2.11.0`, I have updated the changelog file and the release description. Thanks!

PR #1700, opened 2022-04-25
Add new data source and resource attributes of `kubernetes_service(_v1)`

arybolovlev:
### Description

This PR adds the following new attributes in `kubernetes_service(_v1)` and `data_kubernetes_service`:
- `allocate_load_balancer_node_ports`
- `cluster_ips`
- `internal_traffic_policy`
- `load_balancer_class`
- `session_affinity_config`

### Acceptance tests

```
make testacc TESTARGS='-run TestAccKubernetesService_basic'
...
=== RUN   TestAccKubernetesService_basic
--- PASS: TestAccKubernetesService_basic (17.51s)
PASS

make testacc TESTARGS='-run TestAccKubernetesService_loadBalancer_internal_traffic_policy'
...
=== RUN   TestAccKubernetesService_loadBalancer_internal_traffic_policy
--- PASS: TestAccKubernetesService_loadBalancer_internal_traffic_policy (26.84s)
PASS

make testacc TESTARGS='-run TestAccKubernetesService_loadBalancer_class'
...
=== RUN   TestAccKubernetesService_loadBalancer_class
--- PASS: TestAccKubernetesService_loadBalancer_class (7.43s)
PASS


make testacc TESTARGS='-run TestAccKubernetesService_nodePort'
=== RUN   TestAccKubernetesService_nodePort
--- PASS: TestAccKubernetesService_nodePort (11.80s)
PASS


make testacc TESTARGS='-run TestAccKubernetesDataSourceService_basic'
...
=== RUN   TestAccKubernetesDataSourceService_basic
--- PASS: TestAccKubernetesDataSourceService_basic (12.25s)
PASS
```

### Release Note
```release-note
Add new data source and resource attributes of `kubernetes_service(_v1)`
```

### References

Closes:
- https://github.com/hashicorp/terraform-provider-kubernetes/issues/1683

### Community Note


kubernetes/resource_kubernetes_service.go
```diff
@@ -198,6 +235,34 @@ func resourceKubernetesServiceSchemaV1() map[string]*schema.Schema {
...
+									Type:        schema.TypeList,
+									Description: "",
+									Optional:    true,
+									Computed:    true,
+									MaxItems:    1,
+									Elem: &schema.Resource{
+										Schema: map[string]*schema.Schema{
+											"timeout_seconds": {
+												Type:        schema.TypeInt,
+												Description: "",
```
jrhouston, line 256: There is a description for this field: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#clientipconfig-v1-core

Looks like we might be able to add some validation here too, or at least validate that it's non-negative.

kubernetes/structure_service_spec.go
```diff
@@ -34,6 +34,22 @@ func flattenIPFamilies(in []v1.IPFamily) []interface{} {
...
+func flattenSessionAffinityConfigClientIP(in v1.ClientIPConfig) []interface{} {
+	a := make([]interface{}, 0, 0)
+	att := make(map[string]interface{})
+	att["timeout_seconds"] = in.TimeoutSeconds
+	a = append(a, att)
+	return a
+}
+
+func flattenSessionAffinityConfig(in v1.SessionAffinityConfig) []interface{} {
+	a := make([]interface{}, 0, 0)
```
jrhouston, line 46: Because there's only one attribute here we could just write this as one return statement instead of using these `make` and `append` calls, like this:

```golang
return []interface{}{
	map[string]interface{}{
		"client_ip": flattenSessionAffinityConfigClientIP(*in.ClientIP),
	},
}
```

could also do the same in `flattenSessionAffinityConfigClientIP` above.

kubernetes/structure_service_spec.go
```diff
@@ -138,6 +177,30 @@ func expandIPFamilies(l []interface{}) []v1.IPFamily {
 	return obj
 }
 
+func expandSessionAffinityConfigClientIP(l []interface{}) *v1.ClientIPConfig {
+	var to int32
```
jrhouston, line 181: you need to add a guard here to check that the length isn't 0 and the first item isn't null, like this: 

```golang
if len(l) == 0 || l[0] == nil {
	return &v1.ClientIPConfig{}
}
```

It's an edge case but I was able to crash the provider by specifying this block but omitting `timeout_seconds`.

kubernetes/structure_service_spec.go
```diff
@@ -138,6 +177,30 @@ func expandIPFamilies(l []interface{}) []v1.IPFamily {
...
+	in := l[0].(map[string]interface{})
+	if v, ok := in["timeout_seconds"].(int); ok {
+		to = int32(v)
+	}
+	return &v1.ClientIPConfig{TimeoutSeconds: &to}
```
jrhouston, line 186: I would put this statement as a return inside the if statement as above, an then return `&v1.ClientIPConfig{}` here as the default case.

jrhouston (changes requested):
Thanks for adding these attributes @arybolovlev. Noticed a couple things.

kubernetes/resource_kubernetes_service.go
arybolovlev: Good catch, I have added descriptions and the following validation `validation.IntBetween(1, 86400)`.

kubernetes/structure_service_spec.go
arybolovlev: Like this one, thank you!

kubernetes/structure_service_spec.go
arybolovlev: Good catch, thank you!

kubernetes/structure_service_spec.go
arybolovlev: Thanks, @jrhouston! I like this suggestion, I've made some optimizations here.

jrhouston (approved):
Nice, thanks @arybolovlev!

PR #1792, opened 2022-07-26
[TK-1373] Update resource and data of 'kubernetes_(default_)service_account' to handle deprecated 'default_secret_name' in Kubernetes 1.24.0+

arybolovlev:
### Description

Starting from the version `1.24.0` Kubernetes does not generate a token for a service account by default([link](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes) to the change log). In this case, the concept of `default_secret_name` is not applicable anymore for the Kubernetes clusters `v1.24.0+`. This PR updates the provider's behavior accordingly to the cluster version.

#### Affected resource
- `r/kubernetes_default_service_account`
- `d/kubernetes_service_account`
- `r/kubernetes_service_account`

### Acceptance tests

`1.24.0+` [1.24.2-gke.300]
```
$ make testacc TESTARGS='-run ^.*ServiceAccount.*'

=== RUN   TestAccKubernetesDataSourceServiceAccount_basic
--- PASS: TestAccKubernetesDataSourceServiceAccount_basic (10.21s)
=== RUN   TestAccKubernetesDataSourceServiceAccount_default_secret
    provider_test.go:225: This test does not run on cluster versions 1.24.0 and above
--- SKIP: TestAccKubernetesDataSourceServiceAccount_default_secret (0.06s)
=== RUN   TestAccKubernetesDefaultServiceAccount_basic
--- PASS: TestAccKubernetesDefaultServiceAccount_basic (13.19s)
=== RUN   TestAccKubernetesDefaultServiceAccount_secrets
--- PASS: TestAccKubernetesDefaultServiceAccount_secrets (12.79s)
=== RUN   TestAccKubernetesDefaultServiceAccount_automountServiceAccountToken
--- PASS: TestAccKubernetesDefaultServiceAccount_automountServiceAccountToken (13.20s)
=== RUN   TestAccKubernetesServiceAccount_basic
--- PASS: TestAccKubernetesServiceAccount_basic (6.58s)
=== RUN   TestAccKubernetesServiceAccount_default_secret
    provider_test.go:225: This test does not run on cluster versions 1.24.0 and above
--- SKIP: TestAccKubernetesServiceAccount_default_secret (0.06s)
=== RUN   TestAccKubernetesServiceAccount_automount
--- PASS: TestAccKubernetesServiceAccount_automount (5.64s)
=== RUN   TestAccKubernetesServiceAccount_update
--- PASS: TestAccKubernetesServiceAccount_update (13.58s)
=== RUN   TestAccKubernetesServiceAccount_generatedName
--- PASS: TestAccKubernetesServiceAccount_generatedName (5.08s)
PASS
```

`pre-1.24.0` [1.23.8-gke.400]
```
$ make testacc TESTARGS='-run ^.*ServiceAccount.*'

=== RUN   TestAccKubernetesDataSourceServiceAccount_basic
--- PASS: TestAccKubernetesDataSourceServiceAccount_basic (8.83s)
=== RUN   TestAccKubernetesDataSourceServiceAccount_default_secret
--- PASS: TestAccKubernetesDataSourceServiceAccount_default_secret (8.08s)
=== RUN   TestAccKubernetesDefaultServiceAccount_basic
--- PASS: TestAccKubernetesDefaultServiceAccount_basic (12.[30](https://github.com/hashicorp/terraform-provider-kubernetes/runs/7520903353?check_suite_focus=true#step:11:31)s)
=== RUN   TestAccKubernetesDefaultServiceAccount_secrets
--- PASS: TestAccKubernetesDefaultServiceAccount_secrets (11.76s)
=== RUN   TestAccKubernetesDefaultServiceAccount_automountServiceAccountToken
--- PASS: TestAccKubernetesDefaultServiceAccount_automountServiceAccountToken (12.11s)
=== RUN   TestAccKubernetesServiceAccount_basic
--- PASS: TestAccKubernetesServiceAccount_basic (5.86s)
=== RUN   TestAccKubernetesServiceAccount_default_secret
--- PASS: TestAccKubernetesServiceAccount_default_secret (5.[36](https://github.com/hashicorp/terraform-provider-kubernetes/runs/7520903353?check_suite_focus=true#step:11:37)s)
=== RUN   TestAccKubernetesServiceAccount_automount
--- PASS: TestAccKubernetesServiceAccount_automount (5.21s)
=== RUN   TestAccKubernetesServiceAccount_update
--- PASS: TestAccKubernetesServiceAccount_update (12.14s)
=== RUN   TestAccKubernetesServiceAccount_generatedName
--- PASS: TestAccKubernetesServiceAccount_generatedName (4.81s)
PASS
```

### Release Note

```release-note
...
```

### References
Fixes: [1724](https://github.com/hashicorp/terraform-provider-kubernetes/issues/1724)

### Community Note

### DIFF
--- a/kubernetes/provider.go
+++ b/kubernetes/provider.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 
 	"github.com/hashicorp/go-cty/cty"
+	gversion "github.com/hashicorp/go-version"
 	"github.com/mitchellh/go-homedir"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -632,4 +633,27 @@ func useAdmissionregistrationV1beta1(conn *kubernetes.Clientset) (bool, error) {
 	log.Printf("[INFO] Using %s/v1beta1", group)
 	useadmissionregistrationv1beta1 = ptrToBool(true)
 	return true, nil
+}
+
+func getServerVersion(connection *kubernetes.Clientset) (*gversion.Version, error) {
+	sv, err := connection.ServerVersion()
+	if err != nil {
+		return nil, err
+	}
+
+	return gversion.NewVersion(sv.String())
+}
+
+func serverVersionGreaterThanOrEqual(connection *kubernetes.Clientset, version string) (bool, error) {
+	sv, err := getServerVersion(connection)
+	if err != nil {
+		return false, err
+	}
+	// server version that we need to compare with
+	cv, err := gversion.NewVersion(version)
+	if err != nil {
+		return false, err
+	}
+
+	return sv.GreaterThanOrEqual(cv), nil
 }
--- a/kubernetes/resource_kubernetes_default_service_account.go
+++ b/kubernetes/resource_kubernetes_default_service_account.go
@@ -55,7 +55,6 @@ func resourceKubernetesDefaultServiceAccountCreate(ctx context.Context, d *schem
 		log.Printf("[INFO] Default service account exists: %s", metadata.Namespace)
 		return nil
 	})
-
 	if err != nil {
 		return diag.FromErr(err)
 	}
--- a/kubernetes/resource_kubernetes_service_account.go
+++ b/kubernetes/resource_kubernetes_service_account.go
@@ -69,8 +69,9 @@ func resourceKubernetesServiceAccount() *schema.Resource {
 				Default:     true,
 			},
 			"default_secret_name": {
-				Type:     schema.TypeString,
-				Computed: true,
+				Type:       schema.TypeString,
+				Computed:   true,
+				Deprecated: "Starting from version 1.24.0 Kubernetes does not automatically generate a token for service accounts, in this case, `default_secret_name` will be empty and deprecated in future",
 			},
 		},
 	}
@@ -106,12 +107,21 @@ func resourceKubernetesServiceAccountCreate(ctx context.Context, d *schema.Resou
 	if err != nil {
 		return diag.FromErr(err)
 	}
+
 	return resourceKubernetesServiceAccountRead(ctx, d, meta)
 }
 
 func getServiceAccountDefaultSecret(ctx context.Context, name string, config api.ServiceAccount, timeout time.Duration, conn *kubernetes.Clientset) (*api.Secret, error) {
+	b, err := serverVersionGreaterThanOrEqual(conn, "1.24.0")
+	if err != nil {
+		return &api.Secret{}, err
+	}
+	if b {
+		return &api.Secret{}, nil
+	}
+
 	var svcAccTokens []api.Secret
-	err := resource.RetryContext(ctx, timeout, func() *resource.RetryError {
+	err = resource.RetryContext(ctx, timeout, func() *resource.RetryError {
 		resp, err := conn.CoreV1().ServiceAccounts(config.Namespace).Get(ctx, name, metav1.GetOptions{})
 		if err != nil {
 			return resource.NonRetryableError(err)
@@ -166,6 +176,20 @@ func findDefaultServiceAccount(ctx context.Context, sa *api.ServiceAccount, conn
 	   https://github.com/kubernetes/kubernetes/blob/release-1.13/pkg/controller/serviceaccount/tokens_controller.go#L384
 	*/
 	ds := make([]string, 0)
+
+	b, err := serverVersionGreaterThanOrEqual(conn, "1.24.0")
+	if err != nil {
+		return "", diag.FromErr(err)
+	}
+	if b {
+		return "", diag.Diagnostics{
+			diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary:  "'default_secret_name' is no longer applicable for Kubernetes 'v1.24.0' and above",
+				Detail:   "Starting from version 1.24.0 Kubernetes does not automatically generate a token for service accounts, in this case, `default_secret_name` will be empty",
+			},
+		}
+	}
 
 	for _, saSecret := range sa.Secrets {
 		if !strings.HasPrefix(saSecret.Name, fmt.Sprintf("%s-token-", sa.Name)) {
@@ -237,6 +261,9 @@ func diffObjectReferences(origOrs []api.ObjectReference, ors []api.ObjectReferen
 }
 
 func resourceKubernetesServiceAccountRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+
+	var diagMessages []diag.Diagnostic
+
 	exists, err := resourceKubernetesServiceAccountExists(ctx, d, meta)
 	if err != nil {
 		return diag.FromErr(err)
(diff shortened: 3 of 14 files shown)


kubernetes/provider.go
```diff
@@ -633,3 +634,26 @@ func useAdmissionregistrationV1beta1(conn *kubernetes.Clientset) (bool, error) {
 	useadmissionregistrationv1beta1 = ptrToBool(true)
 	return true, nil
 }
+
+func getServerVersion(connection *kubernetes.Clientset) (*gversion.Version, error) {
+	sv, err := connection.ServerVersion()
```
alexsomesan, line 639: We make use of this over here too: https://github.com/hashicorp/terraform-provider-kubernetes/blob/25512e4f8121192651ed5706e585a470cb73b699/kubernetes/resource_kubernetes_service.go#L445

Might be worth trying to unify the two.

kubernetes/resource_kubernetes_default_service_account.go
```diff
@@ -55,16 +55,22 @@ func resourceKubernetesDefaultServiceAccountCreate(ctx context.Context, d *schem
...
 	}
 
-	secret, err := getServiceAccountDefaultSecret(ctx, "default", svcAcc, d.Timeout(schema.TimeoutCreate), conn)
+	var defaultSecret *api.Secret
+	b, err := serverVersionLessThan(conn, "1.24.0")
 	if err != nil {
 		return diag.FromErr(err)
 	}
-	d.Set("default_secret_name", secret.Name)
+	if b {
```
alexsomesan, line 67: Silently skipping setting any value on `default_secret_name` might be confusing to users with existing configurations.
I would suggest returning a Warning level Diagnostic when we do this, to alert users in the Terraform output about the new behaviour.

alexsomesan (commented):
Nicely done! 
I left a comment about how we surface the changed behaviour to users so that they are not caught off-guard by the missing default token.

kubernetes/provider.go
arybolovlev: Thanks, made a small improvement. 👍🏻

kubernetes/resource_kubernetes_default_service_account.go
arybolovlev: I have updated the logic and moved the condition statement that verifies the version a bit lower. Now we should get a warning when creating or updating any of `r/kubernetes_default_service_account`, `d/kubernetes_service_account`, and `r/kubernetes_service_account`, however, when we do the import for `r/kubernetes_service_account` -- we don't get a warning message, because import expects only an error as return, and this is not an error.

flokli:
I applied this patch on top of terraform-provider-kubernetes, and am successfully able to create `kubernetes_service_account` resources again, so that works, thanks :-)

I however noted I always get the "Warning: 'default_secret_name' is no longer applicable for Kubernetes 'v1.24.0' and above" warning, not only in case the attribute is used. Is that expected?

yuriy-yarosh:
@alexsomesan hopefully this will be merged soon-ish™

kubernetes/resource_kubernetes_service_account.go
```diff
@@ -106,12 +107,21 @@ func resourceKubernetesServiceAccountCreate(ctx context.Context, d *schema.Resou
 	if err != nil {
 		return diag.FromErr(err)
 	}
+
 	return resourceKubernetesServiceAccountRead(ctx, d, meta)
 }
 
 func getServiceAccountDefaultSecret(ctx context.Context, name string, config api.ServiceAccount, timeout time.Duration, conn *kubernetes.Clientset) (*api.Secret, error) {
+	b, err := serverVersionGreaterThanOrEqual(conn, "1.24.0")
```
jrhouston, line 115: Super nitpicky but why did we call this `b` everywhere? I'd have called it `v` or `sv` just cause it's server version

kubernetes/resource_kubernetes_service_account.go
```diff
@@ -284,6 +311,17 @@ func resourceKubernetesServiceAccountRead(ctx context.Context, d *schema.Resourc
...
 	defaultSecretName := d.Get("default_secret_name").(string)
+	b, err := serverVersionGreaterThanOrEqual(conn, "1.24.0")
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	if b {
+		diagMessages = append(diagMessages, diag.Diagnostic{
+			Severity: diag.Warning,
+			Summary:  "'default_secret_name' is no longer applicable for Kubernetes 'v1.24.0' and above",
+			Detail:   "Starting from version 1.24.0 Kubernetes does not automatically generate a token for service accounts, in this case, `default_secret_name` will be empty",
```
jrhouston, line 322: I think we usually we `"double quote"` attributes we reference in diagnostic messages rather than single quote them.

kubernetes/resource_kubernetes_service_account.go
```diff
@@ -237,6 +261,9 @@ func diffObjectReferences(origOrs []api.ObjectReference, ors []api.ObjectReferen
 }
 
 func resourceKubernetesServiceAccountRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+
+	var diagMessages []diag.Diagnostic
```
jrhouston, line 265: you could put this in the function signature instead, like this:

```
func resourceKubernetesServiceAccountRead(...) (diag diag.Diagnostics)
```

jrhouston (approved):
Thanks for fixing this @arybolovlev – I've approved but left a couple of cosmetic comments.

kubernetes/resource_kubernetes_service_account.go
arybolovlev: Changed to `sv`, `b` was standing for `bool` since `serverVersionGreaterThanOrEqual` returns bool and error, but `sv` makes more sense in this context. Thanks!

kubernetes/resource_kubernetes_service_account.go
arybolovlev: Changed to double quote, thanks!

kubernetes/resource_kubernetes_service_account.go
arybolovlev: I have updated this function and moved the server version verification to the end of it. Should be more clean and readable now.

arybolovlev:
> I applied this patch on top of terraform-provider-kubernetes, and am successfully able to create `kubernetes_service_account` resources again, so that works, thanks :-)
> 
> I however noted I always get the "Warning: 'default_secret_name' is no longer applicable for Kubernetes 'v1.24.0' and above" warning when creating a `kubernetes_service_account` resource, not only in case the attribute is used. Is that expected?

I would say this is a trade-off now. In the case of the managed Kubernetes cluster solution, this new behavior might not be delivered to the users and they might not be aware of it. So it can be a source of confusion when the provider does not produce desired output anymore. The same behavior you may observe with other tools in the Kubernetes world. They keep repeating the same warning or deprecation message a long time before it actually happens to make sure that everyone is aware of it.

I hope that explanation makes sense. 😊

Thank you!

jrhouston (approved):

flokli:
> > I applied this patch on top of terraform-provider-kubernetes, and am successfully able to create `kubernetes_service_account` resources again, so that works, thanks :-)
> > I however noted I always get the "Warning: 'default_secret_name' is no longer applicable for Kubernetes 'v1.24.0' and above" warning when creating a `kubernetes_service_account` resource, not only in case the attribute is used. Is that expected?
> 
> I would say this is a trade-off now. In the case of the managed Kubernetes cluster solution, this new behavior might not be delivered to the users and they might not be aware of it. So it can be a source of confusion when the provider does not produce desired output anymore. The same behavior you may observe with other tools in the Kubernetes world. They keep repeating the same warning or deprecation message a long time before it actually happens to make sure that everyone is aware of it.
> 
> I hope that explanation makes sense. blush
> 
> Thank you!

I don't see how this explanation is related - the code is not using the `default_secret_name` attribute, so why is it showing a warning?

PR #1797, opened 2022-07-28
add test for loadbalanceingress IP/Hostname check

BBBmau:
### Description

GKE Test outputs the following error
```
    resource_kubernetes_service_test.go:130: Step 1/3 error: Check failed: 1 error occurred:
        	* Check 22/23 error: kubernetes_service.test: Attribute 'status.0.load_balancer.0.ingress.0.hostname' expected to be set
```
This is due to GKE not using hostname field but instead using the IP field, current test was changed in order to deal with either one.

### Acceptance tests

```
$ make testacc TESTARGS="-run TestAccKubernetesService_loadBalancer"

go vet .
TF_ACC=1 go test "/Users/mau/Dev/terraform-provider-kubernetes/kubernetes" -v -run TestAccKubernetesService_loadBalancer -timeout 3h
=== RUN   TestAccKubernetesService_loadBalancer
--- PASS: TestAccKubernetesService_loadBalancer (136.65s)

```

### Release Note

```release-note
NONE
```

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_service_test.go
+++ b/kubernetes/resource_kubernetes_service_test.go
@@ -818,30 +818,29 @@ func testAccCheckloadBalancerIngressCheck(resourceName string) resource.TestChec
 			return fmt.Errorf("Not found: %s", resourceName)
 		}
 
+		node, err := getFirstNode()
+
+		if err != nil {
+			return err
+		}
+
+		labels := node.GetLabels()
+
 		lb := "status.0.load_balancer.0.ingress.0"
 
-		InGke, _ := isRunningInGke()
-		if InGke {
+		if _, ok := labels["cloud.google.com/gke-nodepool"]; ok {
 			ip := fmt.Sprintf("%s.ip", lb)
-
-			if rs.Primary.Attributes[ip] == "" {
-				return fmt.Errorf("Attribute %s expected to be set in GKE cluster", ip)
+			if rs.Primary.Attributes[ip] != "" {
+				return nil
 			}
-
-			return nil
-		}
-
-		InEks, _ := isRunningInEks()
-		if InEks {
+			return fmt.Errorf("Attribute '%s' expected to be set for GKE cluster", ip)
+		} else {
 			hostname := fmt.Sprintf("%s.hostname", lb)
-
-			if rs.Primary.Attributes[hostname] == "" {
-				return fmt.Errorf("Attribute %s expected to be set in EKS cluster", hostname)
+			if rs.Primary.Attributes[hostname] != "" {
+				return nil
 			}
-			return nil
+			return fmt.Errorf("Attribute '%s' expected to be set for EKS cluster", hostname)
 		}
-
-		return fmt.Errorf("Cannot verify status of attribute %s", lb)
 	}
 }
 
--- a/kubernetes/resource_kubernetes_service_test.go
+++ b/kubernetes/resource_kubernetes_service_test.go
@@ -124,6 +124,7 @@ func TestAccKubernetesService_basic(t *testing.T) {
 
 func TestAccKubernetesService_loadBalancer(t *testing.T) {
 	var conf api.Service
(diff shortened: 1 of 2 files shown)


kubernetes/resource_kubernetes_service_test.go
```diff
@@ -811,6 +812,24 @@ func testAccCheckServicePorts(svc *api.Service, expected []api.ServicePort) reso
 	}
 }
 
+func testAccCheckloadBalancerIngressCheck(svc *api.LoadBalancerIngress) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		// Used with EKS Cluster
+		print(svc.Hostname)
+		print(svc.IP)
+		if svc.Hostname != "" {
```
alexsomesan, line 820: I'm a bit weary about relying just on the presence of this field value when trying to tell apart which cluster type we are running against.
We could try looking at the Node API objects instead and look for GKE specific labeling for example, or some other specific marker.

kubernetes/resource_kubernetes_service_test.go
```diff
@@ -811,6 +812,24 @@ func testAccCheckServicePorts(svc *api.Service, expected []api.ServicePort) reso
 	}
 }
 
+func testAccCheckloadBalancerIngressCheck(svc *api.LoadBalancerIngress) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		// Used with EKS Cluster
+		print(svc.Hostname)
```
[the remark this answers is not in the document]
BBBmau: Apologies, I meant to have the print values removed. Just used for testing purposes

kubernetes/resource_kubernetes_service_test.go
BBBmau: Followed your advice and went with using `isRunningInGke()` helper function that checks for the `"cloud.google.com/gke-nodepool"` label within the cluster to verify that we're in GKE. The same can be found when determining if we're in a EKS cluster. Helps in checking for either `hostname` or `ip` depending on type of cluster.

kubernetes/resource_kubernetes_service_test.go
```diff
@@ -811,6 +811,39 @@ func testAccCheckServicePorts(svc *api.Service, expected []api.ServicePort) reso
...
+
+		if err != nil {
+			return err
+		}
+
+		labels := node.GetLabels()
+
+		lb := "status.0.load_balancer.0.ingress.0"
+
+		if _, ok := labels["cloud.google.com/gke-nodepool"]; ok {
```
jrhouston, line 831: I'm not sure I would even be as clever as this. I would just check if the `hostname` is there, if it's not there then try `ip` and if that's not there then surface an error.

jrhouston (changes requested):
One thing. Also what's the context for the deletions that are also present in this PR?

kubernetes/resource_kubernetes_service_test.go
BBBmau: Yeah there were a couple of different ways to have checked for the `hostname` and the `ip`, one of them was just checking for `hostname` like you said but was also brought up that it would be great to check for the specific GKE label also by @alexsomesan. Had both planned and ready for PR but curious as to which approach to go with.

BBBmau:
> One thing. Also what's the context for the deletions that are also present in this PR?
Should've mentioned this, the deletions of `resource_kubernetes_deployment_test` comes from a fail that comes when running a GKE Cluster. This is what's outputted.

```
    resource_kubernetes_deployment_test.go:863: Step 1/1 error: Check failed: 1 error occurred:
        	* Check 1/2 error: deployment has already rolled out
```

I asked both @alexsomesan and @arybolovlev about it and came to the agreement that the test isn't really necessary and would be better of just removing it entirely since it's very flaky. I ran this specific test locally and ran with no problems.

jrhouston (approved):

PR #1807, opened 2022-08-19
chore: Default Timeout for KubernetesServiceAccount has been extended  to 1 minute.

kahirokunn:
### Description

All resources except KubernetesServiceAccount are more than 1 minute.
Only KubernetesServiceAccount is treated unequally, with a timeout of 30 seconds.
If the network is slowing down and you are applying to an environment that is physically far away, 30 seconds may cause a timeout.

Translated with www.DeepL.com/Translator (free version)
スクリーンショット 2022-08-19 15 07 33

Q: But why don't I just extend the timeout by myself?
A: Yes, but I don't understand why only KubernetesServiceAccount is given 30 seconds while all other resources except KubernetesServiceAccount are given more than 1 minute. KubernetesServiceAccount should be treated equally.

### Acceptance tests

```

...
```

### Release Note

```release-note
Default Timeout for KubernetesServiceAccount has been extended to 1 minute.
```

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_service_account.go
+++ b/kubernetes/resource_kubernetes_service_account.go
@@ -29,7 +29,7 @@ func resourceKubernetesServiceAccount() *schema.Resource {
 		},
 
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(30 * time.Second),
+			Create: schema.DefaultTimeout(1 * time.Minute),
 		},
 
 		Schema: map[string]*schema.Schema{


kahirokunn:
@jrhouston Could you please review it? 👀

arybolovlev:
Hi @kahirokunn,

Thank you for opening this PR. However, I will close it without merging it into our `main` branch.

The creation of a service account is a relatively fast operation and the timeout that we have in place was added to cover a case when we are waiting for a service account token to be created for this new service account. Kubernetes does not create service account tokens for new service accounts starting from the `v1.24`. For short time, the provider didn't handle this case and the creation of a new SA ended up with a timeout. This was fixed in `v2.13.0`.

30 seconds is enough for such a small task as this one and we will keep it as it is. If in your case it is not enough, please increase it per resource as described [here](https://developer.hashicorp.com/terraform/language/resources/syntax#operation-timeouts).

Thank you!

PR #1855, opened 2022-10-05
Update EKS GH Action workflow

arybolovlev:
### Description

Move TF_VAR_* variables at the top level of the workflow file so they are set up and available for all steps. In this case, infrastructure provisioning and destroying are executed with the variables.

Additional change: bump up the TF version.

### Acceptance tests


alexsomesan:
Nice catch!
I'm wondering if using global envs won't make it harder to extract those steps into re-usable actions like we were discussing recently?

BBBmau (approved):

alexsomesan (approved):

Issue #1974, opened 2023-01-27
Discussion on #1646: Why CRDs with preserve unknown fields forces recreation?


txomon:
## Terraform version, Kubernetes provider version and Kubernetes version
```
Terraform version:
Kubernetes Provider version: main( c9182af92723d4b1c801a8018c244100ebdc122c )
Kubernetes version: 1.23 and 1.24
```

## Question

I have seen that the code does the assumption that just because a field has changed from a previous state and has the `x-kubernetes-preserve-unknown-fields` annotation in the CRD, the provider now forces replacement.

This introduces several bugs regarding re-creation, which overlap with another issue on imports not populating the manifests.

What is the reasoning behind this change? What logic is being followed to force recreation?

alexsomesan:
Hi, please see my response on your PR for the explanation of why things are this way and cannot be changed.

Issue #1893, opened 2022-11-10
Kubernetes manifest forces recreation. Reopen #1679


txomon:
### Terraform Version, Provider Version and Kubernetes Version

```
Terraform version: v1.1.7
Kubernetes provider version: v2.15.0
Kubernetes version: v1.22.6
```

### Affected Resource(s)

- kubernetes_manifest

### Terraform Configuration Files
```hcl
locals {
  argocd_chart_version = "4.5.0"
  argocd_chart_repo    = "https://raw.githubusercontent.com/argoproj/argo-helm/argo-cd-"
  argocd_crd_path      = "/charts/argo-cd/crds/"
  argocd_raw_crd_path  = "${local.argocd_chart_repo}${local.argocd_chart_version}${local.argocd_crd_path}"
  argocd_crds = [
    "crd-application.yaml",
    "crd-extension.yaml",
    "crd-project.yaml",
  ]
}

data "http" "argocd_crd_files" {
  for_each = toset(local.argocd_crds)
  url      = "${local.argocd_raw_crd_path}${each.value}"
}

resource "kubernetes_manifest" "argocd_crd_files" {
  for_each = toset(local.argocd_crds)
  manifest = yamldecode(data.http.argocd_crd_files[each.key].body)
  lifecycle {
    prevent_destroy = true
  }
}
```

### Steps to Reproduce

1. install the argo CRDs (or any) using kubectl in your k8s cluster
2. import the CRD to your terraform project
```
terraform import 'kubernetes_manifest.argocd_crd_files["crd-project.yaml"]' "apiVersion=apiextensions.k8s.io/v1,kind=CustomResourceDefinition,name=appprojects.argoproj.io"
terraform apply
```

### Expected Behavior
What should have happened?
 It should apply with little to no changes
### Actual Behavior
What actually happened?

The plan is to "replace" the object which means the CRD and all instances of it will be deleted
```
# kubernetes_manifest.argocd_crd_files["crd-project.yaml"] must be replaced
```

### Important Factoids

Downgrading to `v2.8.0` of the provider results in the correct expected behavior.

### References

- #1679

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

BBBmau:
[...]

PR #1646, opened 2022-03-16
Handle 'x-kubernetes-preserve-unknown-fields' type annotation

alexsomesan:
Handle `x-kubernetes-preserve-unknown-fields` type annotation in OpenAPI.

Since we cannot guarantee structural integrity due to missing schema information, changes to attributes of this type will trigger recreation of the whole resource.

### Acceptance tests
- [ ] Have you added an acceptance test for the functionality being added?
- [ ] Have you run the acceptance tests on this branch?

Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccXXX'

...
```

### Release Note
Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-kubernetes/blob/main/CHANGELOG.md):

```release-note
...
```

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #1975, opened 2023-01-27
Stop forcing replacement of `kubernetes_manifest` resources when fields annotated with `x-kubernetes-preserve-unknown` are modified

txomon:
Having a delta between the desired resource status and the current resource status should not trigger a replacement of the resource.

### Description

The `kubernetes_manifest` is forcing replacement when fields are annotated with `x-kubernetes-preserve-unknown`, which I can't figure out the reasoning for, and is breaking my knative deployments

### Acceptance tests
   No but it has tests on it that go through that part of the code

```

...
```

### Release Note

```release-note
* Stop forcing replacement of `kubernetes_manifest` resources when fields annotated with `x-kubernetes-preserve-unknown` are modified
```

### References

#1893, #1974, #1646.

### Community Note

### DIFF
--- a/.changelog/1975.txt
+++ b/.changelog/1975.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fix issue when `kubernetes_manifest` resources would get replaced if there was a change in an attribute marked with `x-kubernetes-preserve-unknown` regardless of immutability
+```
--- a/manifest/provider/plan.go
+++ b/manifest/provider/plan.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
-	"github.com/hashicorp/terraform-provider-kubernetes/manifest"
 	"github.com/hashicorp/terraform-provider-kubernetes/manifest/morph"
 	"github.com/hashicorp/terraform-provider-kubernetes/manifest/payload"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -305,7 +304,7 @@ func (s *RawProviderServer) PlanResourceChange(ctx context.Context, req *tfproto
 	}
 
 	// Request a complete type for the resource from the OpenAPI spec
-	objectType, hints, err := s.TFTypeFromOpenAPI(ctx, gvk, false)
+	objectType, _, err := s.TFTypeFromOpenAPI(ctx, gvk, false)
 	if err != nil {
 		return resp, fmt.Errorf("failed to determine resource type ID: %s", err)
 	}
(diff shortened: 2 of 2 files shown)


artificial-aidan:
FWIW this fixed my problem as well.

alexsomesan:
Hi, unfortunately we cannot accept this change. It was a deliberate decision during early development of this resource to induce a replacement whenever a CR without schema information has any changes.

`x-kubernetes-preserve-unknown` is effectively a way for the CRD author to skip adding schema information. Without this schema information, the provider cannot guarantee an invariant that is strongly required by Terraform itself, namely a resource cannot change the types of any attributes during its lifecycle.
In the absence of schema information, a user can simply change the type of an attribute and nothing would prevent them from applying those changes. The problem would show up after the apply when Terraform would try to update the state file with the new values and would refuse to accept the changes in case any attributes changes type.

Because of that requirement in Terraform, we have to impose the replace strategy, to make sure apply results do not end up breaking the existing state.

alexsomesan:
@artificial-aidan Unfortunately, this is enforced by Terraform itself and there is no API / flag mechanism to disable it from the provider's POV. Providers simply need to comply with this requirement and you will get an error from Terraform when this happens and your state will be left in an inconsistent state.

We want to avoid at all cost breaking our users state files, so going this route seems like the lesser evil.

CRDs without schema are technically violating Kubernetes' own expectations and requirements so we are encouraging everyone who encounters them to raise the issue with their respective authors asking them to make their CRDs compliant.

txomon:
Hello Alex,

Thanks for your answer. I will try to reply inline.

> Hi, unfortunately we cannot accept this change. It was a deliberate decision during early development of this resource to induce a replacement whenever a CR without schema information has any changes.

This is causing a regression in upgrades from `v2.7.1`. We have been using the `kubernetes_manifest` resource since 2020 and this was not a problem before.

> x-kubernetes-preserve-unknown is effectively a way for the CRD author to skip adding schema information. Without this schema information, the provider cannot guarantee an invariant that is strongly required by Terraform itself, namely a resource cannot change the types of any attributes during its lifecycle.

I understand this is something extremely important, but right now the code is forcing a replacement in many other cases than when the type of the attribute changes. 

`x-kubernetes-preserve-unknown` is a field used to enable CRDs to accept new attributes that might come in the future without having to force cluster operators to update the CRD for every specific field that changes. To give an example, our biggest usecase is using this for knative, which has `template` instance similar to the `deployment` resource. Plugins can actually add new functionality in a pod spec, and updating all CRDs for that would be nonsense.

Custom resources usually have a validating webhook that will enforce schema and type requirements.

The current code is putting in the same bag any change to be akin to an attribute type change, which is extremely rare. Or to say it other way, it's elevating the requirements to use the `kubernetes_manifest` resource to a level that makes the resource itself useless.

> In the absence of schema information, a user could simply change the type of an attribute and nothing would prevent them from applying those changes since we have no schema to validate against. The problem would show up after the apply when Terraform would try to update the state file with the new values and would refuse to accept the changes in case any attributes changes type.

The code isn't currently checking that there is an attribute type change, but that there is a change.

To bring the matter of usability to the forefront, there is no way around this hardcoded limitation. Failing to save the state could be fixed manually by the operator by modifying by hand the state file. However, this limitation is imposible to overcome without replacing/fixing the provider.

Another usecase that came to my mind would be actual immutability requirements, such as `selector.matchLabels`, however that's an entirely new feature.

Finally, I might be wrong, but I think that the tests are not checking whether the resource is being updated or replaced, so I'm not sure how the worst case scenario and its workaround is being kept in check.

For anyone that comes to this PR and needs to rely on this fix, I have published the fork with the change at [txomon/kubernetes](https://registry.terraform.io/providers/txomon/kubernetes/latest)

artificial-aidan:
> For anyone that comes to this PR and needs to rely on this fix, I have published the fork with the change at [txomon/kubernetes](https://registry.terraform.io/providers/txomon/kubernetes/latest)

Thank you! I was halfway through this process, but thanks for saving me the time.

alexsomesan:
@txomon you mentioned:
> The current code is putting in the same bag any change to be akin to an attribute type change, which is extremely rare. Or to say it other way, it's elevating the requirements to use the kubernetes_manifest resource to a level that makes the resource itself useless.

This I think holds the key to making some progress on this topic. While we cannot accept a change that blanket turns off the replace behaviour, I would be interested to look at a PR that detects type changes specifically and only forces replacement in that case.

txomon:
@alexsomesan 

Do you mean https://github.com/hashicorp/terraform-plugin-go/blob/main/tftypes/walk.go#L194 ?

txomon:
Hello @alexsomesan, I didn't receive an answer to this. The `Transform` code is already checking for type changes, so I don't think we need that extra check. Please let me know if I read wrong

Issue #1950, opened 2023-01-09
panic: required attribute "preStop" not set


vmotso:
### Terraform Version, Provider Version and Kubernetes Version

```
Terraform version:  v1.3.5
Kubernetes provider version: v2.16.1
Kubernetes version: v1.22.15-gke.100
```

### Terraform Configuration Files
```hcl
locals {
  condition = false
  postStart = {
    exec = {
      command = ["bash", "-c", "echo postStart"]
    }
  }
  preStop = {
    exec = {
      command = ["bash", "-c", "echo preStop"]
    }
  }
}
resource "kubernetes_manifest" "deploy_foo" {
  manifest = {
    apiVersion = "apps/v1"
    kind       = "Deployment"
    metadata = {
      name      = "foo"
      namespace = "foo"
      labels = {
        "app" = "foo"
      }
    }
    spec = {
      selector = {
        matchLabels = {
          "app" = "foo"
        }
      }
      template = {
        metadata = {
          labels = {
            "app" = "foo"
          },
        }
        spec = {
          containers = [
            {
              name  = "foo"
              image = "busybox"
              lifecycle = local.condition ? {
                postStart = local.postStart
                preStop   = local.preStop
                } : {
                postStart = local.postStart
              }
            }
          ]
        }
      }
    }
  }
}
```

### Panic Output

```
Stack trace from the terraform-provider-kubernetes_v2.16.1_x5 plugin:

panic: can't create a tftypes.Value of type tftypes.Object["postStart":tftypes.Object["exec":tftypes.Object["command":tftypes.List[tftypes.String]], "httpGet":tftypes.Object["host":tftypes.String, "httpHeaders":tftypes.List[tftypes.Object["name":tftypes.String, "value":tftypes.String]], "path":tftypes.String, "port":tftypes.String, "scheme":tftypes.String], "tcpSocket":tftypes.Object["host":tftypes.String, "port":tftypes.String]], "preStop":tftypes.Object["exec":tftypes.Object["command":tftypes.List[tftypes.String]], "httpGet":tftypes.Object["host":tftypes.String, "httpHeaders":tftypes.List[tftypes.Object["name":tftypes.String, "value":tftypes.String]], "path":tftypes.String, "port":tftypes.String, "scheme":tftypes.String], "tcpSocket":tftypes.Object["host":tftypes.String, "port":tftypes.String]]], required attribute "preStop" not set

goroutine 43 [running]:
[...]

PR #2018, opened 2023-02-24
Add Diagnostic when determining Terraform type fails

BBBmau:
### Description

Outputs a better error message than what was previously shown if their was an issue with determining a TF type from manifest

Related to #1950 

### Acceptance tests

### Community Note


manifest/morph/morph.go
```diff
@@ -11,8 +11,16 @@ import (
 )
 
 // ValueToType transforms a value along a new type and returns a new value conforming to the given type
-func ValueToType(v tftypes.Value, t tftypes.Type, p *tftypes.AttributePath) (tftypes.Value, []*tfprotov5.Diagnostic) {
-	var diags []*tfprotov5.Diagnostic
+func ValueToType(v tftypes.Value, t tftypes.Type, p *tftypes.AttributePath) (val tftypes.Value, diags []*tfprotov5.Diagnostic) {
+	defer func() {
+		if err := recover(); err != nil {
```
jrhouston, line 16: I was having a look at the godoc for tftypes and found this comment:

```
If the passed value is not a valid value for the passed type, NewValue will panic. Any value and type 
combination that does not return an error from ValidateValue is guaranteed to not panic. When calling 
NewValue with user input with a type not known at compile time, it is recommended to call ValidateValue 
before calling NewValue, to allow graceful handling of the error.
```

So it looks like although `NewValue()` will produce panics, it is guaranteed not to panic if we use `ValidateValue` with the same arguments first, which does return an error which we can then use to return a diagnostic message. This is slightly better than just capturing any panic produced by this function. 

It means we will need to look at each call site for `tftypes.NewValue()` and add a call to `ValidateValue()` and return a diag if it fails. Perhaps we can make a `newValue()` function to wrap this so we don't need to repeat ourselves a bunch of times.

jrhouston (changes requested):

manifest/morph/morph.go
```diff
@@ -23,6 +22,15 @@ func ValueToType(v tftypes.Value, t tftypes.Type, p *tftypes.AttributePath) (tft
 		return tftypes.Value{}, diags
 	}
 	if v.IsNull() {
+		if err := tftypes.ValidateValue(t, nil); err != nil {
```
jrhouston, line 25: This whole block of code is repeated verbatim in quite a few places. Can you think of a way where we could capture this logic once but re-use it in each of these places?

jrhouston (changes requested):

manifest/morph/morph.go
```diff
@@ -710,3 +766,17 @@ func attributePathSummary(p *tftypes.AttributePath) string {
 	}
 	return b.String()
 }
+
+func ValidateValue(t tftypes.Type, val interface{}, p *tftypes.AttributePath) []*tfprotov5.Diagnostic {
```
jrhouston, line 770: We don't need to export this function so it can be called `validateValue()`

manifest/morph/morph.go
```diff
@@ -11,8 +11,7 @@ import (
 )
 
 // ValueToType transforms a value along a new type and returns a new value conforming to the given type
-func ValueToType(v tftypes.Value, t tftypes.Type, p *tftypes.AttributePath) (tftypes.Value, []*tfprotov5.Diagnostic) {
-	var diags []*tfprotov5.Diagnostic
+func ValueToType(v tftypes.Value, t tftypes.Type, p *tftypes.AttributePath) (val tftypes.Value, diags []*tfprotov5.Diagnostic) {
```
jrhouston, line 14: I don't think we need this change to the function signature anymore.

manifest/morph/morph.go
jrhouston: We could maybe go a step further here and have a function called `newValue()` that wraps the calls to `ValidateValue()` and `NewValue()` and use that at every call site that exists now for `NewValue()`.

jrhouston (changes requested):

manifest/morph/morph.go
```diff
@@ -626,7 +626,7 @@ func morphObjectToType(v tftypes.Value, t tftypes.Type, p *tftypes.AttributePath
 		// tftypes.NewValue() fails if any of the attributes in the object don't have a corresponding value
 		for k := range t.(tftypes.Object).AttributeTypes {
 			if _, ok := ovals[k]; !ok {
-				ovals[k] = tftypes.NewValue(t.(tftypes.Object).AttributeTypes[k], nil)
+				ovals[k], _ = newValue(t.(tftypes.Object).AttributeTypes[k], nil, p)
```
jrhouston, line 629: I wouldn't suppress the diag here with `_`. I would check if there's a diag and return it.

manifest/morph/morph.go
```diff
@@ -655,7 +660,10 @@ func morphObjectToType(v tftypes.Value, t tftypes.Type, p *tftypes.AttributePath
 			}
 			mvals[k] = nv
 		}
-		return tftypes.NewValue(t, mvals), diags
+		if diags := validateValue(t, mvals, p); diags != nil {
```
jrhouston, line 663: I'm not sure what the purpose of this function is if newValue is going to call validateValue anyway?

jrhouston (changes requested):

manifest/morph/morph.go
BBBmau: this might've been left when it came to figuring out why one of the manifest tests wasn't passing. Changed it to just using newValue() with checks all passing

jrhouston (approved):
Thanks @BBBmau 🚀

PR #2046, opened 2023-03-15
Remove limit for taints in `kubernetes_node_taint` resource

### LINKED ISSUE #2035: kubernetes_node_taint: Too many taint blocks
```
Terraform version:1.3.5
Kubernetes Provider version:2.18.1
Kubernetes version:1.19.3
```
## Terraform configuration
```hcl
resource "kubernetes_node_taint" "nodetaint" {
  metadata {
    name = "node1"
  }
  taint {
    key    = "taint1"
    value  = "true"
    effect = "NoSchedule"
  }
  taint {
    key    = "taint2"
    value  = "true"
    effect = "NoSchedule"
  }
  force = true
  field_manager = "terraform-taint"
}

```
## Error
```
Error: Too many taint blocks
│ 
│   on test.tf line 10, in resource "kubernetes_node_taint" "nodetaint":
│   10:   taint {
│ 
│ No more than 1 "taint" blocks are allowed
```

BBBmau:
### Description

Fixes #2035

Removes the `MaxItems` line from the "taint" attribute. Should allow the ability of multiple taint blocks if desired.

### Release Note

```release-note
`kubernetes/resource_kubernetes_node_taint.go`: Remove MaxItems from taint attribute
```

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_node_taint.go
+++ b/kubernetes/resource_kubernetes_node_taint.go
@@ -29,6 +29,24 @@ func resourceKubernetesNodeTaint() *schema.Resource {
 		ReadContext:   resourceKubernetesNodeTaintRead,
 		UpdateContext: resourceKubernetesNodeTaintUpdate,
 		DeleteContext: resourceKubernetesNodeTaintDelete,
+		CustomizeDiff: func(ctx context.Context, rd *schema.ResourceDiff, i interface{}) error {
+			if !rd.HasChange("taint") {
+				return nil
+			}
+			// check for duplicate taint keys
+			taintkeys := map[string]int{}
+			for _, t := range rd.Get("taint").([]interface{}) {
+				taint := t.(map[string]interface{})
+				key := taint["key"].(string)
+				taintkeys[key] = taintkeys[key] + 1
+			}
+			for k, v := range taintkeys {
+				if v > 1 {
+					return fmt.Errorf("taint: duplicate taint key %q: taint keys must be unique", k)
+				}
+			}
+			return nil
+		},
 		Schema: map[string]*schema.Schema{
 			"metadata": {
 				Type:     schema.TypeList,
--- a/kubernetes/resource_kubernetes_node_taint.go
+++ b/kubernetes/resource_kubernetes_node_taint.go
@@ -164,9 +164,9 @@ func resourceKubernetesNodeTaintUpdate(ctx context.Context, d *schema.ResourceDa
 			var updated bool
 			newNode, updated = addOrUpdateTaint(node, &newTaint)
 			if !updated {
-				return diag.Errorf("Node %s already has taint %+v", nodeName, newTaints)
+				log.Printf("Node %s already has taint %+v", nodeName, newTaints)
 			}
-			taintList = append(taintList, newNode.Spec.Taints...)
+			taintList = append(taintList, newNode.Spec.Taints[0])
 		}
 	}
 	//panic(fmt.Sprintf("%v", newNode.Spec.Taints))
--- a/kubernetes/resource_kubernetes_node_taint.go
+++ b/kubernetes/resource_kubernetes_node_taint.go
@@ -88,10 +88,8 @@ func resourceKubernetesNodeTaintDelete(ctx context.Context, d *schema.ResourceDa
 }
 
 func resourceKubernetesNodeTaintRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-	nodeName, idTaint, err := idToNodeTaint(d.Id())
-	if err != nil {
-		return diag.FromErr(err)
-	}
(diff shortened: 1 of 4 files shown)


arybolovlev (dismissed):
This PR does not address the issue. Yes, Terraform is not throwing an error message anymore. However, the `kubernetes_node_taint` resource takes into account only the first taint in the list, the rest doesn't apply. As the result, we get diff with every plan run.

Another point here is that tests were not updated to validate changes. I would suggest here the TDD approach. The idea is you first write a test that reproduces an issue without changing the code. In short, the tests should fail at this point. Once you have that done, you make changes in the code and run tests again, if the issue is addressed, the tests will pass.

kubernetes/resource_kubernetes_node_taint.go
```diff
@@ -75,7 +70,7 @@ func resourceKubernetesNodeTaint() *schema.Resource {
 
 func resourceKubernetesNodeTaintCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 	metadata := expandMetadata(d.Get("metadata").([]interface{}))
-	d.SetId(nodeTaintToId(metadata.Name, d.Get("taint").([]interface{})))
+	d.SetId(nodeTaintToId(fmt.Sprintf("%s", metadata.Name), d.Get("taint").([]interface{})))
```
jrhouston, line 73: You can remove the call to Sprintf here and just use `metadata.Name` directly.

jrhouston (changes requested):

kubernetes/resource_kubernetes_node_taint_test.go
```diff
@@ -107,3 +136,31 @@ resource "kubernetes_node_taint" "test" {
...
+  }
+  taint {
+    key    = "%s-1"
+    value  = "%s"
+    effect = "%s"
+  }
+  taint {
+    key    = "%s-2"
+    value  = "%s"
+    effect = "%s"
```
jrhouston, line 156: Anywhere you have `"%s"` you can use the`%q` directive instead.

BBBmau:
Tried the case where a user will add taints with the same key but different values. A panic is outputted but the tfstate still shows the applied config.

tfconfig:
```hcl
provider "kubernetes" {
  config_path    = "~/.kube/config"
}

resource "kubernetes_node_taint" "tainttest" {
  metadata {
    name = "minikube"
  }

  taint {
    key    = "node-role.kubernetes.io/example"
    value  = "false"
    effect = "NoSchedule"
  }
    taint {
    key    = "node-role.kubernetes.io/example"
    value  = "true"
    effect = "NoSchedule"
  }
}
```
tfstate:
```
            "taint": [
              {
                "effect": "NoSchedule",
                "key": "node-role.kubernetes.io/example",
                "value": "false"
              },
              {
                "effect": "NoSchedule",
                "key": "node-role.kubernetes.io/example",
                "value": "true"
              }
            ]
```

doing `terraform refresh` turns it back to the state before, but this extra step shouldn't be needed.

kubernetes/resource_kubernetes_node_taint_test.go
```diff
@@ -44,6 +44,35 @@ func TestAccKubernetesResourceNodeTaint_basic(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		IDRefreshName:     resourceName,
+		ProviderFactories: testAccProviderFactories,
+		CheckDestroy:      testAccKubernetesNodeTaintDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccKubernetesNodeTaintConfig_multipleBasic(),
+				Check: resource.ComposeAggregateTestCheckFunc(
```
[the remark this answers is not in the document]
BBBmau: I'll probably be removing `testAccKubernetesNodeTaintConfig_removeAllTaints()` test step since we need to have minimum 1 taint unless we remove the `minItems` but it feels a bit pointless to have no taints in the node taint resource.

jrhouston:
@BBBmau This is expected behaviour, you can't set multiple taints with the same key. We should add some validation here so that it fails on the plan instead of generating the error at apply time.

jrhouston:
The list type doesn't support validators so I've pushed some logic that does the validation in a custom diff.

jrhouston:
Now it will show on plan time:

```
terraform plan
╷
│ Error: taint: duplicate taint key "node-role.kubernetes.io/example": taint keys must be unique
│
│   with kubernetes_node_taint.tainttest,
│   on main.tf line 5, in resource "kubernetes_node_taint" "tainttest":
│    5: resource "kubernetes_node_taint" "tainttest" {
│
╵
```

jrhouston (approved):

jrhouston (approved):

PR #2067, opened 2023-04-03
Add support for `initContainers` in `kubernetes_env` resource

BBBmau:
### Description

Adds support of initContainers to `kubernetes_env`.

This feature request issue was mentioned here #723

### Acceptance tests

```
➜  terraform-provider-kubernetes git:(add-initContainer-env) ✗ make testacc TESTARGS="-run TestAccKubernetesEnv_CronJob_initContainer"
go vet .
TF_ACC=1 go test "/Users/mau/Dev/terraform-provider-kubernetes/kubernetes" -v -run TestAccKubernetesEnv_CronJob_initContainer -timeout 3h
=== RUN   TestAccKubernetesEnv_CronJob_initContainer
--- PASS: TestAccKubernetesEnv_CronJob_initContainer (9.37s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   9.992s
```
```
➜  terraform-provider-kubernetes git:(add-initContainer-env) make testacc TESTARGS="-run TestAccKubernetesEnv_Deployment_initContainer"
go vet .
TF_ACC=1 go test "/Users/mau/Dev/terraform-provider-kubernetes/kubernetes" -v -run TestAccKubernetesEnv_Deployment_initContainer -timeout 3h
=== RUN   TestAccKubernetesEnv_Deployment_initContainer
--- PASS: TestAccKubernetesEnv_Deployment_initContainer (9.50s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   10.265s
...
```

### Release Note

```release-note
`kubernetes/resource_kubernetes_env.go`: add support for initContainers
```

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_env.go
+++ b/kubernetes/resource_kubernetes_env.go
@@ -59,12 +59,14 @@ func resourceKubernetesEnv() *schema.Resource {
 				Type:         schema.TypeString,
 				Description:  "Name of the container for which we are updating the environment variables.",
 				Optional:     true,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
 				ExactlyOneOf: []string{"container", "init_container"},
 			},
 			"init_container": {
 				Type:         schema.TypeString,
 				Description:  "Name of the initContainer for which we are updating the environment variables.",
 				Optional:     true,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
 				ExactlyOneOf: []string{"container", "init_container"},
 			},
 			"api_version": {
--- a/kubernetes/resource_kubernetes_env.go
+++ b/kubernetes/resource_kubernetes_env.go
@@ -344,6 +344,7 @@ func getResponseEnvs(u *unstructured.Unstructured, containerName string, kind st
 func getManagedEnvs(managedFields []v1.ManagedFieldsEntry, manager string, d *schema.ResourceData, u *unstructured.Unstructured) (map[string]interface{}, error) {
 	var envs map[string]interface{}
 	kind := d.Get("kind").(string)
+	isInitContainer := d.Get("init_container").(bool)
 	for _, m := range managedFields {
 		if m.Manager != manager {
 			continue
@@ -362,11 +363,22 @@ func getManagedEnvs(managedFields []v1.ManagedFieldsEntry, manager string, d *sc
 			return nil, err
 		}
 
-		containers := spec["f:containers"].(map[string]interface{})
-		containerName := fmt.Sprintf(`k:{"name":%q}`, d.Get("container").(string))
-		k := containers[containerName].(map[string]interface{})
-		if e, ok := k["f:env"].(map[string]interface{}); ok {
-			envs = e
+		if !isInitContainer {
+			containers := spec["f:containers"].(map[string]interface{})
+			containerName := fmt.Sprintf(`k:{"name":%q}`, d.Get("container").(string))
+
+			k := containers[containerName].(map[string]interface{})
+			if e, ok := k["f:env"].(map[string]interface{}); ok {
+				envs = e
+			}
+		} else {
+			containers := spec["f:initContainers"].(map[string]interface{})
+			containerName := fmt.Sprintf(`k:{"name":%q}`, d.Get("container").(string))
+
+			k := containers[containerName].(map[string]interface{})
+			if e, ok := k["f:env"].(map[string]interface{}); ok {
+				envs = e
+			}
 		}
 	}
 	return envs, nil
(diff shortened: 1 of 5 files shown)


BBBmau:
Was thinking since we’ll have `kubernetes_env` resource for every `container` or `initContainer` that needs envs. it seems like a good idea to just have a boolean attribute of `initContainer` to determine whether we’re adding env to a regular container or initContainer.

kubernetes/resource_kubernetes_env.go
```diff
@@ -60,6 +60,12 @@ func resourceKubernetesEnv() *schema.Resource {
 				Description: "Name of the container for which we are updating the environment variables.",
 				Required:    true,
 			},
+			"init_container": {
```
jrhouston, line 63: The options we discussed on our call today:

1. Don't add a new attribute, and just use `container` and find the container with that name in initContainers. Container names have to be unique across initContainers and containers so this should work.

2. To be explicit, add a `init_container` attribute which is a string and functions the same as `container` but only applies to the initContainers list. We can use `ExactlyOneOf` to enforce that one of `container` or `init_container` must be set. This would make the configuration a bit more explicit in it's intentions.

kubernetes/resource_kubernetes_env.go
```diff
@@ -348,11 +374,23 @@ func getManagedEnvs(managedFields []v1.ManagedFieldsEntry, manager string, d *sc
...
+			}
+		} else {
+			containers := spec["f:containers"].(map[string]interface{})
+			containerName := fmt.Sprintf(`k:{"name":%q}`, d.Get("container").(string))
+
+			k := containers[containerName].(map[string]interface{})
+			if e, ok := k["f:env"].(map[string]interface{}); ok {
+				envs = e
+			}
+
```
jrhouston, line 393: These two blocks of code are very similar – can you think of a way to make it simpler?

kubernetes/resource_kubernetes_env.go
```diff
@@ -308,13 +325,22 @@ func resourceKubernetesEnvRead(ctx context.Context, d *schema.ResourceData, m in
...
+		if kind == "CronJob" {
+			containers, _, _ = unstructured.NestedSlice(u.Object, "spec", "jobTemplate", "spec", "template", "spec", "initContainers")
+		}
+	} else {
+		containers, _, _ = unstructured.NestedSlice(u.Object, "spec", "template", "spec", "containers")
+
+		if kind == "CronJob" {
+			containers, _, _ = unstructured.NestedSlice(u.Object, "spec", "jobTemplate", "spec", "template", "spec", "containers")
+		}
```
jrhouston, line 342: Since container names are unique, maybe instead of passing down `isInitContainer` you could just append both initContainers and containers together and search for the name?

kubernetes/resource_kubernetes_env.go
```diff
@@ -425,37 +463,75 @@ func resourceKubernetesEnvUpdate(ctx context.Context, d *schema.ResourceData, m
...
+			spec = map[string]interface{}{
+				"jobTemplate": map[string]interface{}{
+					"spec": map[string]interface{}{
+						"template": map[string]interface{}{
+							"spec": map[string]interface{}{
+								"initContainers": []interface{}{
+									map[string]interface{}{
+										"name": d.Get("init_container").(string),
+										"env":  env,
+									},
```
jrhouston, line 525: Similar to my other comment, there's a fair bit of similar code being repeated here. Maybe there's a way you could use a variable to switch between `initContainers` and `containers` since they have the same structure.

kubernetes/resource_kubernetes_env_test.go
```diff
@@ -242,6 +420,77 @@ func createCronJobEnv(t *testing.T, name, namespace string) error {
...
+	if err != nil {
+		return err
+	}
+	ctx := context.Background()
+
+	var failJobLimit int32 = 5
+	var startingDeadlineSeconds int64 = 2
+	var successfulJobsLimit int32 = 2
+	var boLimit int32 = 2
+	var ttl int32 = 2
```
jrhouston, line 434: Just fyi there's a util package [k8s.io/utils/pointer](https://pkg.go.dev/k8s.io/utils/pointer) that you can use to more conveniently set these fields that are pointers.

jrhouston (changes requested):
A couple suggestions

kubernetes/resource_kubernetes_env.go
BBBmau: made the map of name and env a variable, not sure if we can make the whole jobTemplate/spec/template/spec structure a variable but I may have just missed it.

jrhouston (approved):

PR #2080, opened 2023-04-17
Adding a Runtime class resource

sheneska:
### Description

Added a runtime class resource (with handler attribute only for now) & wrote test file that asserts for only the `creating a runtime class` step

### Acceptance tests

```
$ make testacc TESTARGS="-count 1 -run ^TestAccKubernetesruntime_class_v1_basic"

TF_ACC=1 go test "/Users/sacha/Documents/code/hashicorp/terraform-provider-kubernetes/kubernetes" -v -count 1 -run ^TestAccKubernetesruntime_class_v1_basic -timeout 3h
=== RUN   TestAccKubernetesruntime_class_v1_basic
--- PASS: TestAccKubernetesruntime_class_v1_basic (5.24s)
PASS
ok  	github.com/hashicorp/terraform-provider-kubernetes/kubernetes	6.160s
```

### Release Note

```release-note:new-resource
`resource/kubernetes_runtime_class_v1`: Add a new resource `kubernetes_runtime_class_v1`.
```

### References

Fix: https://github.com/hashicorp/terraform-provider-kubernetes/issues/1114

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_runtime_class_v1.go
+++ b/kubernetes/resource_kubernetes_runtime_class_v1.go
@@ -5,21 +5,15 @@ package kubernetes
 
 import (
 	"context"
-	//"fmt"
 	"log"
 	"regexp"
 
-	// "time"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 
-	//"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	nodev1 "k8s.io/api/node/v1"
 
-	// api "k8s.io/api/core/v1"
-	// "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	pkgApi "k8s.io/apimachinery/pkg/types"
@@ -38,14 +32,7 @@ func resourceKubernetesRuntimeClassV1() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"metadata": metadataSchema("runtimeclass", true),
-			// "overhead": {
-			// 	Type:        schema.TypeMap,
-			// 	Description: "Represents the esource overhead associated with running a pod for a given RuntimeClass",
-			// },
-			// "scheduling": {
-			// 	Type:        schema.TypeMap,
-			// 	Description: "Holds the scheduling constraints to ensure that pods running with this RuntimeClass are scheduled to nodes that support it",
-			// },
+
 			"handler": {
 				Type:         schema.TypeString,
 				Description:  "Specifies the underlying runtime and configuration that the CRI implementation will use to handle pods of this class",
@@ -64,12 +51,11 @@ func resourceKubernetesRuntimeClassV1Create(ctx context.Context, d *schema.Resou
 		return diag.FromErr(err)
 	}
 
-	//converting metadata for resource -> HCL for TF
 	metadata := expandMetadata(d.Get("metadata").([]interface{}))
 
 	runtimeClass := nodev1.RuntimeClass{
 		ObjectMeta: metadata,
-		Handler:    d.Get("handler").(string), //returns the string from handler
+		Handler:    d.Get("handler").(string),
 	}
 
 	out, err := conn.NodeV1().RuntimeClasses().Create(ctx, &runtimeClass, metav1.CreateOptions{})
@@ -78,9 +64,9 @@ func resourceKubernetesRuntimeClassV1Create(ctx context.Context, d *schema.Resou
 	}
 
 	log.Printf("[INFO] New runtime class created: %#v", out)
-	d.SetId(out.Name) //id of resource used in the state file, not a namespace refers to its name of the resource
+	d.SetId(out.Name)
 
-	return resourceKubernetesRuntimeClassV1Read(ctx, d, meta) //create & read goes hand in hand basically
+	return resourceKubernetesRuntimeClassV1Read(ctx, d, meta)
 }
 
 func resourceKubernetesRuntimeClassV1Read(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
@@ -100,14 +86,14 @@ func resourceKubernetesRuntimeClassV1Read(ctx context.Context, d *schema.Resourc
 
 	name := d.Id()
 
-	log.Printf("[INFO] Reading Run Time Class %s", name)
+	log.Printf("[INFO] Reading runtime class %s", name)
 	rc, err := conn.NodeV1().RuntimeClasses().Get(ctx, name, metav1.GetOptions{})
 	if err != nil {
 		log.Printf("[DEBUG] Received error: %#v", err)
 		return diag.FromErr(err)
 	}
 
-	log.Printf("[INFO] Received Run Time Class: %#v", rc)
+	log.Printf("[INFO] Received runtime class: %#v", rc)
 	err = d.Set("metadata", flattenMetadata(rc.ObjectMeta, d, meta))
 	if err != nil {
 		return diag.FromErr(err)
@@ -136,14 +122,14 @@ func resourceKubernetesRuntimeClassV1Update(ctx context.Context, d *schema.Resou
 		return diag.Errorf("Failed to marshal update operations: %s", err)
 	}
 
-	log.Printf("[INFO] Updating run time class %s: %#v", d.Id(), patch)
+	log.Printf("[INFO] Updating runtime class %s: %#v", d.Id(), patch)
 
 	out, err := conn.NodeV1().RuntimeClasses().Patch(ctx, name, pkgApi.JSONPatchType, data, metav1.PatchOptions{})
 	if err != nil {
-		return diag.Errorf("Failed to update run time class! API error: %s", err)
+		return diag.Errorf("Failed to update runtime class! API error: %s", err)
 	}
 
-	log.Printf("[INFO] Submitted updated run time class: %#v", out)
+	log.Printf("[INFO] Submitted updated runtime class: %#v", out)
 	d.SetId(out.Name)
 
 	return resourceKubernetesRuntimeClassV1Read(ctx, d, meta)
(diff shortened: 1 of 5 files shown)


kubernetes/resource_kubernetes_runtime_class_v1_test.go
```diff
@@ -0,0 +1,142 @@
...
+      TestAnnotationTwo = "two"
+    }
+    name = %q
+  }
+}
+`, name)
+}
+
+func testAccKubernetesruntime_class_v1_addLabels(name string) string {
+	return fmt.Sprintf(`resource "kubernetes_namespace" "test" {
```
arybolovlev, line 102: Please change the resource name to `kubernetes_runtime_class_v1`.

kubernetes/resource_kubernetes_runtime_class_v1_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+func testAccKubernetesruntime_class_v1_addAnnotations(name string) string {
+	return fmt.Sprintf(`resource "kubernetes_runtime_class_v1" "test" {
+  metadata {
+    annotations = {
+      TestAnnotationOne = "one"
+      TestAnnotationTwo = "two"
+    }
+    name = %q
+  }
```
arybolovlev, line 96: handler is the required attribute, please add the one here.

kubernetes/resource_kubernetes_runtime_class_v1_test.go
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    labels = {
+      TestLabelOne   = "one"
+      TestLabelTwo   = "two"
+      TestLabelThree = "three"
+    }
+
+    name = %q
+  }
```
arybolovlev, line 116: handler is the required attribute, please add the one here.

kubernetes/resource_kubernetes_runtime_class_v1_test.go
```diff
@@ -0,0 +1,142 @@
...
+	var conf nodev1.RuntimeClass
+	rcName := fmt.Sprintf("tf-acc-test-%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
+	resourceName := "kubernetes_runtime_class_v1.test"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		IDRefreshName:     resourceName,
+		IDRefreshIgnore:   []string{"metadata.0.resource_version"},
+		ProviderFactories: testAccProviderFactories,
+		//CheckDestroy:      testAccCheckKubernetesRuntimeClassDestroy,
```
arybolovlev, line 28: Please uncomment `CheckDestroy` and implement the function mentioned here. It would be something similar to [this one](https://github.com/hashicorp/terraform-provider-kubernetes/blob/main/kubernetes/resource_kubernetes_namespace_test.go#L211).

kubernetes/resource_kubernetes_runtime_class_v1_test.go
```diff
@@ -0,0 +1,142 @@
...
+				),
+			},
+			{
+				Config: testAccKubernetesruntime_class_v1_addLabels(rcName),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckKubernetesruntime_class_v1Exists("kubernetes_runtime_class_v1.test", &conf),
+					resource.TestCheckResourceAttr("kubernetes_runtime_class_v1.test", "metadata.0.annotations.%", "2"),
+					resource.TestCheckResourceAttr("kubernetes_runtime_class_v1.test", "metadata.0.annotations.TestAnnotationOne", "one"),
+					resource.TestCheckResourceAttr("kubernetes_runtime_class_v1.test", "metadata.0.annotations.Different", "1234"),
+					resource.TestCheckResourceAttr("kubernetes_runtime_class_v1.test", "metadata.0.labels.%", "2"),
```
arybolovlev, line 65: Here you validate that 2 labels were created, but in the resource itself you have 3 labels. Please either remove one label from the test resource or update the acceptance tests to check 3 labels(in this case, please add the third label here too).

kubernetes/resource_kubernetes_runtime_class_v1_test.go
```diff
@@ -0,0 +1,142 @@
...
+					resource.TestCheckResourceAttrSet("kubernetes_runtime_class_v1.test", "metadata.0.uid"),
+				),
+			},
+			{
+				Config: testAccKubernetesruntime_class_v1_addLabels(rcName),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckKubernetesruntime_class_v1Exists("kubernetes_runtime_class_v1.test", &conf),
+					resource.TestCheckResourceAttr("kubernetes_runtime_class_v1.test", "metadata.0.annotations.%", "2"),
+					resource.TestCheckResourceAttr("kubernetes_runtime_class_v1.test", "metadata.0.annotations.TestAnnotationOne", "one"),
+					resource.TestCheckResourceAttr("kubernetes_runtime_class_v1.test", "metadata.0.annotations.Different", "1234"),
```
arybolovlev, line 64: In the test resource you have the second annotation `TestAnnotationTwo` with value `two`. Please update it here.

kubernetes/resource_kubernetes_runtime_class_v1_test.go
```diff
@@ -0,0 +1,142 @@
...
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"metadata.0.resource_version"},
+			},
+			{
+				Config: testAccKubernetesruntime_class_v1_addAnnotations(rcName),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckKubernetesruntime_class_v1Exists("kubernetes_runtime_class_v1.test", &conf),
+					resource.TestCheckResourceAttr("kubernetes_runtime_class_v1.test", "metadata.0.annotations.%", "2"),
```
arybolovlev, line 48: Could you please replace `"kubernetes_runtime_class_v1.test"` with the variable `resourceName` as you have it done on the first step? I commented it here, but please take a look on all steps. Thanks!

arybolovlev (changes requested):
Great job, @sheneska!

I have left a few comments regarding the acceptance tests. Once you address them, please make sure that they pass on your local cluster:

```console
$ make testacc TESTARGS="-count 1 -run ^TestAccKubernetesruntime_class_v1_basic"
```

Please add the changelog file `.changelog/2080.txt`.

Let me know if you have any questions or need any assistance. Thanks!

website/docs/r/runtime_class_v1.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+  A runtime class is used to determine which container runtime is used to run all containers in a pod. 
+---
+
+# kubernetes_runtime_class_v1
+
+A runtime class is used to determine which container runtime is used to run all containers in a pod.
+
+
+## Example usage
+
```
BBBmau, line 15: Would add a simple example code block of using the `kubernetes_runtime_class_v1`

kubernetes/resource_kubernetes_runtime_class_v1_test.go
```diff
@@ -0,0 +1,168 @@
...
+		},
+	})
+}
+
+func testAccKubernetesruntime_class_v1_basic(name string) string {
+	return fmt.Sprintf(`resource "kubernetes_runtime_class_v1" "test" {
+  metadata {
+    name = %q
+  }
+  handler = "myclass"
```
BBBmau, line 85: Since we use the runtime class resource with pods, we should have a pod using a runtimeclass resource as part of the tests.

BBBmau (dismissed):
Looks Good, just left some things to add such as Docs and adding pod tests

kubernetes/resource_kubernetes_runtime_class_v1_test.go
jrhouston: You don't need to create a Pod to actually test that this resource works. A RuntimeClass can exist without any Pods using it.

kubernetes/resource_kubernetes_runtime_class_v1.go
```diff
@@ -0,0 +1,177 @@
...
+		UpdateContext: resourceKubernetesRuntimeClassV1Update,
+		DeleteContext: resourceKubernetesRuntimeClassV1Delete,
+
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"metadata": metadataSchema("runtimeclass", true),
+
```
jrhouston, line 35: What about the `overhead` and `scheduling` fields? See API ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#runtimeclass-v1-node-k8s-io

jrhouston (changes requested):
Wondering about `overhead` and `scheduling` being missing – otherwise this looks good 🚀

kubernetes/resource_kubernetes_runtime_class_v1.go
sheneska: Sacha had recommended that i focus on the handler at first and get it working perfectly, he also mentioned that since the handler is will be the most commonly used we should merge this first and then add `overhead` and `scheduling` afterwards

kubernetes/resource_kubernetes_runtime_class_v1_test.go
BBBmau: @jrhouston wouldn't we want to have a test that checks that it works as expected with a Pod? I left this suggestion since not having a proper handler value would result in the Pod not being created. We could always have a message that informs the user what values to use to prevent this from happening.

kubernetes/resource_kubernetes_runtime_class_v1.go
```diff
@@ -0,0 +1,177 @@
...
+			"handler": {
+				Type:         schema.TypeString,
+				Description:  "Specifies the underlying runtime and configuration that the CRI implementation will use to handle pods of this class",
+				Required:     true,
+				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?`), ""),
```
jrhouston, line 40: Kubernetes actually has a utility function to validate RFC 1123 DNS labels: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Label


So we don't need to roll our own regex here.

kubernetes/resource_kubernetes_runtime_class_v1_test.go
jrhouston: The schema already has validation for the `handler` field so they will get an error message if the handler value is not in the proper format. 

Other than that the API server doesn't produce any error if you supply a handler that doesn't exist on the node. So we don't really have a way of informing them that they've specified a bad handler value.

jrhouston (approved):

PR #1141, opened 2021-02-03
Add `ConflictsWith` to provider authentication config

dak1n1:
### Description

In the past, we relied on Kubernetes client-go to determine the order of precedence when given a list of mutually-exclusive parameters for authenticating to the Kubernetes cluster. This lead to some confusion, since client-go's decision could override even an explicit provider configuration. For example, if a `token` were specified in the Kubernetes provider config, *and* the environment variable `KUBE_CONFIG_PATH` was present, client-go would silently decide to use the `KUBE_CONFIG_PATH` to authenticate to the cluster.

In order to create a more explicit configuration, and more predictable behavior, this PR adds `ConflictsWith` to specific provider authentication attributes, effectively disallowing the use of conflicting methods of authentication, such as `config_path` and `token`. The provider now generates an error to tell the user that these options are mutually exclusive.

### Acceptance tests
- [ ] Have you added an acceptance test for the functionality being added?
- [ ] Have you run the acceptance tests on this branch?

Output from acceptance testing:

```
$ TESTARGS="-run 'TestAccKubernetesProviderConfig'" make testacc
=== RUN   TestAccKubernetesProviderConfig_config_path
--- PASS: TestAccKubernetesProviderConfig_config_path (3.68s)
=== RUN   TestAccKubernetesProviderConfig_config_paths
--- PASS: TestAccKubernetesProviderConfig_config_paths (2.90s)
=== RUN   TestAccKubernetesProviderConfig_config_paths_env
--- PASS: TestAccKubernetesProviderConfig_config_paths_env (2.33s)
=== RUN   TestAccKubernetesProviderConfig_config_paths_env_wantError
--- PASS: TestAccKubernetesProviderConfig_config_paths_env_wantError (0.43s)
=== RUN   TestAccKubernetesProviderConfig_host_env_wantError
--- PASS: TestAccKubernetesProviderConfig_host_env_wantError (0.55s)
=== RUN   TestAccKubernetesProviderConfig_host
--- PASS: TestAccKubernetesProviderConfig_host (8.68s)
PASS
```

### Release Note
Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-kubernetes/blob/master/CHANGELOG.md):

```release-note
Display an error when the provider is configured with mutually-exclusive authentication options.
```

### Community Note

[...]

Issue #1973, opened 2023-01-26
Support KUBECONFIG environment variable


reegnz:
### Description

It is a common expectation that tools that talk to the kubernetes API can be configured with the KUBECONFIG environment variable. When a tool uses the kubernetes client libraries, it gets that functionality out of the box. The terraform kubernetes provider however does not seem to be using the official kubernetes go client, and it's using the KUBE_CONFIG_PATH and KUBE_CONFIG_PATHS environment variables.

The kubernetes provider should also start supporting the KUBECONFIG variable, to meet the common expectation that setting this environment variable will allow all tools to work with the given selected cluster (unless the provider is explicitly configured to use a different configuration).

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

txomon:
This is a bad idea because you might be managing different clusters that are not in your kubeconfig, nor have access to them, through terraform.

reegnz:
@txomon The same is true for kubectl. Should kubectl not have KUBECONFIG support either?
The request is not about using KUBECONFIG environment variables exclusively, but to introduce a common way of configuring a connection to a kubernetes cluster.

txomon:
So the proposal would be to have KUBECONFIG support if no other configuration variable would be available?

Just in case, this is already how the system works.

Also, it seems like my tone was a bit off in my comment, apologies for that.

reegnz:
Pretty much the same behavior that KUBE_CONFIG_PATH does, but supporting the more standard KUBECONFIG environment variable for the same purpose as well.

reegnz:
There's one caveat I'm seeing that would be a bit different to `KUBE_CONFIG_PATH`:
The `KUBECONFIG` variable can hold multiple kubeconfigs separated by `:`, while the terraform provider has a separate `KUBE_CONFIG_PATHS` environment variable if you have multiple kubeconfig files. I think the logic to collapse multiple files into a single effective config is also done differently by this terraform provider, so supporting multiple configs in `KUBECONFIG` might be difficult to achieve as it would need a brand new code branch to keep it in line with how `KUBECONFIG` env var has to be consumed, as it might not be compatible with the existing `KUBE_CONFIG_PATHS` mechanism.

[...]

Issue #1175, opened 2021-02-24
Source config_path provider argument from $KUBECONFIG env var


drewgingerich:
### Description

Can the `config_path` provider argument be sourced from `$KUBECONFIG`? The `kubectl` command uses `KUBECONFIG` by default so it makes sense to me that this provider would look in the same place. I think this could provide a smoother integration with other tools via consistent defaults.

My specific use-case is with Gitlab CI. Gitlab CI has a nice k8s cluster integration that automatically exports the `KUBECONFG` env var for each CI job. Because this provider looks for `KUBE_CONFIG_PATH`, I have to add some extra logic to re-export the kube config path. This sounds simple, but is made a bit more complex by what I'd consider a bug in Gitlab CI: adding

```
variables:
   KUBE_CONFIG_PATH: $KUBECONFIG
```

to my `.gitlab-ci.yml` script actually exports the contents of the kube config file. The solution I've found is to add a script to the container running the job, to properly export `KUBE_CONFIG_PATH` before running Terraform. 

To maintain backward compatibility, sourcing from `KUBECONFIG` would be in addition to sourcing from `KUBE_CONFIG_PATH`. `KUBE_CONFIG_PATH` could have priority to help prevent surprises. 

I'm wondering if this has been looked at before. The choice of `KUBE_CONFIG_PATH` certainly feels intentional. If so, why was it decided not to use `KUBECONFIG`?

If this sounds reasonable, I can work on it  and submit a PR. 

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

dak1n1:
Thanks for writing up this issue! It was an intentional decision with the 2.0 release to remove the automatic reading of `KUBECONFIG`, since many users were having issues with the wrong cluster being targeted by the provider due to the `KUBECONFIG` being read. However, this is something we're open to revisiting, if there is enough support from the community.

[...]

PR #2084, opened 2023-04-21
add `conflictsWith` to provider schema

BBBmau:
### Description

This resolves the issue that comes up when wanting to use 2 k8s configurations 

Relates to #1141, updates the original PR

### Acceptance tests

```

...
```

### Release Note

```release-note
...
```

### Community Note


kubernetes/provider.go
```diff
@@ -40,46 +40,58 @@ func Provider() *schema.Provider {
...
-				Type:        schema.TypeString,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("KUBE_CLIENT_KEY_DATA", ""),
-				Description: "PEM-encoded client certificate key for TLS authentication.",
+				Type:          schema.TypeString,
+				Optional:      true,
+				DefaultFunc:   schema.EnvDefaultFunc("KUBE_CLIENT_KEY_DATA", nil),
+				Description:   "PEM-encoded client certificate key for TLS authentication.",
+				ConflictsWith: []string{"config_path", "config_paths", "username", "password", "exec", "insecure"},
+				RequiredWith:  []string{"client_certificate", "cluster_ca_certificate", "host"},
```
alexsomesan, line 86: `cluster_ca_certificate` isn't related to the client certificate credentials. It's pretty much independent and shouldn't depend on be required by any of the attributes, AFAIK.

kubernetes/provider.go
```diff
@@ -40,46 +40,57 @@ func Provider() *schema.Provider {
...
 			"client_key": {
-				Type:        schema.TypeString,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("KUBE_CLIENT_KEY_DATA", ""),
-				Description: "PEM-encoded client certificate key for TLS authentication.",
+				Type:          schema.TypeString,
+				Optional:      true,
+				DefaultFunc:   schema.EnvDefaultFunc("KUBE_CLIENT_KEY_DATA", nil),
+				Description:   "PEM-encoded client certificate key for TLS authentication.",
+				ConflictsWith: []string{"config_path", "config_paths", "username", "password", "exec", "insecure"},
```
alexsomesan, line 85: I'm not 100% sure that `insecure` is a conflict in this case. TLS certs can be client-side only in which case they aren't involved in any server auth. The `insecure` flag would still be needed if the API doesn't have a trusted server-side cert.

kubernetes/provider.go
```diff
@@ -40,46 +40,57 @@ func Provider() *schema.Provider {
...
 			"client_certificate": {
-				Type:        schema.TypeString,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("KUBE_CLIENT_CERT_DATA", ""),
-				Description: "PEM-encoded client certificate for TLS authentication.",
+				Type:          schema.TypeString,
+				Optional:      true,
+				DefaultFunc:   schema.EnvDefaultFunc("KUBE_CLIENT_CERT_DATA", nil),
+				Description:   "PEM-encoded client certificate for TLS authentication.",
+				ConflictsWith: []string{"config_path", "config_paths", "username", "password", "insecure"},
```
alexsomesan, line 77: I'm not 100% sure that `insecure` is a conflict in this case. TLS certs can be client-side only in which case they aren't involved in any server auth. The `insecure` flag would still be needed if the API doesn't have a trusted server-side cert.

kubernetes/provider.go
```diff
@@ -40,46 +40,57 @@ func Provider() *schema.Provider {
...
 			"insecure": {
-				Type:        schema.TypeBool,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("KUBE_INSECURE", false),
-				Description: "Whether server should be accessed without verifying the TLS certificate.",
+				Type:          schema.TypeBool,
+				Optional:      true,
+				DefaultFunc:   schema.EnvDefaultFunc("KUBE_INSECURE", nil),
+				Description:   "Whether server should be accessed without verifying the TLS certificate.",
+				ConflictsWith: []string{"client_key", "client_certificate", "exec"},
```
alexsomesan, line 70: See comments on client_cert

kubernetes/provider.go
```diff
@@ -40,46 +40,57 @@ func Provider() *schema.Provider {
...
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("KUBE_CLUSTER_CA_CERT_DATA", ""),
-				Description: "PEM-encoded root certificates bundle for TLS authentication.",
+				Type:          schema.TypeString,
+				Optional:      true,
+				DefaultFunc:   schema.EnvDefaultFunc("KUBE_CLUSTER_CA_CERT_DATA", nil),
+				Description:   "PEM-encoded root certificates bundle for TLS authentication.",
+				ConflictsWith: []string{"config_path", "config_paths", "insecure"},
 			},
 			"config_paths": {
```
alexsomesan, line 95: There are a couple of attributes above that include this in their `conflicts_with` lists. I wonder if these should be mutually exclusive and they would also need to be mentioned here... I'm not 100% sure.

BBBmau: The way it's designed is to have `ConflictsWith` / `RequiredWith` to mutually exclusive between attributes. I was able to find an example from another provider that shows a good example of [this](https://github.com/hashicorp/terraform-provider-nomad/blob/77c12b04db15efd9f7bc84d13f229e9bba861c1c/nomad/provider.go#L47)

kubernetes/provider.go
alexsomesan: @BBBmau I think this needs changing. `client_certificate`, `client_key` and `exec` are not incompatible with `insecure`
IMO, only `cluster_ca_certificate` is a conflict.

kubernetes/provider.go
```diff
@@ -40,46 +40,57 @@ func Provider() *schema.Provider {
...
-				Type:        schema.TypeString,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("KUBE_CLIENT_KEY_DATA", ""),
-				Description: "PEM-encoded client certificate key for TLS authentication.",
+				Type:          schema.TypeString,
+				Optional:      true,
+				DefaultFunc:   schema.EnvDefaultFunc("KUBE_CLIENT_KEY_DATA", nil),
+				Description:   "PEM-encoded client certificate key for TLS authentication.",
+				ConflictsWith: []string{"config_path", "config_paths", "username", "password", "exec"},
+				RequiredWith:  []string{"client_certificate", "host"},
```
alexsomesan, line 86: same here, "host" isn't a requirement for the same reason as the cert.

kubernetes/provider.go
```diff
@@ -40,46 +40,57 @@ func Provider() *schema.Provider {
 	p := &schema.Provider{
 		Schema: map[string]*schema.Schema{
 			"host": {
```
jrhouston, line 42: When i tried this out I was still able to set both `host` and `config_path`.

I think you need to set ConflictsWith for both attributes, i.e you need to add `host` to the ConflictsWith for `config_path`. You'll need to this for the other attributes you've added this too as well.

jrhouston (changes requested):

jrhouston (approved):

aquarhead:
While I totally agree this change is technically correct, the old behaviour has a useful side effect.

What I've been doing for years is to only set `host` in the provider configuration:

```
provider "kubernetes" {
  host = "..."
}
```

The actual auth will be configured through env vars (e.g. I use `KUBE_CONFIG_PATH` and `KUBE_CTX` env var). And if the kube cluster the auth/context refers to is different than the `host`, it will error out thus preventing deploying things in the wrong kube cluster by accident. (The error is actually something related to certificate, but once you know what it means it serves the purpose pretty well)

This is useful in the sense that it allows Terraform code to enforce the exact cluster a config is supposed to be used, while at the same time allowing auth information to be configured through potentially different manners, not just between developers, but between local and CI as well (which iirc was certainly a case at my previous job). (And yes it did happen that we used a wrong config in the wrong cluster, hence I looked around for way to prevent it)

I'm a bit upset that this is not released as a major version even though I'll be the first to admit I was abusing the behaviour, still, I think the ability to enforce the correct cluster without hardcoding the auth has some merit.

I would like to propose an alternative solution, instead of using `conflictsWith` directly, could we try to merge all the different config options, and error out if anything mismatch? In my use case this will actually give useful error message like "`host` has conflict with context" but only if the wrong context is loaded.

---

Before you ask, I run EKS and I'm aware of the `exec` config method. But in general I'd like Terraform code itself to not enforce how authentication to be done. Also, if the terraform provider is fully configured, there could be a situation where the Terraform operates on a different cluster than `kubectl` is, that can be quite confusing - it happens a lot that you deploy through tf and then check things through `kubectl`. I think it's nice to configure both at the same time, e.g. by pointing to a context.

---

I want to stress again I totally see where this change is coming from, but I want to explain my (ab)use case and I do believe it will be helfpul especially for people operating many kube clusters, to be able to enforce the correct cluster in code, while keeping the auth flexible. (Maybe, we could introduce a `host_match` or probably `certificate_match` would be more correct? :thinking:)

I will totally understand if it doesn't happen, but I do wish that this change be reverted in this major version and explore some other options instead, or introduce only in the next major version. We do not use Terraform lockfiles because of a whole slew of other issues, and in general we would like to use latest providers (locked to major version).

Pell17:
Breaking change for us. Would have been nice if this could have initially been a Warning instead of Error, giving some time to "fix" our configuration. We usually auto-upgrade providers to the newest minor version, not expecting things to break, as required by SemVer.

emoshaya:
breaking change for us too :(

Nuru:
I have a few problems with this.

As others have mentioned, this is a breaking change in a minor version release, which violates SemVer, and has caused a lot of disruption with no notice.

The documentation has not been updated to indicate what conflicts with what. For example, does `username` conflict with `client_key`? 

This change removes our ability to override environment variables. In [#1175](https://github.com/hashicorp/terraform-provider-kubernetes/issues/1175#issuecomment-786121967), @dak1n1 explained that v2 of this provider consciously avoided taking configuration from the `KUBECONFIG` environment variable to avoid having it direct Terraform to the wrong cluster accidentally. (This provider uses `KUBE_CONFIG_PATH` and `KUBE_CONFIG_PATHS` instead.) In our modules, we want to further ensure the environment variables are not used, and we have been doing that by setting `config_path = ""`. This change breaks that.

After we fix the breakage by removing `config_path`, now when both `host` and `KUBE_CONFIG_PATH` are set, we get the confusing error message:

```
│ Error: Conflicting configuration arguments
│ 
│   with provider["registry.terraform.io/hashicorp/kubernetes"],
│   on providers.tf line 14, in provider "kubernetes":
│  14:   host                   = local.eks_cluster_endpoint
│ 
│ "host": conflicts with config_path
╵
```

(Confusing because `config_path` is not present in the providers block. A more helpful error would be "host conflicts with KUBE_CONFIG_PATH".)


### Suggestion

I believe the better path forward is to arrange the configuration options in a hierarchy, rather than cause them to conflict. Ideally the provider would use configuration provided directly in the provider block in preference to environment variables, and not complain (or perhaps issue a warning) if the environment variables provide a different configuration than what is specified in the provider block. This would allow the provider to support `KUBECONFIG` as has been requested more than once (#1973, #1175, and perhaps more) while avoiding the problem that environment variables override specifically configured authentication. But perhaps that is best deferred until v3 of this provider.

Until v3 is ready, please revert this change and discuss backwards compatible improvements instead.

jrhouston:
Thank you for your thoughtful responses @aquarhead @Pell17 @Nuru – we are going to revert this change for now. We will put out a hotfix release once the above PR is merged.

ingnil:
This change also does not care about the value of the parameters. I just ran into a conflict between `cluster_ca_certificate` and `insecure` despite the latter being set to `false`.

PR #2166, opened 2023-06-29
Add lint-type-assertion check for `/manifest`

BBBmau:
### Acceptance tests

```

...
```

### Release Note

```release-note
...
```

### Community Note


scripts/errcheck.sh
```diff
@@ -0,0 +1,28 @@
+# Check gofmt
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+echo "==> Checking for unchecked errors..."
+
+if ! which errcheck > /dev/null; then
+    echo "==> Installing errcheck..."
+    go get -u github.com/kisielk/errcheck
```
alexsomesan, line 9: Using `go get` like this will install errcheck as a dependency in the current go module (the provider) and will also modify the provider's `go.mod` file in the process.

Instead, use `go install github.com/kisielk/errcheck@latest`. This will install the binary under $GOPATH/bin and not affect the provider Go module. It's important here to have the version designator in place (`@latest`), otherwise it will still get installed as a provider dependency.

Check out `go help install` for a lot more detailed explanation.

scripts/errcheck.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+err_files=$(errcheck -exclude scripts/errcheck_excludes.txt \
```
alexsomesan, line 12: The bin location might not always be in the PATH, but it's easy to reference the binary directly like this:
```suggestion
err_files=$($(go env GOPATH)/bin/errcheck -exclude scripts/errcheck_excludes.txt \
```

alexsomesan (commented):
Tried it out locally but it's not building the errcheck binary as expected. Left some comments with the changes I think are needed to get it working reliably.

arybolovlev (approved):

PR #2193, opened 2023-07-18
Fix formatting for Markdown links

vastep:
P.S. Many links like `docs/user-guide` are stale and will be replaced in folowing PRs.


vastep:
How many weeks are needed to simply scroll down 7 times?

sheneska (approved):
👍🏻

PR #2200, opened 2023-07-24
Make target pod state / phase configurable

arybolovlev:
### Description

Replace Pod resource option `legacy_lifecycle_states` to `target_state`. It allows to specify the Pod phase(s) that indicate whether it was successfully created. Default to `Running`. It allows all 5 currently available phases even though some of them may not be an indication of a successfully created pod.

**Links:**

- https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

### Acceptance tests
    - All tests: hhttps://github.com/hashicorp/terraform-provider-kubernetes/actions/runs/5691661382/job/15427262240
    - Pod tests: https://github.com/hashicorp/terraform-provider-kubernetes/actions/runs/5691660004/job/15427258099

### Release Note

```release-note:enhancement
`resource/kubernetes_pod`: add a new attribute `target_state` to specify the Pod phase(s) that indicate whether it was successfully created.

`resource/kubernetes_pod_v1`: add a new attribute `target_state` to specify the Pod phase(s) that indicate whether it was successfully created.
```

### References

Enhance: https://github.com/hashicorp/terraform-provider-kubernetes/pull/2141

### Community Note

### DIFF
--- a/kubernetes/structures_pod.go
+++ b/kubernetes/structures_pod.go
@@ -661,6 +661,18 @@ func flattenReadinessGates(in []v1.PodReadinessGate) ([]interface{}, error) {
 
 // Expanders
 
+func expandPodTargetState(p []interface{}) []string {
+	if len(p) > 0 {
+		t := make([]string, len(p))
+		for i, v := range p {
+			t[i] = v.(string)
+		}
+		return t
+	}
+
+	return []string{string(v1.PodRunning)}
+}
+
 func expandPodSpec(p []interface{}) (*v1.PodSpec, error) {
 	obj := &v1.PodSpec{}
 	if len(p) == 0 || p[0] == nil {
--- a/.changelog/2200.txt
+++ b/.changelog/2200.txt
@@ -1,0 +1,7 @@
+```release-note:enhancement
+`resource/kubernetes_pod`: add a new attribute `target_state` to specify the Pod phase(s) that indicate whether it was successfully created.
+```
+
+```release-note:enhancement
+`resource/kubernetes_pod_v1`: add a new attribute `target_state` to specify the Pod phase(s) that indicate whether it was successfully created.
+```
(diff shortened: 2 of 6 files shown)


kubernetes/structures_pod.go
```diff
@@ -661,6 +661,18 @@ func flattenReadinessGates(in []v1.PodReadinessGate) ([]interface{}, error) {
...
+func expandPodTargetState(p []interface{}) []string {
+	if len(p) > 0 {
+		t := make([]string, len(p))
+		for i, v := range p {
+			t[i] = v.(string)
+		}
+		return t
+	}
+
+	return []string{string(v1.PodRunning)}
```
alexsomesan, line 673: Shouldn't this rather be done with a `Default:` or `DefaultFunc:` attribute in the schema? 
Feels like setting the default here makes it less obvious that there is a default value.

arybolovlev: This is actually what I have stuck with and ended up with the current solution:

- I can't do `Default: []string{string(corev1.PodRunning)}`. In this case, `apply` ends up with the error message saying that `Default` is not valid for lists or sets.
- When I do `Default: string(corev1.PodRunning),` on the `Elem` level, it doesn't work too. This is what `Default` documentation says about this **_Default cannot be used if the Schema is directly an implementation of an Elem field of another Schema, such as trying to set a default value for a TypeList or TypeSet._**
- `DefaultFunc` doesn't work too. I tried `DefaultFunc: func() (interface{}, error) {return []string{string(corev1.PodRunning)}, nil}` on the List level and `DefaultFunc: func() (interface{}, error) {return string(corev1.PodRunning), nil}` on the `Elem` level. In both cases, `d.Get("target_state")` returns an empty slice.

Maybe I miss something.

alexsomesan: I don't think you're missing anything. The List and Set types in the SDK have always had problems with defaults.

Let's just keep it the way you did it then, it also covers the case where `target_state` is set to an empty string or nil.

alexsomesan (approved):
LGTM. Thanks for taking another look at this.

vastep:
Is `__debug_bin` added by mistake?

PR #2307, opened 2023-10-06
Add os tests

BBBmau:
### Description

OS attribute was missing tests and was not working properly as expected.

### Acceptance tests

```
 mau@mau-JKDT676NCP  ~/Dev/terraform-provider-kubernetes   add-os-tests ● ⍟4  make testacc TESTARGS="-run TestAccKubernetesPodV1_basic"
go vet .
TF_ACC=1 go test "/Users/mau/Dev/terraform-provider-kubernetes/kubernetes" -v -vet=off -run TestAccKubernetesPodV1_basic -parallel 8 -timeout 3h
# github.com/hashicorp/terraform-provider-kubernetes/kubernetes.test
ld: warning: -bind_at_load is deprecated on macOS
=== RUN   TestAccKubernetesPodV1_basic
=== PAUSE TestAccKubernetesPodV1_basic
=== CONT  TestAccKubernetesPodV1_basic
--- PASS: TestAccKubernetesPodV1_basic (10.30s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   11.410s
```

### Release Note

```release-note
...
```

### Community Note


kubernetes/schema_pod_spec.go
```diff
@@ -344,6 +344,30 @@ func podSpecFields(isUpdatable, isComputed bool) map[string]*schema.Schema {
 							Type: schema.TypeInt,
 						},
 					},
+					"windows_options": {
+						Type:        schema.TypeSet,
```
jrhouston, line 348: Let's make this a `TypeList` and set `MaxItems: 1` to follow the convention that's been used for other nested blocks in the pod spec.

kubernetes/schema_pod_spec.go
```diff
@@ -344,6 +344,30 @@ func podSpecFields(isUpdatable, isComputed bool) map[string]*schema.Schema {
...
+									Type:        schema.TypeString,
+									Description: "GMSACredentialSpecName is the name of the GMSA credential spec to use.",
+									Required:    true,
+								},
+								"run_as_username": {
+									Type:        schema.TypeString,
+									Description: "The UserName in Windows to run the entrypoint of the container process. Defaults to the user specified in image metadata if unspecified. May also be set in PodSecurityContext. If set in both SecurityContext and PodSecurityContext, the value specified in SecurityContext takes precedence.",
+									Optional:    true,
+								},
+							},
```
jrhouston, line 368: There is also a `host_process` field here.

kubernetes/schema_pod_spec.go
```diff
@@ -344,6 +344,30 @@ func podSpecFields(isUpdatable, isComputed bool) map[string]*schema.Schema {
...
+					"windows_options": {
+						Type:        schema.TypeSet,
+						Description: "The Windows specific settings applied to all containers. If unspecified, the options within a container's SecurityContext will be used. If set in both SecurityContext and PodSecurityContext, the value specified in SecurityContext takes precedence. Note that this field cannot be set when spec.os.name is linux.",
+						Optional:    true,
+						Elem: &schema.Resource{
+							Schema: map[string]*schema.Schema{
+								"gmsa_credential_spec": {
+									Type:        schema.TypeString,
+									Description: "GMSACredentialSpec is where the GMSA admission webhook inlines the contents of the GMSA credential spec named by the GMSACredentialSpecName field",
+									Required:    true,
```
jrhouston, line 356: How do you know this field and the one below are required? I don't see them marked as required in the K8s API docs.


edit: Looking at the type for that field, all the attributes seem to be optional: 

https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L5941-L5965

jrhouston (changes requested):

kubernetes/schema_pod_spec.go
```diff
@@ -210,7 +210,8 @@ func podSpecFields(isUpdatable, isComputed bool) map[string]*schema.Schema {
 				Schema: map[string]*schema.Schema{
 					"name": {
 						Type:         schema.TypeString,
-						Required:     true,
+						Optional:     true,
+						Default:      "linux",
```
alexsomesan, line 214: Since this attribute is optional, I'm wondering if we really need to set a default. Being optional, the API likely already tolerates missing values. I would follow the principle of least surprise here and leave the default value out.

alexsomesan (commented):
Looks good, but the default value got me thinking. See my spot comment there.

kubernetes/schema_pod_spec.go
arybolovlev: `spec.OS` is [optional](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3230C1-L3231) but `spec.OS.Name` is [required](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3321-L3327).

If we keep `name` optional with the default value `linux` we will always send this value to the Kubernetes API and Pods will have it set to `linux` which does not match the default Kubernetes behavior.  The default one -- if not sent, keep it empty. That is why now we return [`nil`](https://github.com/hashicorp/terraform-provider-kubernetes/blob/main/kubernetes/structures_pod.go#L710-L712) if `os.name` is not set.

I agree with what Alex said, I would keep it as it is now.

arybolovlev: I think the test failures observed here are because of the replacement of the `expandOS` function. I would keep the one we have right now.

We might want to update it though.

```golang
func expandOS(l []interface{}) *v1.PodOS {
	if len(l) == 0 || l[0] == nil {
		return nil
	}

	in := l[0].(map[string]interface{})
	obj := &v1.PodOS{}

	if v, ok := in["name"].(string); ok {
		obj.Name = v1.OSName(v)
	}

    // If the above assertion fails and we return the PodOS structure with an empty Name field
    // then that will cause an error message from the Kubernetes API. The name cannot be empty.
    // We either need to return obj within the if statement and return nil here 
    // or introduce an error return if the above assertion fails.
    // We even can think whether we need assertion since the name would be a string
    // due to its type in the schema and it would be one of two values since we do validation.
	return obj
}
```

I left some comments in the code snippet.

I think the current expandOS might potentially look like this:

```golang
func expandOS(l []interface{}) *v1.PodOS {
	if len(l) == 0 || l[0] == nil {
		return nil
	}

	in := l[0].(map[string]interface{})

	return &v1.PodOS{
		Name: v1.OSName(in["name"].(string)),
	}
}
```

kubernetes/resource_kubernetes_pod_v1_test.go
```diff
@@ -76,6 +76,7 @@ func TestAccKubernetesPodV1_basic(t *testing.T) {
 					resource.TestCheckResourceAttrSet(resourceName, "metadata.0.generation"),
 					resource.TestCheckResourceAttrSet(resourceName, "metadata.0.resource_version"),
 					resource.TestCheckResourceAttrSet(resourceName, "metadata.0.uid"),
+					resource.TestCheckResourceAttr(resourceName, "spec.0.os.0.name", "linux"),
```
arybolovlev, line 79: I think this could be removed from the basic test since we have a [separated test](https://github.com/hashicorp/terraform-provider-kubernetes/blob/main/kubernetes/resource_kubernetes_pod_v1_test.go#L1572) to validate `os`. The OS tests can be extended to validate other fields that depend on OS one, that is the motivation to keep it separated.

kubernetes/resource_kubernetes_pod_v1_test.go
```diff
@@ -1757,6 +1758,9 @@ resource "kubernetes_pod_v1" "test" {
 
   spec {
     automount_service_account_token = false
+    os {
+      name = "linux"
+    }
```
arybolovlev, line 1763: Same as https://github.com/hashicorp/terraform-provider-kubernetes/pull/2307/files#r1365201482

jrhouston (approved):

kubernetes/schema_pod_spec.go
BBBmau: @jrhouston this was overlooked on my part but I found the reason why I had marked it as `Required` in the first place. When using the `windows_options` block as optional and using it, we are given this error when it isn't used:

```
│ Error: Pod "terraform-example" is invalid: [spec.securityContext.windowsOptions.gmsaCredentialSpecName: Invalid value: "": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.securityContext.windowsOptions.gmsaCredentialSpec: Invalid value: "": gmsaCredentialSpec cannot be an empty string]
```

leading to the need for it to be `Required` instead of `Optional` since we can't just leave it as empty for both `gmsaCredentialSpec` and `gmsaCredentialSpecName`

I'll make this update in an upcoming PR that adds `windows_options` to the container block also.

PR #2364, opened 2023-12-08
🧹 Remove unnecessary error returns and handlings

arybolovlev:
### Description

We have a bunch of functions that, according to their signatures, are supposed to return an error, but all return statements within these functions stay with `nil` only at the error place. In this case, it doesn't make much sense to keep them and handle an error on return.

This PR aims to remove unnecessary error returns as well as error handling, as no errors are returned anymore.

### Release Note

```release-note
NONE.
```

### References

N/A.

### Community Note

### DIFF
--- a/kubernetes/structure_network_policy.go
+++ b/kubernetes/structure_network_policy.go
@@ -131,11 +131,8 @@ func expandNetworkPolicyV1Spec(in []interface{}) (*networkingv1.NetworkPolicySpe
 		}
 		spec.Egress = *egress
 	}
-	policyTypes, err := expandNetworkPolicyV1Types(m["policy_types"].([]interface{}))
-	if err != nil {
-		return nil, err
-	}
-	spec.PolicyTypes = *policyTypes
+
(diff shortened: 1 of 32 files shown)


kubernetes/structure_network_policy.go
```diff
@@ -214,7 +203,7 @@ func expandNetworkPolicyV1Ports(l []interface{}) (*[]networkingv1.NetworkPolicyP
 			policyPorts[i].Protocol = &v
 		}
 	}
-	return &policyPorts, nil
+	return policyPorts
```
SarahFrench, line 206: nit: Changing the function to also no longer return a pointer could/should have been a separate PR

SarahFrench (approved):
This joke might only work for a UK audience but I'm going for it

kubernetes/structure_network_policy.go
arybolovlev: Reverted this change. Thanks!

PR #2428, opened 2024-02-20
Add provider defined functions for encoding and decoding Kubernetes manifests

jrhouston:
### Description

This PR introduces the following provider defined functions:
- `manifest_encode` to encode a Terraform object representation of a Kubernetes manifest as a YAML string.
- `manifest_decode` to decode a string containing a Kubernetes manifest in YAML format and return it as a Terraform object.
- `manifest_decode_multi` To do the above but for YAML manifests containing multiple resources.

Solves https://github.com/hashicorp/terraform/issues/29729

### TODO
- [x] decoding
    - [x] validate yaml is a manifest
    - [x] error handling / diagnostics
    - [x] happy path test case
    - [x] edge case test case
    - [x] error handling test case
    - [x] manifest_decode_multi
    - [x] manifest_decode
    - [x] documentation page
    - [x] Remove `context.TODO()` 
- [x] encoding
  - [x] manifest_encode
  - [x] validate object is a manifest
  - [x] error handling / diagnostics
  - [x] test cases 
  - [x] documentation page
- [x] move functions directory into `/internal/framework/provider` directory
- [x] Change tests to use state checker once this PR is merged: https://github.com/hashicorp/terraform-plugin-testing/pull/276
- [x] Change `go.mod` to reference latest plugin framework once this is merged: https://github.com/hashicorp/terraform-plugin-framework/pull/931

```

...
```

### Release Note

```release-note
...
```

### Community Note


internal/framework/provider/functions/manifest_decode_multi.go
```diff
@@ -0,0 +1,155 @@
...
+
+type ManifestDecodeMultiFunction struct{}
+
+func (f ManifestDecodeMultiFunction) Metadata(_ context.Context, req function.MetadataRequest, resp *function.MetadataResponse) {
+	resp.Name = "manifest_decode_multi"
+}
+
+func (f ManifestDecodeMultiFunction) Definition(_ context.Context, req function.DefinitionRequest, resp *function.DefinitionResponse) {
+	resp.Definition = function.Definition{
+		Summary:             "manifest_decode_multi Function",
```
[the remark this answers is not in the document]
jrhouston: Nice, thanks for the example! Have updated the strings to follow that pattern.

internal/framework/provider/functions/manifest_decode_multi_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+	outputName := "test"
+
+	resource.UnitTest(t, resource.TestCase{
+		ProtoV5ProviderFactories: testAccProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testManifestDecodeMultiConfig("testdata/decode_single.yaml"),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					// FIXME: terraform-plugin-testing doesn't support dynamic yet
```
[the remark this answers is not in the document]
jrhouston: That `statecheck` package looks awesome, I've added a TODO above to use that once it's merged.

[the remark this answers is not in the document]
austinvalle: And it's out! https://github.com/hashicorp/terraform-plugin-testing/releases/tag/v1.7.0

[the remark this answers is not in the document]
jrhouston: Awesome! Will update 😄

internal/framework/provider/functions/manifest_decode_multi_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	outputName := "test"
+
+	resource.UnitTest(t, resource.TestCase{
+		ProtoV5ProviderFactories: testAccProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testManifestDecodeMultiConfig("testdata/decode_single.yaml"),
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownOutputValue(outputName, knownvalue.ListExact([]knownvalue.Check{
```
austinvalle, line 30: Not released yet, but in `terraform-plugin-testing` you can use the essentially equivalent `TupleExact` for a little better readability 👍🏻

internal/framework/provider/functions/manifest_decode_multi_test.go
```diff
@@ -0,0 +1,219 @@
...
+								"name": knownvalue.StringExact("test-configmap"),
+							}),
+						}),
+					})),
+				},
+			},
+			{
+				Config: testManifestDecodeMultiConfig("testdata/decode_multi.yaml"),
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownOutputValue(outputName, knownvalue.ListExact([]knownvalue.Check{
```
austinvalle, line 50: Curious, did you run into this panic bug? (since this is returning a tuple): https://github.com/hashicorp/terraform-plugin-testing/issues/310

I'm wondering if your tuple elements have equivalent types and that's why you didn't run into a panic here 🤔 

Either way, just an FYI that this bug does exist and is fixed, planned for release in `terraform-plugin-testing v1.8.0`

internal/framework/provider/functions/encode.go
```diff
@@ -0,0 +1,153 @@
...
+	switch vv := v.(type) {
+	case basetypes.StringValue:
+		return vv.ValueString(), nil
+	case basetypes.NumberValue:
+		f, _ := vv.ValueBigFloat().Float64()
+		return f, nil
+	case basetypes.Float64Value:
+		return vv.ValueFloat64(), nil
+	case basetypes.Int64Value:
+		return vv.ValueInt64(), nil
```
austinvalle, line 27: Just an FYI, the only possible branch for all number types is `basetypes.NumberValue` since floats/ints are both represented in Terraform by the underlying `tftypes.Number`.

The logic for how we determine what framework type something should be is here: https://github.com/hashicorp/terraform-plugin-framework/blob/484f311c99cf62e7bbf638ede03e1c7df97151db/types/basetypes/dynamic_type.go#L116

internal/framework/provider/functions/encode.go
```diff
@@ -0,0 +1,153 @@
...
+	m := make(map[string]any, len(elems))
+	for k, v := range elems {
+		var err error
+		m[k], err = encodeValue(v)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return m, nil
+}
```
austinvalle, line 111: Another FYI for possible use-cases, the `AllowNullValues` (implicitly set to false for this function) won't prevent null element values (for list, map, set, tuple) or null attribute values (for objects). If that changes how you want to decode things, might be worth adding some test scenarios for that, otherwise you'll get default values from the `Value*()` functions (string for example, null -> "")

jrhouston: This is good to know, in our case we want to actually render a `null` in the YAML and not have it result in a default/empty value for that type.

internal/framework/provider/functions/encode.go
jrhouston: Ah ha, interesting thanks! I've removed the cases for Int64 and Float64 and sure enough my tests with floats and ints still pass.

internal/framework/provider/functions/manifest_decode_multi_test.go
jrhouston: I actually didn't run into this panic.. and my test case tuple has a bunch of objects of different types in it 🤔

austinvalle: Oh! Coming back to this, I think the testing panic bug was really only for tuples with "primitives" as the elements. Looking back it's completely fine with tuples of objects. Apologies for the confusion!

internal/framework/provider/functions/encode.go
austinvalle: Gotcha, so that sounds like you'll want to handle that then.

There's also the possibility that Terraform will know the underlying type (like a string) but not know the underlying value (null or unknown). For null list elements I'd imagine this would be the most common case. I wrote up some initial docs on that here: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/dynamic-data#handling-underlying-null-and-unknown-values

Will probably need to expand on that more in future docs, but let me know if that doesn't make sense 😆

BBBmau (approved):

PR #2432, opened 2024-02-21
fix crash when unexpeceted annotation appears

### LINKED ISSUE #1591: Error: unexpected new value in metadata.annotations: default computed_fields should cover it

```
Terraform version: v1.1.4
Kubernetes provider version: v2.7.1
Kubernetes version: 1.21.5-gke.1302
```

### Affected Resource(s)
- `kubernetes_resource`

```hcl
resource "kubernetes_manifest" "foo" {
  manifest = {
    apiVersion = "apps/v1"
    kind       = "DaemonSet"
    ...
  }
}
```

### Steps to Reproduce
1. `terraform apply`
2. edit the spec of the manifest
3. `terraform apply`

### Expected Behavior
The terraform apply works

### Actual Behavior
The terraform apply fails with the following error:
```

Error: Provider produced inconsistent result after apply

When applying changes to kubernetes_manifest.foo, provider "provider[\"registry.terraform.io/hashicorp/kubernetes\"]" produced an unexpected new value: .object.metadata.annotations["deprecated.daemonset.template.generation"]: was cty.StringVal("1"), but now cty.StringVal("2"). This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

### Important Factoids
- I read about [`computed_fields`](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest#computed-fields). If I understand correctly, the default value includes the `metadata.annotations`, so it should cover my scenario, shouldn't it?
- A subsequent apply does work
- An explicit `computed_fields = ["metadata.annotations","metadata.labels"]` doesn't change anything

### References

- GH-1370 is the closest issue I could find. maybe it was before the `computed_fields` though? (and thus fixed by it?)

### Community Note

BBBmau:
### Description

The idea is to treat all values in `computed_field` as an `UnknownValue`, before the values would only be treated as an `UnknownValue` when a change was present. This will insure that values in annotations/labels are covered especially in cases where the user did not set the label/annotation themselves.

Fixes #1591

### Acceptance tests

```

...
```

### Release Note

```release-note
...
```

### Community Note

### DIFF
--- a/manifest/provider/apply.go
+++ b/manifest/provider/apply.go
@@ -193,7 +193,7 @@ func (s *RawProviderServer) ApplyResourceChange(ctx context.Context, req *tfprot
 				return v, nil
 			}
 			if v.IsKnown() {
-				return v, nil
+				return tftypes.NewValue(v.Type(), tftypes.UnknownValue), nil
 			}
 			ppMan, restPath, err := tftypes.WalkAttributePath(plannedStateVal["manifest"], ap)
 			if err != nil {
--- a/manifest/provider/plan.go
+++ b/manifest/provider/plan.go
@@ -422,6 +422,9 @@ func (s *RawProviderServer) PlanResourceChange(ctx context.Context, req *tfproto
 		}
 		updatedObj, err := tftypes.Transform(completePropMan, func(ap *tftypes.AttributePath, v tftypes.Value) (tftypes.Value, error) {
 			_, isComputed := computedFields[ap.String()]
+			if isComputed {
+				return tftypes.NewValue(v.Type(), tftypes.UnknownValue), nil
+			}
 			if v.IsKnown() { // this is a value from current configuration - include it in the plan
 				hasChanged := false
 				wasCfg, restPath, err := tftypes.WalkAttributePath(priorMan, ap)
@@ -430,20 +433,12 @@ func (s *RawProviderServer) PlanResourceChange(ctx context.Context, req *tfproto
 				}
 				nowCfg, restPath, err := tftypes.WalkAttributePath(ppMan, ap)
 				hasChanged = err == nil && len(restPath.Steps()) == 0 && wasCfg.(tftypes.Value).IsKnown() && !wasCfg.(tftypes.Value).Equal(nowCfg.(tftypes.Value))
+
 				if hasChanged {
 					h, ok := hints[morph.ValueToTypePath(ap).String()]
 					if ok && h == manifest.PreserveUnknownFieldsLabel {
 						apm := append(tftypes.NewAttributePath().WithAttributeName("manifest").Steps(), ap.Steps()...)
 						resp.RequiresReplace = append(resp.RequiresReplace, tftypes.NewAttributePathWithSteps(apm))
-					}
-				}
-				if isComputed {
-					if hasChanged {
-						return tftypes.NewValue(v.Type(), tftypes.UnknownValue), nil
-					}
-					nowVal, restPath, err := tftypes.WalkAttributePath(proposedVal["object"], ap)
-					if err == nil && len(restPath.Steps()) == 0 {
-						return nowVal.(tftypes.Value), nil
 					}
 				}
 				return v, nil


BBBmau:
latest commit solves the issue from referenced issue, however we still run into this error if the user chooses to set computed_fields and not include `annotations/labels` in the list of paths. However this does still follow the intended design from the docs:
```
IMPORTANT: By default, metadata.labels and metadata.annotations are already included in the list. You don't have to set them explicitly in the computed_fields list. To turn off these defaults, set the value of computed_fields to an empty list or a concrete list of other fields. For example computed_fields = [].
```

If the solution to this looks good then i think the only thing left is to add a test for this case to check that annotations/labels are seen as `UnknownValues` always

BBBmau:
I've attempted to run the test on my machine but running into issues, could be because im not using kind and instead using minikube:

```
└─(17:43:54 on computed_fields_allUnknown ✹)──> make testacc TESTARGS="-run TestKubernetesManifest_ComputedFields"                                   2 ↵ ──(Fri,Mar01)─┘
go test -count=1 -tags acceptance "./test/acceptance" -v -run TestKubernetesManifest_ComputedFields -timeout 120m
2024/03/01 17:44:13 Testing against Kubernetes API version: v1.29.0-eks-c417bb3
=== RUN   TestKubernetesManifest_ComputedFields
2024-03-01T17:45:04.939-0800 [ERROR] [ApplyResourceChange][Apply]: API error="Internal error occurred: failed calling webhook \"tf-acc-test-wbezqriecd.hashicorp.com\": failed to call webhook: Post \"https://tf-acc-test-wbezqriecd.tf-acc-test-cnbarqjfjb.svc:443/mutate?timeout=10s\": no endpoints available for service \"tf-acc-test-wbezqriecd\"" API response=<nil>
    computed_attr_test.go:78: Error when trying to get resource tf-acc-test-cnbarqjfjb/tf-acc-test-wbezqriecd: configmaps "tf-acc-test-wbezqriecd" not found
--- FAIL: TestKubernetesManifest_ComputedFields (109.39s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x103e41168]

goroutine 70 [running]:
testing.tRunner.func1.2({0x104420dc0, 0x1056a9360})
        /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:1631 +0x1c4
testing.tRunner.func1()
        /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:1634 +0x33c
panic({0x104420dc0?, 0x1056a9360?})
        /opt/homebrew/Cellar/go/1.22.0/libexec/src/runtime/panic.go:770 +0x124
github.com/hashicorp/terraform-provider-kubernetes/manifest/test/helper/state.getAttributesValuesFromResource(0x103ef09c3?, {0x140012d1e78, 0x18})
        /Users/mau/Dev/terraform-provider-kubernetes/manifest/test/helper/state/state_helper.go:37 +0x28
github.com/hashicorp/terraform-provider-kubernetes/manifest/test/helper/state.(*Helper).GetAttributeValue(0x1400089bac8, 0x140004309c0, {0x103ef09c3, 0x3f})
        /Users/mau/Dev/terraform-provider-kubernetes/manifest/test/helper/state/state_helper.go:113 +0x60
github.com/hashicorp/terraform-provider-kubernetes/manifest/test/helper/state.(*Helper).AssertAttributeValues(0x140018cdac8, 0x140004309c0, 0x1400089ba98)
        /Users/mau/Dev/terraform-provider-kubernetes/manifest/test/helper/state/state_helper.go:147 +0x98
github.com/hashicorp/terraform-provider-kubernetes/manifest/test/acceptance.TestKubernetesManifest_ComputedFields(0x140004309c0)
        /Users/mau/Dev/terraform-provider-kubernetes/manifest/test/acceptance/computed_attr_test.go:85 +0xb34
testing.tRunner(0x140004309c0, 0x104701c68)
        /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:1689 +0xec
created by testing.(*T).Run in goroutine 1
        /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:1742 +0x318
FAIL    github.com/hashicorp/terraform-provider-kubernetes/manifest/test/acceptance     112.438s
FAIL
make: *** [testacc] Error 1
┌─(~/Dev/terraform-provider-kubernetes/manifest)───────────────────────────────────────────────────────────────────────────────────────
┌─(~/Dev/terraform-provider-kubernetes/manifest)──────────────────────────────────────────────────────────────────────────────────────────────────────────────────(mau@mau-JKDT676NCP:s023)─┐
└─(17:46:52 on computed_fields_allUnknown ✹)──> git commit -m "cleanup plan code and add computedFields test when setting computedFields"
[computed_fields_allUnknown a86a18182] cleanup plan code and add computedFields test when setting computedFields
 4 files changed, 62 insertions(+), 9 deletions(-)
 ```

File: manifest/provider/apply.go [hunk 1]
```diff
@@ -192,9 +192,7 @@ func (s *RawProviderServer) ApplyResourceChange(ctx context.Context, req *tfprot
 			if !isComputed {
 				return v, nil
 			}
-			if v.IsKnown() {
```
BBBmau, line 195: This could be causing all the errors to show up, the reason for removing this is because all computedfields are now set to unknown, a value will never be known making this check pointless.

BBBmau: This comment can be ignored. Issue is due to the new test that i added:

```
=== RUN   TestKubernetesManifest_ComputedFields
    computed_attr_test.go:85: Could not find resource "kubernetes_manifest.deployment_resource_diff" in state
--- FAIL: TestKubernetesManifest_ComputedFields (25.12s)
```

alexsomesan: Values in the manifest could be unknown even without being computed. For example, when being set from attributes of upstream resources that are only known after applying that resource. 

But regardless of that, this statement is where we skip over all the known values. Removing this line would prevent known values from being left in the manifest after the transform.

alexsomesan: Never mind, I think we're good removing this since the known values and non-computed unknown will be handled by the `if` above. Good catch!

BBBmau: Passing Test has been added that checks when a Deploment value is changed by the API. Ready to Merge now! 🚀

manifest/test/acceptance/testdata/ComputedFields/computed_deployment.tf
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+resource "kubernetes_manifest" "deployment_resource_diff" {
+computed_fields = ["spec.template.spec.containers[0].resources.limits"]
```
alexsomesan, line 5: Let's test specifically for the case that was causing the initial issue (keys in a map). 
```suggestion
computed_fields = ["spec.template.spec.containers[0].resources.limits[\"cpu\"]"]
```

manifest/provider/apply.go
```diff
@@ -192,9 +192,7 @@ func (s *RawProviderServer) ApplyResourceChange(ctx context.Context, req *tfprot
...
-				return v, nil
-			}
+
 			ppMan, restPath, err := tftypes.WalkAttributePath(plannedStateVal["manifest"], ap)
 			if err != nil {
 				if len(restPath.Steps()) > 0 {
```
alexsomesan, line 198: No need for this `if` anymore (it's addressed as part of `findBackfillValue()`)

manifest/provider/apply.go
```diff
@@ -192,9 +192,7 @@ func (s *RawProviderServer) ApplyResourceChange(ctx context.Context, req *tfprot
 			if !isComputed {
 				return v, nil
 			}
-			if v.IsKnown() {
-				return v, nil
-			}
+
 			ppMan, restPath, err := tftypes.WalkAttributePath(plannedStateVal["manifest"], ap)
```
alexsomesan, line 196:
```suggestion
			ppMan, restPath, err := findBackfillValue(plannedStateVal["manifest"], ap)
```

alexsomesan (changes requested):
We still need logic to look up the original value of the computed attribute in "manifest" with the appropriate attribute path adaptations, to avoid the "attribute disappeared" errors.

We also need to revert back to only setting values to computed during planning if they actually changed in configuration. Otherwise, values modified by the API during apply (such as "cpu" in the test example) will trigger a perpetual diff.


Here's an example function for looking up the values in manifest, that I tested and found to work decently:

```

// findBackfillValue tries to optimistically find an attribute value pointed to by an attribute path (ap) inside 
// a complex type container value (m) by making semantically equivalent adaptations to the attribute path steps.
// It considers Object and Map types as semantically equivalent AFA indexing attributes goes.
func findBackfillValue(m interface{}, ap *tftypes.AttributePath) (interface{}, *tftypes.AttributePath, error) {
	v, restPath, err := tftypes.WalkAttributePath(m, ap)
	if err != nil {
		if len(restPath.Steps()) > 0 {
			// Attribute might not be found, because the attribute path step type  doesn't matcht 
			// the container type being indexed (core parses HCL to only Object and Tupple, but not Map and List).
			// In that case, the attribute paths constructed for values in "object" 
			// will need adjusting for type differences between "manifest"
			fs := restPath.Steps()[0]
			if e, ok := fs.(tftypes.ElementKeyString); ok {
				// if expecting a Map, try indexing with AttributeName instead
				tp := tftypes.NewAttributePath().WithAttributeName(string(e))
				tv, rp, err := tftypes.WalkAttributePath(v, tp)
				if err != nil {
					return v, rp, err
				}
				return findBackfillValue(tv, tftypes.NewAttributePathWithSteps(restPath.Steps()[1:]))
			}
			if e, ok := fs.(tftypes.AttributeName); ok {
				// if expecting an Object, try indexing with ElementKeyString instead
				tp := tftypes.NewAttributePath().WithElementKeyString(string(e))
				tv, rp, err := tftypes.WalkAttributePath(v, tp)
				if err != nil {
					return v, rp, err
				}
				return findBackfillValue(tv, tftypes.NewAttributePathWithSteps(restPath.Steps()[1:]))
			}
		}
		return nil, nil, err
	}
	return v, restPath, err
}
```

BBBmau:
Latest commit addresses the case where API changes the config value set by the user.

How it appears in config:
```
...
resources = {
     limits = {
       cpu    = "0.25"
       memory = "512Mi"
}
```

How it appears in YAML
```
...
    Limits:
      cpu:     250m
      memory:  512Mi
```

BBBmau:
I played around more with this and found this config from the original issue to cause a change despite no change actually happening. The original panic doesn't happen however.
TFCONFIG:
```hcl
resource "kubernetes_manifest" "daemonset_prometheus_daemonset" {
  manifest = {
    apiVersion = "apps/v1"
    kind       = "DaemonSet"
    metadata = {
      name      = "prometheus-daemonset"
      namespace = "default"
    }
    spec = {
      selector = {
        matchLabels = {
          name = "prometheus-exporter"
          tier = "monitoring"
        }
      }
      template = {
        metadata = {
          labels = {
            name = "prometheus-exporter"
            tier = "monitoring"
          }
        }
        spec = {
          containers = [
            {
              image = "prom/node-exporter"
              name  = "prometheus"
              ports = [
                {
                  containerPort = 80
                },
              ]
            },
          ]
        }
      }
    }
  }
}
```
```hcl
  # kubernetes_manifest.daemonset_prometheus_daemonset will be updated in-place
  ~ resource "kubernetes_manifest" "daemonset_prometheus_daemonset" {
      ~ object   = {
          ~ metadata   = {
              ~ annotations                = {
                  - "deprecated.daemonset.template.generation" = "1"
                } -> (known after apply)
              + labels                     = (known after apply)
                name                       = "prometheus-daemonset"
                # (12 unchanged attributes hidden)
            }
            # (3 unchanged attributes hidden)
        }
    }

```

Changes to the `findBackfillValue` could resolve this.

gnuletik:
Hi ! Is there anything blocking this PR?
Thanks!

jwhitaker-gridcog:
pinging @alexsomesan - this PR fixes a dealbreaker with using the terraform k8s provider for anything serious, is there anything blocking its merging?

Issue #1829, opened 2022-09-07
kubernetes_manifest imported resources are recreated not updated (reopening)


zonybob:
### Terraform Version, Provider Version and Kubernetes Version

Effectively duplicate of #1712 but on v2.12.1

```
Terraform version: 1.1.1 (also tried 1.2.8)
Kubernetes provider version: 2.12.1
Kubernetes version: 1.21
```

### Affected Resource(s)

- kubernetes_manifest

### Terraform Configuration Files
```hcl

resource "kubernetes_manifest" "worker_machine_set" {
  for_each = var.az_subnet_map
  manifest = yamldecode(templatefile("${path.module}/manifests/machineset/worker-machine-set.yaml", {
    az            = "${var.region}${each.key}",
    region        = var.region,
    environment   = var.environment,
    subnet_id     = each.value,
    ami_id        = var.worker_ami_id,
    instance_type = var.worker_instance_type
  }))

  field_manager {
    force_conflicts = true
  }

  lifecycle {
    prevent_destroy = true
  }
}
```

worker-machine-set.yaml
```yaml
apiVersion: machine.openshift.io/v1beta1
kind: MachineSet
metadata:
  name: ${environment}-worker-${az}
  namespace: openshift-machine-api
  labels:
    machine.openshift.io/cluster-api-cluster: ${environment}
spec:
  selector:
    matchLabels:
      machine.openshift.io/cluster-api-cluster: ${environment}
      machine.openshift.io/cluster-api-machineset: ${environment}-worker-${az}
  template:
    metadata:
      labels:
        machine.openshift.io/cluster-api-cluster: ${environment}
        machine.openshift.io/cluster-api-machine-role: worker
        machine.openshift.io/cluster-api-machine-type: worker
        machine.openshift.io/cluster-api-machineset: ${environment}-worker-${az}
    spec:
      metadata:
        labels:
          kubic.nga.mil/deployment: worker
      providerSpec:
        value:
          userDataSecret:
            name: worker-user-data
          placement:
            availabilityZone: ${az}
            region: ${region}
          credentialsSecret:
            name: aws-cloud-credentials
          instanceType: ${instance_type}
          blockDevices:
            - ebs:
                encrypted: true
                iops: 2000
                kmsKey:
                  arn: ""
                volumeSize: 500
                volumeType: io1
          securityGroups:
            - filters:
                - name: "tag:Name"
                  values:
                    - ${environment}-worker-sg*
[...]

PR #2437, opened 2024-03-04
manifest: remove forced replacement with `x-kubernetes-preserve-unknown-fields`

BBBmau:
### Description

This check will prevent `x-kubernetes-preserve-unknown-fields` to cause a replacement whenever the value is changed. The idea is that it should ONLY cause a replacement if the type itself is changed and not the value. 

Fixes https://github.com/hashicorp/terraform-provider-kubernetes/issues/2371
Fixes https://github.com/hashicorp/terraform-provider-kubernetes/issues/2375
Fixes https://github.com/hashicorp/terraform-provider-kubernetes/issues/1928
Fixes https://github.com/hashicorp/terraform-provider-kubernetes/issues/2410

### Acceptance tests

```
┌─(~/Dev/terraform-provider-kubernetes/manifest)───────────────────────────────────────────────(mau@mau-JKDT676NCP:s077)─┐
└─(14:48:09 on fix-unknownFieldsLogic-manifest ✹ ✭)──> make testacc TESTARGS="-run TestKubernetesManifest_CustomResource_x_preserve_unknown_fields"
go test -count=1 -tags acceptance "./test/acceptance" -v -run TestKubernetesManifest_CustomResource_x_preserve_unknown_fields -timeout 120m
2024/03/12 14:48:28 Testing against Kubernetes API version: v1.27.3
=== RUN   TestKubernetesManifest_CustomResource_x_preserve_unknown_fields
--- PASS: TestKubernetesManifest_CustomResource_x_preserve_unknown_fields (14.23s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/manifest/test/acceptance     14.832s```
```
### Release Note

```release-note
`kubernetes_manifest`: add TypeCheck for `x-kubernetes-preserve-unknown-fields` to prevent unnecessary replacement
```

### Community Note

### DIFF
--- a/manifest/provider/plan.go
+++ b/manifest/provider/plan.go
@@ -458,7 +458,8 @@ func (s *RawProviderServer) PlanResourceChange(ctx context.Context, req *tfproto
 				hasChanged = err == nil && len(restPath.Steps()) == 0 && wasCfg.(tftypes.Value).IsKnown() && !wasCfg.(tftypes.Value).Equal(nowCfg.(tftypes.Value))
 				if hasChanged {
 					h, ok := hints[morph.ValueToTypePath(ap).String()]
-					if ok && h == manifest.PreserveUnknownFieldsLabel {
+					typeChanged := !(wasCfg.(tftypes.Value).Type().Equal(nowCfg.(tftypes.Value).Type()))
+					if ok && h == manifest.PreserveUnknownFieldsLabel && typeChanged {
 						apm := append(tftypes.NewAttributePath().WithAttributeName("manifest").Steps(), ap.Steps()...)
 						resp.RequiresReplace = append(resp.RequiresReplace, tftypes.NewAttributePathWithSteps(apm))
 					}
--- a/manifest/provider/plan.go
+++ b/manifest/provider/plan.go
@@ -458,8 +458,8 @@ func (s *RawProviderServer) PlanResourceChange(ctx context.Context, req *tfproto
 				hasChanged = err == nil && len(restPath.Steps()) == 0 && wasCfg.(tftypes.Value).IsKnown() && !wasCfg.(tftypes.Value).Equal(nowCfg.(tftypes.Value))
 				if hasChanged {
 					h, ok := hints[morph.ValueToTypePath(ap).String()]
-					typeChanged := !(wasCfg.(tftypes.Value).Type().Equal(nowCfg.(tftypes.Value).Type()))
-					if ok && h == manifest.PreserveUnknownFieldsLabel && typeChanged {
+					_, diff := wasCfg.(tftypes.Value).Diff(nowCfg.(tftypes.Value)) // passing values of two different types will result in an error.
+					if ok && h == manifest.PreserveUnknownFieldsLabel && diff != nil {
 						apm := append(tftypes.NewAttributePath().WithAttributeName("manifest").Steps(), ap.Steps()...)
 						resp.RequiresReplace = append(resp.RequiresReplace, tftypes.NewAttributePathWithSteps(apm))
 						resp.Diagnostics = append(resp.Diagnostics, &tfprotov5.Diagnostic{
--- a/manifest/provider/plan.go
+++ b/manifest/provider/plan.go
@@ -462,6 +462,11 @@ func (s *RawProviderServer) PlanResourceChange(ctx context.Context, req *tfproto
 					if ok && h == manifest.PreserveUnknownFieldsLabel && typeChanged {
 						apm := append(tftypes.NewAttributePath().WithAttributeName("manifest").Steps(), ap.Steps()...)
 						resp.RequiresReplace = append(resp.RequiresReplace, tftypes.NewAttributePathWithSteps(apm))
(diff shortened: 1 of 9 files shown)


BBBmau:
@alexsomesan 

I'll need some feedback since when I run it through the debug. the plan produces a plannedchange of just the value and not the entire resource. However if we were to run it without debug mode it still produces the same replacement action, which goes against the logic that's been applied in this PR.

After resolving that we'll look into adding a test for this. 💯

manifest/provider/plan.go
alexsomesan, line 1: We need to remove lines 347-350. It enforces a replace as soon as the type of `object` isn't fully defined, regardless if it changed or not.

```
		resp.RequiresReplace = []*tftypes.AttributePath{
			tftypes.NewAttributePath().WithAttributeName("manifest"),
			tftypes.NewAttributePath().WithAttributeName("object"),
		}
```

We also have to remove the last part of the warning at line 324, but keep the warning about missing schema.

>  Updates to this resource will cause a forced replacement.

Instead, this should be issued as a warning whenever we detect a type change.

BBBmau: Looks like we're missing some info, when running through debug this block of code is until hit when the object_type of either `object` or `manifest` is changed. This wouldn't be touching anything else such as `valuesObject` that's found in referenced issues.

I went ahead and added the warning message in the same area where I added the type check. I tested it myself and it looks to be working now. Only does a replacement if the type changes, does a normal update if value changes but type is still the same. The wording could be changed, let me know what you think.

```
Plan: 1 to add, 0 to change, 1 to destroy.
╷
│ Warning: This value's type has changed
│ 
│   with kubernetes_manifest.raw,
│   on main.tf line 3, in resource "kubernetes_manifest" "raw":
│    3: resource "kubernetes_manifest" "raw" {
│ 
│ Changes to the type will cause a forced replacement.
```

BBBmau: improved warning
```
Plan: 1 to add, 0 to change, 1 to destroy.
╷
│ Warning: The attribute path AttributeName("spec").AttributeName("sources").ElementKeyInt(-1).AttributeName("helm").AttributeName("valuesObject") value's type has changed
│ 
│   with kubernetes_manifest.raw,
│   on main.tf line 3, in resource "kubernetes_manifest" "raw":
│    3: resource "kubernetes_manifest" "raw" {
│ 
│ Changes to the type will cause a forced replacement.
```

BBBmau: Added tests, Ready to be merged!

BBBmau:
Adding a note here that when adding a value into the tuple it will cause a recreation despite the type being the same:
```
          valuesObject = {
            "mau": ["test"] -> ["test", "test2"]
            }
```
```
Plan: 1 to add, 0 to change, 1 to destroy.
╷
│ Warning: The attribute path AttributeName("spec").AttributeName("sources").ElementKeyInt(-1).AttributeName("helm").AttributeName("valuesObject") value's type has changed
│ 
│   with kubernetes_manifest.raw,
│   on main.tf line 3, in resource "kubernetes_manifest" "raw":
│    3: resource "kubernetes_manifest" "raw" {
│ 
│ Changes to the type will cause a forced replacement.
```

BBBmau:
> Adding a note here that when adding a value into the tuple it will cause a recreation despite the type being the same:
> 
> ```
>           valuesObject = {
>             "mau": ["test"] -> ["test", "test2"]
>             }
> ```
> 
> ```
> Plan: 1 to add, 0 to change, 1 to destroy.
> ╷
> │ Warning: The attribute path AttributeName("spec").AttributeName("sources").ElementKeyInt(-1).AttributeName("helm").AttributeName("valuesObject") value's type has changed
> │ 
> │   with kubernetes_manifest.raw,
> │   on main.tf line 3, in resource "kubernetes_manifest" "raw":
> │    3: resource "kubernetes_manifest" "raw" {
> │ 
> │ Changes to the type will cause a forced replacement.
> ```
> 
> An extra check is needed before we can get this merged.

Recent commit addresses this check by using `Diff` method from `tftypes.Value` which returns an Error if a change in type occurs on values.

`_, diff := wasCfg.(tftypes.Value).Diff(nowCfg.(tftypes.Value)) // passing values of two different types will result in an error.`

Prior we would check by using the `Equal` method:
`typeChanged := !(wasCfg.(tftypes.Value).Type().Equal(nowCfg.(tftypes.Value).Type()))` this didn't work when adding to the tuple despite the type still staying the same. Could be an issue with the `Equal` method type. From the docs: `Equal performs deep type equality checks, including attribute/element types and whether attributes are optional or not.`

BBBmau:
In the past, type changes was something that terraform panic to. This was early in the development of terraform core and appears it is not an issue anymore. Because of this we can remove the forced replacement for `x-kubernetes-preserve-unknown-fields` but will add a warning when users are intending to change types.

hongshaoyang:
This PR fixes #1829

TessaIO:
Any update on this PR?

FarhadF:
@BBBmau @alexsomesan whats blocking this PR?

alexsomesan (approved):
Looking good. 
I added an extra test case, just to be on the safe side.

Issue #2367, opened 2023-12-14
Error: Provider produced inconsistent result after apply


mailbox171:
### Terraform Version, Provider Version and Kubernetes Version

```
Terraform version: 1.6.1
Kubernetes provider version:  v2.24.0 (signed by HashiCorp)
Kubernetes version: 1.27.7
```

### Affected Resource(s)

argocd Application resource

### Terraform Configuration Files
```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### Steps to Reproduce

we run PLAN and then APPLY, this is out main.tf
```
resource "kubernetes_manifest" "buildbuddy" {
  manifest = yamldecode(templatefile(var.argocd_application_file, {
    "app_path" = var.app_path
  }))
}
```

### Expected Behavior
What should have happened?

We did not change the resource manifest.
APPLY should have returned 0

### Actual Behavior

PLAN
-------
```
Terraform will perform the following actions:

  # kubernetes_manifest.buildbuddy will be updated in-place
  ~ resource "kubernetes_manifest" "buildbuddy" {
      - computed_fields = [
          - "spec.source.targetRevision",
        ] -> null
        # (2 unchanged attributes hidden)

      - field_manager {
          - force_conflicts = true -> null
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

APPLY
--------
The apply seems to have being done.
But the output is the following

```
Run terraform apply -input=false 6843e10cf3135688ae16e0d49d32bbc0b7242927.tfplan
  
kubernetes_manifest.buildbuddy: Modifying...
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to kubernetes_manifest.buildbuddy, provider
│ "provider[\"registry.terraform.io/hashicorp/kubernetes\"]" produced an
│ unexpected new value: .object.operation.initiatedBy.automated: was null,
│ but now cty.True.
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to kubernetes_manifest.buildbuddy, provider
│ "provider[\"registry.terraform.io/hashicorp/kubernetes\"]" produced an
│ unexpected new value: .object.operation.retry.limit: was null, but now
[...]

PR #2489, opened 2024-05-14
Remove ArgoCD operation field to prevent perpetual diff

corylevine:
### Description

When using the `kubernetes_manifest` resource to manage an ArgoCD Application, it is possible for the status of Argo's sync operation to show up in the manifest when applying, but not during planning. This shows up under the top-level `operation` field, which is only present when an Argo sync operation is underway, potentially triggered by the Terraform apply.

This commit simply drops the `operation` field from manifests returned by the K8s API and thusly prevents it from being considered when diffing.

### Acceptance tests

```

...
```

### Release Note

```release-note
Prevent `operation` field in manifests from causing plan/apply diffs
```

### References

Issue #2367
### Community Note

### DIFF
--- a/manifest/provider/resource.go
+++ b/manifest/provider/resource.go
@@ -184,6 +184,9 @@ func RemoveServerSideFields(in map[string]interface{}) map[string]interface{} {
 	// Remove "status" attribute
 	delete(in, "status")
 
+	// Remove ArgoCD "operation" attribute
+	delete(in, "operation")
+
 	meta := in["metadata"].(map[string]interface{})
 
 	// Remove "uid", "creationTimestamp", "resourceVersion" as


jastr945 (approved):

alexsomesan:
Whatever is causing the issue with the `operation` attribute of ArgoCD's Application CR will likely cause similar issues in similar conditions on other resources. We need to understand the root cause of it and address it generically.

This approach is not the way forward for this issue, because it introduces special handling of a specific CR (not even a core K8s resource). We designed the `kubernetes_manifest` resource with the express intent of handling all K8s object types through the same codepath and logic. We strive to keep it along that path and address issues in a generic manner.

Closing this, let's go back to the conversation in #2367 and find the root cause.

PR #2509, opened 2024-05-30
kubernetes_manifest flag to skip validation during plan

DanielHaba:
### Description
The common walkaround for missing CRDs is to use other providers like `kubectl_manifest` or `helm_release`. The deferred flag on `kubernetes_manifest` will be more elegant solution, with the same downsides.

### Acceptance tests

```

...
```

### Release Note

```release-note
...
```

### Community Note

### DIFF
--- a/manifest/provider/plan.go
+++ b/manifest/provider/plan.go
@@ -9,16 +9,21 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
-	"github.com/hashicorp/terraform-provider-kubernetes/manifest"
-	"github.com/hashicorp/terraform-provider-kubernetes/manifest/morph"
-	"github.com/hashicorp/terraform-provider-kubernetes/manifest/payload"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+
+	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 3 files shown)


alexsomesan:
Hi Daniel,

I hate to be a spoiler here, but this approach is not going to work.
The `kubernetes_manifest` resource doesn't just do validation during the plan, it actually retrieves the schema for the resource from the OpenAPI definition available on the cluster API. Even if you could get away with what this PR proposes, there still is a need for schema information. There is no way to work around that, as Terraform requires a uniform and consistent resource structure throughout the lifecycle of the resource.

Just wanted to let you know that we can't merge this kind of change. I do agree the internals of the `kubernetes_manifest` should be better documented from a developer POV in order to avoid wasting effort like this.

BBBmau:
> Hi Daniel,
> 
> I hate to be a spoiler here, but this approach is not going to work. The `kubernetes_manifest` resource doesn't just do validation during the plan, it actually retrieves the schema for the resource from the OpenAPI definition available on the cluster API. Even if you could get away with what this PR proposes, there still is a need for schema information. There is no way to work around that, as Terraform requires a uniform and consistent resource structure throughout the lifecycle of the resource.
> 
> Just wanted to let you know that we can't merge this kind of change. I do agree the internals of the `kubernetes_manifest` should be better documented from a developer POV in order to avoid wasting effort like this.

Closing this PR with the reasoning mentioned from a previous comment.

PR #2539, opened 2024-07-03
add namespace to sa error message

tomkukral:
ServiceAccount name may not be unique across namespaces, so it may be hard to understand which ServiceAccount is being referred.


arybolovlev (approved):
Nice addition, @tomkukral, thank you!

tomkukral:
Thank you for very fast merging :+1:

PR #2596, opened 2024-10-01
Handle Jobs with ttl_seconds_after_finished = 0 correctly

### LINKED ISSUE #2531: Terraform lifecycle of a job that has ttl_seconds_after_finished set
This is basically a copy of issue #2059, which unfortunately went stale due to a lack of activity. The issue is still relevant and affecting me.

```
Terraform version: v1.5.7
Kubernetes Provider version: v2.23.0
Kubernetes version: v1.28.9
```
## Terraform configuration
```hcl
terraform {
  required_providers {
    kubernetes = {
      source  = "hashicorp/kubernetes"
      version = ">= 2.18.0"
    }
  }

  required_version = ">= 1.4.0"
}

provider "kubernetes" {
  config_path = "~/.kube/config"
}


variable "test" {
  type    = number
  default = 1
}

resource "terraform_data" "kubernetes_job_control" {
  input = var.test
}

resource "kubernetes_job" "mock_job" {
  lifecycle {
    replace_triggered_by = [
      terraform_data.kubernetes_job_control
    ]
  }
  metadata {
    name = "mock-job"
  }
  spec {
    ttl_seconds_after_finished = 0
    template {
      metadata {}
      spec {
        container {
          name  = "mock-job"
          image = "hello-world"
        }
        restart_policy = "Never"
      }
    }
    backoff_limit = 0
  }
  wait_for_completion = true
}
```

## Question
Hello,

I have a `kubernetes_job` that sets up some initial configuration in my db when the database is created, for that I did `lifecycle { replace_triggered_by = [ my_database_resource.id ] }`. The thing is, I don't want this job to stick around in the cluster so I've set `ttl_seconds_after_finished=0`. I thought that by setting the lifecycle block the job would only be applied whenever I need to recreate the database, but because the job deletes itself from cluster it is recreated in every `terraform apply` that I do. It looks like a bug to me but I think it would be better to ask for opinions here first.

The code I pasted here is a simplified terraform project to reproduce this problem in case someone wants to try it locally. The principle is the same, I have a variable controlling the lifecycle and the job's `ttl_seconds_after_finished` set to 0.

JaylonmcShan03:
### Description
Fixes #2531 
When a Kubernetes Job has ttl_seconds_after_finished set to 0, Kubernetes deletes the Job immediately after it completes. This can cause Terraform to plan the recreation of the Job in subsequent runs, which is sort of undesirable behavior. The expected behavior is for Terraform to recognize that the Job was deleted intentionally and not attempt to recreate it or show any diffs.

**Changes made**
    - Read function:
        - When the read crud func encounters a `NotFound` error for a job, it now checks if `ttl `is set to 0, if so it keeps the 
        resource in the state without marking it as destroyed, this helps prevent the recreation of the job. If it is not 0, it 
        behaves as before by removing the resource from the state.

CustomizeDiff function introduced:
        - Supressing diffs when the job has been deleted due to due to ttl = 0

### Acceptance tests

```
└─(10:13:50 on fix-job-ttl-zero-handling)──> make test TESTARGS="-run TestAccKubernetesJobV1_customizeDiff_ttlZero"
go vet ./...
go test "/Users/mau/Dev/terraform-provider-kubernetes/kubernetes" -vet=off -run TestAccKubernetesJobV1_customizeDiff_ttlZero -parallel 8 -timeout=30s
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes   0.924s
...
```

### Release Note

```release-note
Properly handle Kubernetes Jobs with ttl_seconds_after_finished = 0 to prevent unnecessary recreation.
```

### References
Fixes #2531 

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_job_v1.go
+++ b/kubernetes/resource_kubernetes_job_v1.go
@@ -93,30 +93,28 @@ func resourceKubernetesJobV1CustomizeDiff(ctx context.Context, d *schema.Resourc
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			// Job is missing
-			if oldTTLInt == 0 {
-				if newTTLInt == 0 {
-					// TTL remains 0; suppress diff to prevent recreation
-					log.Printf("[DEBUG] Job %s not found and ttl_seconds_after_finished remains 0; suppressing diff", d.Id())
+			if oldTTLInt >= 0 {
+				if oldTTLInt != newTTLInt {
+					// TTL value changed; force recreation
+					log.Printf("[DEBUG] Job %s not found and ttl_seconds_after_finished changed from %d to %d; forcing recreation", d.Id(), oldTTLInt, newTTLInt)
+					d.ForceNew("spec.0.ttl_seconds_after_finished")
+					return nil
+				} else {
+					// TTL remains the same; suppress diff
+					log.Printf("[DEBUG] Job %s not found and ttl_seconds_after_finished remains %d; suppressing diff", d.Id(), oldTTLInt)
 					d.Clear("spec")
 					d.Clear("metadata")
 					return nil
-				} else {
-					// TTL changed from 0 to non-zero; force recreation
-					log.Printf("[DEBUG] Job %s not found and ttl_seconds_after_finished changed from 0 to %d; forcing recreation", d.Id(), newTTLInt)
-					d.ForceNew("spec.0.ttl_seconds_after_finished")
-					return nil
 				}
-			} else {
-				return nil
 			}
 		} else {
 			return err
 		}
 	} else {
-		// Job exists
-		if oldTTLInt == 0 && newTTLInt != 0 {
-			// TTL changing from 0 to non-zero; force recreation
-			log.Printf("[DEBUG] Job %s exists and ttl_seconds_after_finished changed from 0 to %d; forcing recreation", d.Id(), newTTLInt)
+		// Job exists, check if TTL changed
+		if oldTTLInt != newTTLInt {
+			// TTL changed; force recreation
+			log.Printf("[DEBUG] Job %s exists and ttl_seconds_after_finished changed from %d to %d; forcing recreation", d.Id(), oldTTLInt, newTTLInt)
 			d.ForceNew("spec.0.ttl_seconds_after_finished")
 			return nil
 		}
@@ -263,21 +261,22 @@ func resourceKubernetesJobV1Update(ctx context.Context, d *schema.ResourceData, 
 	_, err = conn.BatchV1().Jobs(namespace).Get(ctx, name, metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			// Job is missing; cannot update
+			// Job is missing; check TTL
 			ttlAttr := d.Get("spec.0.ttl_seconds_after_finished")
 			ttlStr, _ := ttlAttr.(string)
 			ttlInt, err := strconv.Atoi(ttlStr)
 			if err != nil {
 				ttlInt = 0
 			}
-			if ttlInt == 0 {
-				// Job was deleted due to TTL = 0; nothing to update
-				log.Printf("[INFO] Job %s not found but ttl_seconds_after_finished = 0; nothing to update", d.Id())
+
+			if ttlInt >= 0 {
+				// Job was deleted due to TTL nothing to update
+				log.Printf("[INFO] Job %s not found but ttl_seconds_after_finished = %v; nothing to update", d.Id(), ttlInt)
 				return nil
-			} else {
-				// Job was deleted unexpectedly; return an error
-				return diag.Errorf("Job %s not found; cannot update because it has been deleted", d.Id())
 			}
+
+			// Job was deleted unexpectedly; return an error
+			return diag.Errorf("Job %s not found; cannot update because it has been deleted", d.Id())
 		}
 		return diag.Errorf("Error retrieving Job: %s", err)
 	}
--- a/kubernetes/resource_kubernetes_job_v1.go
+++ b/kubernetes/resource_kubernetes_job_v1.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"strconv"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -17,7 +16,6 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	pkgApi "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -30,7 +28,6 @@ func resourceKubernetesJobV1() *schema.Resource {
 		ReadContext:   resourceKubernetesJobV1Read,
 		UpdateContext: resourceKubernetesJobV1Update,
 		DeleteContext: resourceKubernetesJobV1Delete,
-		CustomizeDiff: resourceKubernetesJobV1CustomizeDiff,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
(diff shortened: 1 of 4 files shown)


kubernetes/resource_kubernetes_job_v1_test.go
```diff
@@ -516,3 +550,28 @@ func testAccKubernetesJobV1Config_modified(name, imageName string) string {
   wait_for_completion = false
 }`, name, imageName)
 }
+
+func testAccKubernetesJobV1Config_customizeDiff_ttlZero(name, imageName string) string {
```
BBBmau, line 554: When applying the `tfconfig` manually it works as expected. The apply goes through with no diff occurring. 

I did however attempt to update `ttl_seconds_after_finished` from 0 to 5 and got the following error:

```hcl
(base) ┌─(~/Dev/Scratch/ttl_test)────────────────────────────────────(mau@mau-JKDT676NCP:s017)─┐
└─(10:16:32)──> tfa                                                       ──(Thu,Oct10)─┘
kubernetes_job_v1.test: Refreshing state... [id=default/ttl-test]

Terraform used the selected providers to generate the following execution plan. Resource
actions are indicated with the following symbols:
  ~ update in-place

  # kubernetes_job_v1.test will be updated in-place
  ~ resource "kubernetes_job_v1" "test" {
        id                  = "default/ttl-test"

      ~ spec {
          ~ ttl_seconds_after_finished = "0" -> "5"
            # (8 unchanged attributes hidden)

            # (2 unchanged blocks hidden)
        }

        # (1 unchanged block hidden)
    }

kubernetes_job_v1.test: Modifying... [id=default/ttl-test]
╷
│ Error: Failed to update Job! API error: jobs.batch "ttl-test" not found
│ 
│   with kubernetes_job_v1.test,
│   on main.tf line 1, in resource "kubernetes_job_v1" "test":
│    1: resource "kubernetes_job_v1" "test" {
```

From my understanding we're wanting to prevent this error from happening when set to `0` which has been achieved. But in doing so we are now unable to update the existing job that's in tfstate assuming we want to update the already existing job that's part of state despite it having a `ttl` of `0`.

We'll want to consider how to solve this since if this is left users will need to destroy every job that has `ttl_seconds_after_finished = 0` if wanting to apply an update to the job already existing in state.

JaylonmcShan03: Correct, initially the issue was when ttl_seconds_after_finished = 0, the job deletes, and we don't want terraform to recreate the job in the next apply.

I modified the update function, first I attempt to get the job, if a NotFound error occurs, we proceed to check the previous ttl_seconds_after_finished, I then check if ttl = 0. If so set the resource id to "" to remove it from the state, and with my understanding I thought terraform would recreate it.

But when attempting that solution, I get this error produced an unexpected new value: Root object was present, but now absent.
And I believe that's due to, during an update terraform expects the resource to remain in the state unless explicitly destroyed by the config.

I will give it some more thought today, with that being said do you have another idea in mind? Is this ttl solution viable, meaning with us solving one issue, another edge case arises.

kubernetes/resource_kubernetes_job_v1.go
```diff
@@ -48,6 +51,78 @@ func resourceKubernetesJobV1() *schema.Resource {
...
+	if oldTTLRaw != nil {
+		oldTTLStr, _ = oldTTLRaw.(string)
+	}
+	if newTTLRaw != nil {
+		newTTLStr, _ = newTTLRaw.(string)
+	}
+
+	oldTTLInt, err := strconv.Atoi(oldTTLStr)
+	if err != nil {
+		oldTTLInt = 0
```
jrhouston, line 74: I'm not sure this makes sense – if the number supplied in the config doesn't even parse as an integer we should produce an error rather than defaulting it to `0` I think.

kubernetes/resource_kubernetes_job_v1.go
```diff
@@ -48,6 +51,78 @@ func resourceKubernetesJobV1() *schema.Resource {
...
+		return nil
+	}
+
+	// Retrieve old and new TTL values as strings
+	oldTTLRaw, newTTLRaw := d.GetChange("spec.0.ttl_seconds_after_finished")
+
+	var oldTTLStr, newTTLStr string
+
+	if oldTTLRaw != nil {
+		oldTTLStr, _ = oldTTLRaw.(string)
```
jrhouston, line 66: You can just omit the `, _` here if you're not going to check the type assertion.

kubernetes/resource_kubernetes_job_v1.go
```diff
@@ -48,6 +51,78 @@ func resourceKubernetesJobV1() *schema.Resource {
...
+	}
+
+	// Check if the Job exists
+	_, err = conn.BatchV1().Jobs(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// Job is missing
+			if oldTTLInt >= 0 {
+				if oldTTLInt != newTTLInt {
+					// TTL value changed; force recreation
```
jrhouston, line 98: I actually don't think you need to put the logic for forcing recreation when the ttl changes here. You just need to set `ForceNew: true` here: 

https://github.com/hashicorp/terraform-provider-kubernetes/blob/69263f8d4b7eb1b1227dbd7ee985fd107d195a54/kubernetes/schema_job_spec.go#L235-L247

jrhouston (changes requested):

kubernetes/resource_kubernetes_job_v1.go
JaylonmcShan03: In ttl_seconds_after_finished attr, the field ForceNew is now set to true!

kubernetes/resource_kubernetes_job_v1.go
JaylonmcShan03: ,_ is not omitted, thanks for the feedback!

kubernetes/resource_kubernetes_job_v1_test.go
```diff
@@ -516,3 +592,53 @@ func testAccKubernetesJobV1Config_modified(name, imageName string) string {
...
+        restart_policy = "Never"
+      }
+    }
+  }
+  wait_for_completion = false
+}
+`, name, imageName)
+}
+
+func testAccKubernetesJobV1Config_customizeDiff_ttlFive(name, imageName string) string {
```
jrhouston, line 621: Instead of having 2 configs, you could just make the ttl value of a parameter to the function and template it in.

kubernetes/resource_kubernetes_job_v1.go
```diff
@@ -48,6 +51,51 @@ func resourceKubernetesJobV1() *schema.Resource {
...
+
+	// Suppress the diff if the old and new TTL values are the same
+	if oldTTLInt == newTTLInt {
+		log.Printf("[DEBUG] ttl_seconds_after_finished has not changed; suppressing diff")
+		d.Clear("spec")
+		d.Clear("metadata")
+	}
+
+	return nil
+}
```
jrhouston, line 97: It doesn't make sense to me why we would have a diff function that checks if the value of `ttl_seconds_after_finished` hasn't changed and then suppress the diff? 

I tried removing this function completely and your tests still pass.

kubernetes/resource_kubernetes_job_v1.go
```diff
@@ -173,6 +230,31 @@ func resourceKubernetesJobV1Update(ctx context.Context, d *schema.ResourceData,
...
+				// Job was deleted due to TTL nothing to update
+				log.Printf("[INFO] Job %s not found but ttl_seconds_after_finished = %v; nothing to update", d.Id(), ttlInt)
+				return nil
+			}
+
+			// Job was deleted unexpectedly; return an error
+			return diag.Errorf("Job %s not found; cannot update because it has been deleted", d.Id())
+		}
+		return diag.Errorf("Error retrieving Job: %s", err)
+	}
```
jrhouston, line 255: I don't think you need this logic either. In the read function it's going to check if the job exists and then set the ID to `""` if it doesn't which will force recreation and so the update function wont get called. 

I tried commenting this code out and all your tests still pass too.

jrhouston (changes requested):

jrhouston (changes requested):

kubernetes/resource_kubernetes_job_v1_test.go
```diff
@@ -237,6 +237,82 @@ func TestAccKubernetesJobV1_ttl_seconds_after_finished(t *testing.T) {
...
+				Config: testAccKubernetesJobV1Config_Diff(name, imageName, 0),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckKubernetesJobV1Exists(resourceName, &conf),
+					resource.TestCheckResourceAttr(resourceName, "spec.0.ttl_seconds_after_finished", "0"),
+				),
+			},
+			// Step 2: Wait for the Job to complete and be deleted
+			{
+				PreConfig: func() {
+					time.Sleep(120 * time.Second)
```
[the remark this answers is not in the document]
JaylonmcShan03: HAHA, I was experimenting with sleep when testing out my logic. I totally forgot to remove that outrageous time. Will fix it!

kubernetes/resource_kubernetes_job_v1_test.go
```diff
@@ -516,3 +592,28 @@ func testAccKubernetesJobV1Config_modified(name, imageName string) string {
...
+  }
+  spec {
+    ttl_seconds_after_finished = %d
+    template {
+      metadata {}
+      spec {
+        container {
+          name    = "wait-test"
+          image   = "%s"
+          command = ["sleep", "60"]
```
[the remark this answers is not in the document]
JaylonmcShan03: Will fix this as well!

jrhouston (approved):

PR #2623, opened 2024-11-18
Update CODEOWNERS

iBrandyJackson:
### Description

update to HybridCloud team from legacy K8s team


SarahFrench (commented):
You need to give this team permissions on the repo first I think, currently this group isn't listed under here: https://github.com/hashicorp/terraform-provider-kubernetes/settings/access

Issue #2608, opened 2024-10-24
Create kubernetes_persistent_volume_claim from snapshot


divomen:
### Description

Currently kubernetes_persistent_volume_claim does not allow to specify data source.
This prevents us from using terraform to automate volume backups.

### Potential Terraform Configuration

```
resource "kubernetes_persistent_volume_claim" "default" {
  metadata {
    name = "my-pvc-clone"
  }
  spec {
    access_modes = ["ReadWriteOnce"]
    resources {
      requests = {
        storage = "1Gi"
      }
    }
    data_source {
      persistent_volume_claim {
        claim_name = "my-pvc-original"
      }
    }
  }
}
```

Сorresponding kubectl manifest

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-pvc-clone
spec:
  dataSource:
    kind: PersistentVolumeClaim
    name: my-pvc-original
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi

```

BBBmau:
Looks like this was opened previously but wasn't worked on due to needing to support older k8s versions at the time. This isn't an issue now as it's been stable since `v1.20`

related issue: https://github.com/hashicorp/terraform-provider-kubernetes/issues/1232

JaylonmcShan03:
@BBBmau I can take this on for Q4 :)

PR #2644, opened 2024-12-09
add  field for PersistentVolumeClaim resource

JaylonmcShan03:
### Description
Addresses #2608 

This pull request introduces the `data_source` field for the PersistentVolumeClaim resource. 

### Changes
1. Added `data_source` field to the PersistentVolumeClaim schema.
2. Implemented validation in the `Create` function to ensure the specified `data_source` PVC exists.
3. Included the new field in flatten and expand method.

### Acceptance tests

```

...
```

### Release Note

```release-note
...
```

### Community Note

### DIFF
--- a/kubernetes/resource_kubernetes_persistent_volume_claim_v1.go
+++ b/kubernetes/resource_kubernetes_persistent_volume_claim_v1.go
@@ -103,6 +103,15 @@ func resourceKubernetesPersistentVolumeClaimV1Create(ctx context.Context, d *sch
 	if err != nil {
 		return diag.FromErr(err)
 	}
+	if claim.Spec.DataSource != nil {
+		_, err := conn.CoreV1().PersistentVolumeClaims(claim.Namespace).Get(ctx, claim.Spec.DataSource.Name, metav1.GetOptions{})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return diag.Errorf("data source PersistentVolumeClaim '%s' not found", claim.Spec.DataSource.Name)
+			}
+			return diag.FromErr(err)
+		}
+	}
 	log.Printf("[INFO] Creating new persistent volume claim: %#v", claim)
 	out, err := conn.CoreV1().PersistentVolumeClaims(claim.Namespace).Create(ctx, claim, metav1.CreateOptions{})
 	if err != nil {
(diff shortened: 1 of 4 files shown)


BBBmau (changes requested):
although tests are passing it seems like we're missing two fields that are mentioned in the [docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#using-volume-populators) `kind` and `apiGroup`

It looks like the PR is assuming that both kind and apiGroup are setup in a way that sets up the dataSource as PVC but from the docs users have the option to input their own values (though this is alpha something to note)

I'm thinking we should stick more with how the api and manifest is formatted meaning instead of 

```hcl
    data_source {
      persistent_volume_claim {
        claim_name = "example-source-pvc"
      }
    }
```

we should do 

```hcl
    data_source {
        name = "example-source-pvc"
    }
```

leaving the other fields as defaults but not part of the schema. once they are moved out of alpha we can add them in. This method would prevent any extra worked needed later on while also preventing some unexpected changes for users that would use this.

Hope this makes sense, let me know your thoughts!

evilhamsterman:
@BBBmau The kind and apiGroup field is actually very important in the stable API. It is possible to specify the dataSource as either a VolumeSnapshot or a PVC. You need the kind field to differentiate in that case you have both with the same name.

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#create-persistent-volume-claim-from-volume-snapshot
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#create-persistent-volume-claim-from-an-existing-pvc

BBBmau:
> @BBBmau The kind and apiGroup field is actually very important in the stable API. It is possible to specify the dataSource as either a VolumeSnapshot or a PVC. You need the kind field to differentiate in that case you have both with the same name.
> 
> [kubernetes.io/docs/concepts/storage/persistent-volumes#create-persistent-volume-claim-from-volume-snapshot](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#create-persistent-volume-claim-from-volume-snapshot) [kubernetes.io/docs/concepts/storage/persistent-volumes#create-persistent-volume-claim-from-an-existing-pvc](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#create-persistent-volume-claim-from-an-existing-pvc)

This makes sense, not sure how I glanced over that. Thanks for bringing it up.

cc: @JaylonmcShan03

evilhamsterman:
Sorry to nag, but I've been waiting for this update to unblock some functionality at work. Any update on when this will get updated and merged?

Issue #2649, opened 2024-12-16
Regression in Job deployment caused by #2596


justinmchase:
### Terraform Version, Provider Version and Kubernetes Version

```
Terraform version:
Kubernetes provider version: v2.34.0
Kubernetes version: AKS 1.31.1
```

### Affected Resource(s)
- `kubernetes_job`

### Terraform Configuration Files
What do you mean by "configuration"? Certainly you don't mean the entire terraform module.

### Debug Output
> Working on this, will attach soon.

### Panic Output
`N/A`

### Steps to Reproduce
1. Use kuberenetes provivder `v2.34.0` or greater.
2. Create a `kubernetes_job` resource which does not have `ttl_seconds_after_finished` set.

### Expected Behavior
1. The job should be recreated on every deployment
2. The deployment should succeed

### Actual Behavior
1. The job is created and the job pod runs successuflly
2. Terraform returns a non-zero exit code and reports the Job as failing.

### Important Factoids
This code has been running fine for a long time and the issue may be caused by this PR that was just merged #2596

Also, and I don't fully understand the PR creators scenario but this assumption seems bad to me:

> This can cause Terraform to plan the recreation of the Job in subsequent runs, which is sort of undesirable behavior. 

In fact this is _highly desirable and required_ behavior for us. We explicitly want the Job to be re-created on every single deployment.

So is this change a regression? Is there some setting I can set it to that will cause it to be recreated on every deployment?

Regardless, having it fail even though the pod is succeeding seems like a bug.

### References
- #2596

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

justinmchase:
We were able to rollback to version `2.33.0` and it resolved our issue, so its definitely a recent issue probably this PR specifically.

PR #2650, opened 2024-12-16
Revert "Handle Jobs with ttl_seconds_after_finished = 0 correctly"

justinmchase:
Reverts hashicorp/terraform-provider-kubernetes#2596

Per issue #2649, please revert this breaking change which regresses a critical feature.


.changelog/2596.txt
```diff
@@ -1,3 +0,0 @@
-```release-note:enhancement
-Properly handle Kubernetes Jobs with ttl_seconds_after_finished = 0 to prevent unnecessary recreation.
-```
```
arybolovlev, line 3: Please don't remove this file. Instead, please add a new one that will reflect your changes. Your changes fall under the _bug fix_ category. Please mention there that this bug fix reverts changes introduced in 2.34.0.

arybolovlev (changes requested):
Hi @justinmchase,

Thank you for reverting this change and properly reporting the related issue. Indeed, it looks like a breaking change and we need to find a better way to cover both scenarios without introducing a breaking change in the minor release.

I have left a small comment. Once it is addressed we can merge this change and cut a patch release.

.changelog/2596.txt
justinmchase: Makes sense, will do.

justinmchase:
I updated the release notes files as requested and merged upstream main.

arybolovlev (approved):
Thanks, @justinmchase!

dmpakhar:
This change broke functionality, when job should be created once and then deleted.
In my scenario k8s job should be launched only once. Now this job recreates every apply and there are no ability not to skip creation of this job in the second run or small changes that would not require execution of the job.

justinmchase:
@dmpakhar Is it possible to add a separate field to get the behavior you want without regressing the behavior that is currently expected when ttl seconds after finish is 0?

If it helps, I will tell you that our Job is running a migration pod, the migration pod does run on every deployment but internally it is keeping track of the state of migration steps and applying them only once. This is a pretty standard way of handling run-once behavior rather than trying to manage it with terraform and kubernetes.

dmpakhar:
@justinmchase 
Hi,
I think this can be managed as an extra flag. From my perspective this behaviour should be defined via extra var. this will make logic clear for everyone.

Now we went in same issue, when same parameter changes behaviour.

I think execute_once/recreate_after_finished = true/false variable may be a good idea

PR #2699, opened 2025-03-08
Do less Kubernetes API requests

g7r:
### Description

The first reason why provider does so many requests is because its caches aren't goroutine-safe. For example, Terraform invokes provider concurrently and every individual goroutine starts its own getOAPIv2Foundry() invocation. Every getOAPIv2Foundry() starts its own Kubernetes API request and these requests consume Kubernetes client Burst leading eventually to a stall.

The second reason is that CRDs should also be cached because production Kubernetes clusters may have lots and lots of CRDs and getting them all is not cheap. Furthermore, getting them over and over consumes Kubernetes client Burst leading to a major stall.

### Release Note

```release-note
Do less Kubernetes API invocations
Make `RawProviderServer` caches goroutine-safe
```

### Community Note

### DIFF
--- a/.changelog/2699.txt
+++ b/.changelog/2699.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fixed goroutine-safety in the CRD and metadata cache, resulting in far fewer provider metadata requests.
+```
--- a/manifest/provider/cache.go
+++ b/manifest/provider/cache.go
@@ -1,3 +1,6 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
 package provider
 
 import "sync"
--- a/manifest/provider/cache.go
+++ b/manifest/provider/cache.go
@@ -1,0 +1,16 @@
+package provider
+
+import "sync"
+
+type cache[T any] struct {
+	once  sync.Once
+	value T
+	err   error
+}
+
+func (c *cache[T]) Get(f func() (T, error)) (T, error) {
+	c.once.Do(func() {
+		c.value, c.err = f()
+	})
+	return c.value, c.err
+}
--- a/manifest/provider/clients.go
+++ b/manifest/provider/clients.go
@@ -31,98 +31,71 @@ const (
 
 // getDynamicClient returns a configured unstructured (dynamic) client instance
 func (ps *RawProviderServer) getDynamicClient() (dynamic.Interface, error) {
-	if ps.dynamicClient != nil {
-		return ps.dynamicClient, nil
-	}
 	if ps.clientConfig == nil {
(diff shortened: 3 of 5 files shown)


alexsomesan:
> Hello @alexsomesan, @jrhouston, and @arybolovlev,
> 
> 
> 
> Hope you're having a good week.
> 
> 
> 
> I submitted this PR about a month ago and just wanted to gently follow up to make sure the notification came through and it landed on your radar.
> 
> 
> 
> This PR has two main benefits: it addresses a significant performance issue (reducing `terraform plan` time from ~3 minutes back down to ~15 seconds) and also helps make the provider's behavior more correct in this context.
> 
> 
> 
> I'd be grateful if you could take a look when you have a moment.
> 
> 
> 
> Thank you for your consideration and all your work on this provider!

Hello @g7r 
Thanks for taking the time to look into these performance issues. 

Our whole team is away this week, taking part in a company event. I will have a look at your PR first thing next week, when I'm back to a regular schedule.

g7r:
@alexsomesan, I still badly need this fix merged 🙏🏻🙏🏻🙏🏻

This little PR only fixes an erroneous behaviour under the hood without introducing anything new. It, however, makes the life of those with hundreds of Kubernetes resources in their Terraform stacks much much easier.

alexsomesan:
@g7r Apologies, I'm reviewing this today.

alexsomesan (commented):
I thin the changes look technically correct. However, I'm finding it difficult to understand where the performance gains are coming from. See my spot comment about the caching mechanism, as an example.

Could you please elaborate a bit how you came to this solution?

g7r:
> Could you please elaborate a bit how you came to this solution?

Sure. Terraform may use multiple concurrent requests to the provider when there are multiple Kubernetes resources in the stack. When the provider serves a request, it first tries to retrieve CRDs from Kubernetes cluster. The provider has caches but the implementation lacks goroutine-safety.

So, when multiple concurrent requests come, the provider checks whether it already has CRDs cached. But it doesn't. So it starts multiple concurrent requests to Kubernetes API, because there are no syncronization primitives. And the main issue with this behaviour is that these redundant requests consume Kubernetes client request budget. The Kubernetes client uses client-level throttling by default. And its request budget is quite limited. When the budget is exhausted, all Kubernetes API requests stall for a minute or so.

We have stacks with a lot of Kubernetes resources. When we run Terraform on these stacks with higher `N` in `-parallelism=N`, the Kubernetes request queue in the provider may take several minutes to be served by Kubernetes client. This PR solves the problem by fixing goroutine safety in CRDs requests. The provider now makes at most one request for the CRDs. All requests except the first either wait for the first request to finish or take the cached result.

PR #2882, opened 2026-05-10
fix: remove nil pointer dereference in service account data source on 404

### LINKED ISSUE #2888: panic: nil pointer dereference in kubernetes_service_account_v1 data source on 404

Any version since #2464 (v2.30.0+).

### Affected Resource(s)

- `kubernetes_service_account_v1` (data source)

### Description

`dataSourceKubernetesServiceAccountV1Read` crashes with a nil pointer dereference when the Kubernetes API returns a 404 (service account not found).

The `IsNotFound` branch at line 75 calls `buildId(sa.ObjectMeta)` on the nil response object:

```go
if apierrors.IsNotFound(err) {
    d.SetId(buildId(sa.ObjectMeta))  // sa is nil here, panics
    return nil
}
```

All 13 other data sources in this provider handle 404 by simply returning nil without calling `buildId`.

### Steps to Reproduce

1. Reference a `kubernetes_service_account_v1` data source for a service account that does not exist
2. Run `terraform plan`
3. Provider crashes with a nil pointer dereference panic

### Expected Behavior

The data source should return empty state (as all other data sources do on 404).

### Actual Behavior

Provider crashes with:
```
panic: runtime error: invalid memory address or nil pointer dereference
```

### References

- Introduced in #2464
- Fix: #2882

SebTardif:
## Rollback Plan

No changes to security controls.

### Description

Fix a nil pointer dereference in `dataSourceKubernetesServiceAccountV1Read` that crashes the provider when a service account is not found.

When the Kubernetes API returns a 404, the response object `sa` is nil. The `IsNotFound` branch at line 75 calls `buildId(sa.ObjectMeta)` on the nil response, which panics:

```go
if apierrors.IsNotFound(err) {
    d.SetId(buildId(sa.ObjectMeta))  // sa is nil, panics here
    return nil
}
```

**Fix:** Remove the `d.SetId(buildId(sa.ObjectMeta))` line, matching the pattern used by all 13 other data sources in this provider, which simply `return nil` on 404 to signal the resource does not exist.

This bug was introduced in #2464 (2024-04-10) when reverting data sources from erroring on not-found to returning empty state. The `buildId` call was left in by mistake.

### Acceptance tests

This is a one-line deletion that aligns with the established pattern across all other data sources.

```release-note
fix: remove nil pointer dereference in `kubernetes_service_account_v1` data source when the service account does not exist
```

Fixes #2888

### References
- #2464 (introduced the bug)
- #731 (original data source implementation, without this issue)


rigalGit:
@SebTardif acctests are failing, can you please check
```
=== NAME  TestAccKubernetesDataSourceServiceAccountV1_not_found
    data_source_kubernetes_service_account_v1_test.go:87: Step 1/1 error: Check failed: Check 1/3 error: data.kubernetes_service_account_v1.test: Attribute 'metadata.0.name' not found
--- FAIL: TestAccKubernetesDataSourceServiceAccountV1_not_found (0.72s)
```

SebTardif:
@rigalGit

> acctests are failing, can you please check

Fixed in 5c0e8d64. The original code called `d.SetId(buildId(sa.ObjectMeta))` where `sa` is nil on a 404 response (the nil deref this PR fixes). My first fix removed the `SetId` call entirely, which caused the data source to lose its state and broke the `_not_found` acceptance test.

The fix now uses the `metadata` variable (already extracted from the Terraform config on line 72) to build the ID: `d.SetId(buildId(metadata))`. This preserves state on 404 while avoiding the nil pointer dereference.

PR #2956, opened 2026-08-13
Security Vulnerability Fix

isivaselvan:
manifest/openapi: bump golang.org/x/crypto, grpc, and kin-openapi for security fixes

## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

No changes to access controls, encryption, or logging. This PR only bumps three dependencies flagged by an automated security scan and fixes the resulting compile/behavior breakage in `manifest/openapi/`. See the reachability notes below - none of the flagged advisories have a live attack surface in this repo's actual usage of these libraries, but all three are bumped regardless as dependency hygiene.

### Description

Bumps three dependencies to their patched versions:

| Dependency | From | To | Advisory |
|---|---|---|---|
| `golang.org/x/crypto` | v0.49.0 | v0.52.0 | 7 SSH auth-bypass/DoS GHSAs |
| `google.golang.org/grpc` | v1.79.3 | v1.82.1 | GHSA-hrxh-6v49-42gf (xDS RBAC bypass/panic, HTTP/2 Rapid Reset) |
| `github.com/getkin/kin-openapi` | v0.111.0 | v0.144.0 | GHSA-r277-6w6q-xmqw, GHSA-jpcw-4wr7-c3vq (`openapi3filter` auth bypass / nil-deref panic) |

**Required code changes (not a drop-in bump):** the `kin-openapi` jump spans 33 minor versions and breaks API compatibility in ways caught by the compiler and test suite:

1. `Schema.Type` changed `string` → `*openapi3.Types` (OpenAPI 3.1 multi-type support) - added a `schemaType()` helper in `manifest/openapi/schema.go`.
2. `AdditionalProperties` changed `*SchemaRef` → `BoolSchema{Has, Schema}` - updated the four call sites in `schema.go` that branch on object/map/array shape.
3. `openapi2.SchemaRef` is now its own type, no longer aliased to `openapi3.SchemaRef` - `manifest/openapi/foundry_v2.go` now converts v2 (Swagger) definitions to their v3 equivalent via the library's own `openapi2conv.ToV3Schemas`.
4. `Schema.Extensions` map values changed from `json.RawMessage` to already-decoded `any` - fixed three call sites reading `x-kubernetes-group-version-kind`, `x-kubernetes-preserve-unknown-fields`, and `x-kubernetes-int-or-string`, which previously panicked on a bad type assertion.
[...]

PR #2958, opened 2026-08-20
ci: Fix broken Jira search block PR merges

isivaselvan:
## Rollback Plan

If a change needs to be reverted, revert this commit - it only changes CI workflow behavior, nothing in the provider binary.

None. This is a CI-only workflow change; no provider code, schema, or runtime behavior is affected.

### Description

The `Jira sync` check (`.github/workflows/jira_pr.yaml`) has been failing on every PR event other than `opened` (e.g. edits, closes, reopens) - including on [#2956](https://github.com/hashicorp/terraform-provider-kubernetes/pull/2956), where it was the only failing check. Traced the root cause to the workflow's `Search` step, which calls `tomhjp/gh-action-jira-search`. That action's `main.go` hardcodes a call to Jira's `GET /rest/api/3/search` endpoint, which Atlassian has since retired:

```
API call GET /rest/api/3/search failed (410):
"The requested API has been removed. Please migrate to the /rest/api/3/search/jql API."
```

Since this workflow is internal Jira-ticket bookkeeping with no bearing on code correctness, this PR marks the `Search` step `continue-on-error: true` so it stops blocking `mergeStateStatus` on unrelated PRs while upstream remains unfixed. The downstream steps (`Sync comment`, `Close PR`, `Reopen PR`) already gate on `steps.search.outputs.issue`, so when `Search` fails without setting that output, they skip cleanly instead of the job erroring - no change needed there.

Scope note: `.github/workflows/jira_issues.yaml` has the identical pattern and will fail the same way, but is intentionally left out of this PR to keep the change scoped to what's actually blocking PR merges; happy to follow up separately if useful.

### Acceptance tests

Not applicable — this is a GitHub Actions workflow change, not provider code, so `make testacc` doesn't apply. Verified instead via:

```
$ actionlint .github/workflows/jira_pr.yaml
# same pre-existing shellcheck warnings as origin/main (unrelated Set-ticket-type
# script quoting); nothing new introduced by this change

$ act workflow_dispatch -W <synthetic repro workflow> -P ubuntu-latest=catthehacker/ubuntu:act-latest
# reproduced the exact step/output mechanism locally against Docker:
# Search step fails (continue-on-error: true) -> "Failed but continue next step"
# downstream step gated on steps.search.outputs.issue -> skipped cleanly, never runs
# job result -> "Job succeeded"
```

### Release Note

```release-note
NONE
```

### References

- Observed failing on [#2956](https://github.com/hashicorp/terraform-provider-kubernetes/pull/2956)
- Upstream root cause: [tomhjp/gh-action-jira-search main.go](https://github.com/tomhjp/gh-action-jira-search/blob/main/main.go) hardcodes `/rest/api/3/search`
- Atlassian's deprecation notice: `/rest/api/3/search` removed, migrate to `/rest/api/3/search/jql`

### Community Note

### DIFF
--- a/.changelog/2958.txt
+++ b/.changelog/2958.txt
@@ -1,0 +1,4 @@
+Internal CI-only change: no user-facing effect on the provider, so no release-note entry.
+Marks the "Jira sync" workflow's Search step as continue-on-error, since upstream
+(tomhjp/gh-action-jira-search) still calls a Jira REST endpoint Atlassian has removed,
+which was otherwise blocking PR merges on unrelated code changes.
--- a/.changelog/2958.txt
+++ b/.changelog/2958.txt
@@ -1,4 +1,0 @@
-Internal CI-only change: no user-facing effect on the provider, so no release-note entry.
-Marks the "Jira sync" workflow's Search step as continue-on-error, since upstream
-(tomhjp/gh-action-jira-search) still calls a Jira REST endpoint Atlassian has removed,
-which was otherwise blocking PR merges on unrelated code changes.
--- a/.github/workflows/jira_pr.yaml
+++ b/.github/workflows/jira_pr.yaml
@@ -75,6 +75,13 @@ jobs:
       - name: Search
         if: github.event.action != 'opened'
         id: search
+        # Upstream (tomhjp/gh-action-jira-search) still calls the deprecated
+        # Jira /rest/api/3/search endpoint, which Atlassian has removed
+        # (HTTP 410). continue-on-error keeps this housekeeping sync from
+        # blocking PR merges until upstream migrates to /rest/api/3/search/jql;
+        # downstream steps already gate on steps.search.outputs.issue, so they
+        # skip cleanly when this step fails without setting it.
+        continue-on-error: true
         uses: tomhjp/gh-action-jira-search@04700b457f317c3e341ce90da5a3ff4ce058f2fa # v0.2.2
         with:
           # cf[10089] is Issue Link (use JIRA API to retrieve)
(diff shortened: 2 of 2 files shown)


.changelog/2958.txt
```diff
@@ -0,0 +1,4 @@
+Internal CI-only change: no user-facing effect on the provider, so no release-note entry.
+Marks the "Jira sync" workflow's Search step as continue-on-error, since upstream
+(tomhjp/gh-action-jira-search) still calls a Jira REST endpoint Atlassian has removed,
+which was otherwise blocking PR merges on unrelated code changes.
```
rigalGit: this file is not needed.

.github/workflows/jira_pr.yaml
```diff
@@ -75,6 +75,13 @@ jobs:
       - name: Search
         if: github.event.action != 'opened'
         id: search
+        # Upstream (tomhjp/gh-action-jira-search) still calls the deprecated
+        # Jira /rest/api/3/search endpoint, which Atlassian has removed
+        # (HTTP 410). continue-on-error keeps this housekeeping sync from
+        # blocking PR merges until upstream migrates to /rest/api/3/search/jql;
+        # downstream steps already gate on steps.search.outputs.issue, so they
+        # skip cleanly when this step fails without setting it.
+        continue-on-error: true
```
rigalGit: instead of this, let's disable the whole action for now.

.changelog/2958.txt
isivaselvan: without this file, changelog check was failing at pr. If not needed will remove this one.

.github/workflows/jira_pr.yaml
isivaselvan: Okay!

iam404 (approved):

PR #2633, opened 2024-11-25
fix cluster role binding namespace handling

JaylonmcShan03:
### Description
Fixes #710
This PR addresses and fixes the handling of namespaces in the kubernetes_cluster_role_binding_v1 resource

I am looking for feedback on my fix! Initially, I implemented logic in structures_rbac.go to handle namespace values directly. However, during acceptance testing, I encountered plan difference errors, where the namespace field was toggling between "default" and null. These differences caused Terraform to repeatedly attempt updates, believing the resource was not in the desired state.

To address this issue, I modified the schema to include diff suppression for the namespace field:
Diff suppression ensures that when the namespace field is either "default" (defaulted by Kubernetes) or null (irrelevant for Group and User), Terraform does not detect a difference.

### Acceptance tests
- [ ] Have you added an acceptance test for the functionality being added?
- [ ] Have you run the acceptance tests on this branch?

Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccXXX'

...
irectory=/var/folders/bp/l16ph9cj7958ml9t254_zcv00000gn/T/plugintest2388791179 test_step_number=1
2024-11-25T08:43:36.326-0600 [DEBUG] sdk.helper_resource: Finished TestCase: test_name=TestAccKubernetesClusterRoleBindingV1_namespaceHandling
--- PASS: TestAccKubernetesClusterRoleBindingV1_namespaceHandling (1.77s)
PASS
ok      github.com/hashicorp/terraform-provider-kubernetes/kubernetes
```

### Release Note
Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-kubernetes/blob/main/CHANGELOG.md):

```release-note
...
```

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #2971, opened 2026-09-13
Stop defaulting RBAC subject namespace for User and Group subjects

### LINKED ISSUE #710: kubernetes_cluster_role_binding & kubernetes_role_binding adding namespace when subject kind is Group
Hi

When applying a clusterrolebinding or rolebinding where the subject kind is `Group`, there should not be a namespace as a group is not a namespaced resource.
There's documentation here: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-examples

### Terraform Version
Terraform v0.12.16
+ provider.aws v2.42.0
+ provider.kubernetes v1.10.0

### Affected Resource(s)
- kubernetes_cluster_role_binding
- kubernetes_role_binding

```hcl
resource "kubernetes_cluster_role_binding" "developer_cluster" {
  metadata {
    name = "company-developer"
  }
  role_ref {
    api_group = "rbac.authorization.k8s.io"
    kind      = "ClusterRole"
    name      = kubernetes_cluster_role.developer_cluster.metadata.0.name
  }
  subject {
    api_group = "rbac.authorization.k8s.io"
    kind      = "Group"
    name      = "company:developer"
  }
}

resource "kubernetes_role_binding" "developer_namespace" {
  metadata {
    name      = "company-developer"
    namespace = kubernetes_namespace.app.metadata.0.name
  }
  role_ref {
    api_group = "rbac.authorization.k8s.io"
    kind      = "ClusterRole"
    name      = kubernetes_cluster_role.developer_namespace.metadata.0.name
  }
  subject {
    api_group = "rbac.authorization.k8s.io"
    kind      = "Group"
    name      = "company:developer"
  }
}
```

### Expected Behavior
The subject blocks of the role bindings should be created as per the config without a namespace. 

### Actual Behavior
The role bindings were created and the namespace field was added with a value of `default`.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

KR-Ravindra:
## Rollback Plan

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

No. Subjects of kind `User` and `Group` are now sent to the API server without an injected namespace, which is how `kubectl` creates them. The RBAC authorizer does not use the namespace for those kinds, so the permissions granted by existing bindings do not change.

#### Problem

`kubernetes_role_binding_v1` and `kubernetes_cluster_role_binding_v1` (and their unversioned aliases) set `namespace = "default"` on every `subject`, including subjects of kind `User` and `Group`. Those kinds are not namespaced, and the RBAC API documents that a non-empty namespace on them should be treated as an error. Users currently have to write `namespace = ""` on each subject to work around it.

#### Root cause

`rbacSubjectSchema()` in `kubernetes/schema_rbac.go` declares `namespace` with `Default: "default"`, which applies to every subject kind, and `expandRBACSubjects()` in `kubernetes/structures_rbac.go` copies the value unconditionally.

#### Fix

- Drop the schema default. `expandRBACSubjects()` sends the namespace only when it is set, and falls back to `default` only for `ServiceAccount` subjects (the API requires a namespace there), so configurations that omit it keep working.
- Add a `DiffSuppressFunc` that treats an unset namespace and `"default"` as equal, so state written by earlier provider versions does not produce a diff after upgrading. The suppression only applies while the subject at that list index still has the same `kind` and `name`; when a subject is removed and the following ones shift down, the stale `"default"` is not carried over to the subject that now occupies the index.
- Update the attribute description in the docs.

Same approach as #2633, which was approved but closed as stale.

**Upgrade trade-off for maintainers to weigh.** Bindings created by earlier provider versions keep `namespace = "default"` on their `User`/`Group` subjects in the cluster. Because `"default"` and unset are treated as equal for an unchanged subject, neither a later edit to another subject nor an explicit `namespace = ""` clears that stale value; only recreating the resource (or changing the subject's `kind`/`name`) does. The RBAC authorizer ignores the namespace for those kinds, so there is no permission impact. The alternative is to suppress only for `ServiceAccount` subjects, which would surface a one-time repair diff on every existing binding after upgrading. This PR takes the no-diff option; happy to switch if you prefer the repair diff.

Unrelated, observed while writing the test: `subject.api_group` is `Optional` + `Computed`, so replacing a `User`/`Group` subject at an index with a `ServiceAccount` fails with `subjects[i].apiGroup: Unsupported value: "rbac.authorization.k8s.io"` because the old value is kept. That exists on `main` and is not addressed here.

#### How tested

Unit tests cover `expandRBACSubjects`/`flattenRBACSubjects`, the schema (a `Group`/`User` subject without a namespace must expand to an empty namespace) and the diff suppression through `Resource.Diff` against legacy state (`namespace = "default"` on every subject): unchanged subjects plan no diff, removing the leading `ServiceAccount` plans the `Group` that shifts into its index with an empty namespace, and an explicit namespace is never suppressed. The acceptance test creates a binding with `ServiceAccount`, `Group`, `User` and `ServiceAccount`/`kube-system` subjects, checks the namespaces on the object returned by the API server, then removes the leading `ServiceAccount` and checks the `Group`/`User` namespaces on the API object are still empty. Upgrade path checked manually: apply with a build of `main` (state and API object hold `namespace = "default"` on the `Group` subject), then plan with this branch: `No changes.`

Against `main`'s schema and expander, `TestExpandRBACSubjects`, `TestRBACSubjectSchema_namespace` and `TestRBACSubjectSchema_namespaceDiffSuppress` fail, and the acceptance test fails at step 1 (`Group`/`User` namespaces come back as `"default"`). With an unconditional `""`/`"default"` suppression (the first revision of this PR), step 2 of the acceptance test fails (excerpt):

```
Step 2/2 error: Check failed: Check 4/7 error: kubernetes_cluster_role_binding_v1.test: Attribute 'subject.0.namespace' expected "", got "default"
Check 7/7 error: subject[0] (Group "notagroup"): expected namespace "", got "default"
```

### Acceptance tests

Output from acceptance testing (kind, Kubernetes v1.37.0):

```
$ TF_ACC=1 go test ./kubernetes/ -v -vet=off -run 'TestAccKubernetesClusterRoleBindingV1|TestAccKubernetesRoleBindingV1' -parallel 4
--- PASS: TestAccKubernetesRoleBindingV1_sa_subject (2.48s)
--- PASS: TestAccKubernetesRoleBindingV1_Bug (4.52s)
--- PASS: TestAccKubernetesClusterRoleBindingV1_basic (5.65s)
--- PASS: TestAccKubernetesClusterRoleBindingV1_subjectNamespace (3.51s)
--- PASS: TestAccKubernetesRoleBindingV1_basic (6.10s)
--- PASS: TestAccKubernetesClusterRoleBindingV1_serviceaccount_subject (2.37s)
--- PASS: TestAccKubernetesClusterRoleBindingV1_generatedName (2.42s)
--- PASS: TestAccKubernetesClusterRoleBindingV1_group_subject (2.91s)
--- PASS: TestAccKubernetesClusterRoleBindingV1_UpdatePatchOperationsOrderWithRemovals (4.66s)
--- PASS: TestAccKubernetesRoleBindingV1_group_subject (2.39s)
--- PASS: TestAccKubernetesRoleBindingV1_generatedName (2.38s)
--- PASS: TestAccKubernetesClusterRoleBindingV1_identity (3.05s)
--- PASS: TestAccKubernetesRoleBindingV1_identity (2.99s)
PASS
ok  	github.com/hashicorp/terraform-provider-kubernetes/kubernetes	12.215s
```

### Release Note

```release-note:bug
`kubernetes_role_binding_v1`, `kubernetes_cluster_role_binding_v1`: stop setting `namespace = "default"` on subjects of kind `User` and `Group`; the namespace is only defaulted for `ServiceAccount` subjects.
```

### References

Fixes https://github.com/hashicorp/terraform-provider-kubernetes/issues/710
Supersedes https://github.com/hashicorp/terraform-provider-kubernetes/pull/2633

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### Community Note


KR-Ravindra:
Round 1 self-review.

Verified against `main`:
1. The `Default: "default"` on the subject `namespace` schema in `kubernetes/schema_rbac.go` is the only place the value is injected; `expandRBACSubjects` copies it unconditionally, which is why Group/User subjects end up with `namespace: default` in the API object.
2. The `DiffSuppressFunc` (`""` vs `"default"`) matches #2633 (approved by @BBBmau, closed as stale), so existing state that carries `namespace = "default"` on Group/User subjects produces no plan diff and there is no ForceNew on the attribute.
3. ServiceAccount subjects still fall back to `metav1.NamespaceDefault` in expand, which is what the API server requires for ClusterRoleBindings.
4. `.changelog/2971.txt` is present in the format `changelog-checker` expects; the four docs edits are the regenerated schema description.
5. One behaviour note: pre-existing Group/User subjects keep `namespace: default` server-side until the next subject update triggers a patch; state-only suppression, same trade-off as #2633.

The CLA check is pending on my side and will be cleared separately. Marking ready so the unit-test and lint workflows run.

KR-Ravindra:
Round 2 self-review.

1. Revision 96efda6 scopes the `""`/`"default"` suppression to list elements whose `kind` and `name` are unchanged (`d.HasChange` on the sibling keys), so the stale namespace is no longer carried over when a leading subject is removed and the rest shift down. Checked through `Resource.Diff` against legacy state in `TestRBACSubjectSchema_namespaceDiffSuppress`; the four new unit tests pass at this head.
2. `expandRBACSubjects` falls back to `default` only for `ServiceAccount` subjects; `User`/`Group` subjects without a namespace expand to an empty one, matching the body.
3. The `Jira sync` failure is the workflow's Atlassian search call returning 410 (endpoint removed), unrelated to this change. `license/cla` is pending signature.

Stays ready.

